PR #663, opened 2014-03-30
Site menu: Fix broken layout for site menu - IE7

m14e:
(1) Fix broken layout for site menu - IE7; (2) workaround for IE z-index issues; (3) fix horizontal placement

### DIFF
--- a/src/sass/parts/_siteMenu-mediumViewOver.scss
+++ b/src/sass/parts/_siteMenu-mediumViewOver.scss
@@ -71,3 +71,35 @@
 		}
 	}
 }
+
+/* IE */
+
+.lt-ie9 {
+	header {
+		position: static !important;
+	}
+	.wb-menu {
+		.menu {
+			> li.active {
+				ul {
+					margin-left: 0 !important;
+				}
+			}
+		}
+	}
+	#wb-sm {
+		.menu {
+			display: block;
+			z-index: 1000 !important;
+			> li {
+				display: list-item;
+				float: left !important;
+				list-style: none outside none;
+				margin: 0;
+				padding: 0;
+				position: relative;
+				z-index: 1001 !important;
+			}
+		}
+	}
+}


LaurentGoderre:
I will fix this pr on monday

masterbee:
@LaurentGoderre - ok but I just want to be sure that we are working in the right direction since IE7 seems a bit to old on the browser support list, and do not want to add more support than necessary, or even work that will be removed shortly after.

What is your take on this?

LaurentGoderre:
We don't have to make it look good, just functional and this is a small footprint. We should segregate it even further.

pjackson28:
The fixes could be added if they don't affect the download size of other browsers (excluding IE8) but as a project IE7 is outside of the support window for WET so we will not be testing for it. The earliest version of IE that we currently support is IE8 and even that support is coming into question as visits with IE8 are dropping below the 5% support threshold. As reminder, here is the browser support criteria for WET:

* The current and previous major version of each browser that accounts for at least 5 per cent of visits(1) to the website;
* Any major version of a browser that, on its own, accounts for at least 5 per cent of visits(1) to the website; and
* The current major version of each mobile browser that accounts for at least 5 per cent of mobile device visits(1).

(1) For Government of Canada websites: Visits to all public-facing Government of Canada websites and web applications, excluding visits from Government of Canada workstations.

LaurentGoderre:
Is it possible that many of those IE7 come from service points instead of users devices ?

pjackson28:
Another possibility is that certain web servers and/or browsers could be configured in such a way as to trigger compatibility mode in IE (which is effectively IE7). We've seen that a few times with GC websites and GC workstations where it caused IE9 and even IE10 to go into compatibility mode despite accessing a public-facing website (normally this behaviour is reserved or intranets). This was easily fixed on the server by doing the fixes described on the following page: http://wet-boew.github.io/v4.0-ci/docs/server-en.html

MRicher:
We added the condition to force out from compatibility mode for canada.ca with the mod_headers.c in Apache for this issue.

LaurentGoderre:
Hmmm...there is one problem. It seems like there is no IE8 stylesheret for this theme. Adding one would require a markup change...

What do you think @masterbee

masterbee:
@LaurentGoderre - yep I figured and I am hesitant since we will need to create more files and for a browser that is not supported. If this was a simple inclusion to existing files I would gladly say sure, but adding more files for a browser that is not supported makes me think we are heading the wrong direction. As @pjackson28 said earlier lets keep to our defined criteria since messing with grunt task and adding new files have proven to be problematic in the past and since we will not be testing for it, makes me more uneasy about adding an addition resource to the theme for it.

Make sense?

PR #878, opened 2015-01-09
Allow the theme to override WET core variables

LaurentGoderre:


LaurentGoderre:
@shawnthompson the build will fail until this gets merged so this should go in before anything else

PR #892, opened 2015-02-04
Enable the printing of glyphicons

### LINKED ISSUE #888: Glyph icons are not printing properly in Canada.ca theme template only
Hi,

I've noticed a bug in the Canada.ca templates when using Glyph icons where they do not appear when you print the page to paper, ie

<span class=”glyphicon glyphicon-arrow-up”></span>

In Canada.ca template only (works in the other ones I tested it on), when using Glyphicons they appear properly in the browser, however when I goto "Print Preview" (ie in firefox of chrome), none of the glyph icons appear in the print version, see screen shot below:

glyph-icon2

Note I also tested adding the glyph icons to the base and GCWU templates and they work fine in print mode

glyph-icon1
, see screenshot

I believe the issue is unique to canada.ca template.

If there is a resolution before the next release, could i also get the code snippet to implement in our existing templates.

thanks

LaurentGoderre:
Fixes #888

### DIFF
--- a/src/views/_print.scss
+++ b/src/views/_print.scss
@@ -11,10 +11,6 @@
 	@extend %gcweb-print-display-none-important;
 }
 
-.glyphicon {
-	@extend %gcweb-print-display-none-important;
-}
-
 #wb-lng,
 #wb-glb-mn,
 #wb-srch,


ehcee:
Is this in the 4.10 build or the updated style sheets available somewhere for me to implement into our site?

MRicher:
It's not in the stable release but should be in the development release on this page: [Downloads](http://wet-boew.github.io/wet-boew/docs/versions/dwnld-en.html)

ehcee:
What files would i need to grab from the development build? I presume all that was changed was just a few lines in a css file or 2 right? or JS as well?

PR #899, opened 2015-02-12
Images

MRicher:


LaurentGoderre:
If the size of the image is constantly changing, it might be a good idea to not have the dimension of the image in the filename

MRicher:
This should be the last for the carousel images.  They are now using the same ratio so we just need one file (largest) and make them fit the same without cropping.

PR #911, opened 2015-02-24
removed title attribute to follow links using lightbox

shawnthompson:


shawnthompson:
@LaurentGoderre any issues with removing the `title` from the lightbox links here?

LaurentGoderre:
Not on my end no

PR #934, opened 2015-02-26
added share button on pages

### LINKED ISSUE #925: Standardize the "date modified", "feedback" and "share" elements across all templates
Most of the templates use these three consistently, but there are a number of pages that don't. I'm guessing we should probably standardize these across all templates across the board.

One of these days I suppose we should also get around to documenting these in the spec, too :)

shawnthompson:
fixes #925


LaurentGoderre:
why did you open two PRs?

shawnthompson:
I need to revert something.. I just started over... it was faster so I deleted this one, I think

LaurentGoderre:
it was the same branch...you could have forced push your changes over to re-use the same PR

shawnthompson:
still learning, I though it was going to create a new SHA because I git reset HEAD~1, git add (just the modified files) git commit -m "same message"...

PR #975, opened 2015-04-20
changed CSS for mobile/social media pages, issues with changes to v1.5

shawnthompson:


src/mobile-centre-ie.scss
```diff
@@ -1,3 +1,65 @@
 /*
  * Mobile centre
  */
+
+table.product-listing {
```
LaurentGoderre: do you need to add the `table`?

shawnthompson: good question... I'll take a look but I don't think so

PR #985, opened 2015-06-09
Build: Moved core and themes resource in their own subfolders

LaurentGoderre:
Partially implements wet-boew/wet-boew#6809
Also aligned the build to other themes


shawnthompson:
There is a GCWeb sub-folder in the dist.

\dist\GCWeb\GCWeb

LaurentGoderre:
Weird, not anymore

PR #1003, opened 2015-07-23
CSS: Added a rule above the activities and initiatives block.

thomasgohard:
Fixes #997.

PR #1004, opened 2015-07-24
Build: Update the SCSS-Lint config

nschonni:
This will remove the upstream updates from causing failures like in #1003
There are CSS changes in here so review the changes using the `&w=1` to see the non-whitespace changes.

I also added in the upstream `html-proofer` setup, but there currently too many broken links to deal with in this PR.


LaurentGoderre:
What are these `// scss-lint:disable QualifyingElement` for?

nschonni:
@LaurentGoderre the allow turning off rules for small blocks of code, rather than disabling the rule for the whole file. I think some of them can be revisited to fix up later though.

PR #1008, opened 2015-07-28
Build: Update to Grunt-Sass 1.0.0

nschonni:
Move to the libsass 3.3.x core to resolve old extend issues.
Basic diff of changes can be found https://gist.github.com/nschonni/03e8e6f675e595869617. Their are selector changes in a few places, but these appear to be bugs with the old node-sass, since the selectors match the Ruby Sass when comparing on Sassmeister.

The precision was increased to 10 for now, but the later Bootstrap-Sass releases only require 8.

This also frees developers to update to Node 0.12 or iojs.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -250,6 +250,8 @@ module.exports = (grunt) ->
 
 		sass:
 			all:
+				options:
+					precision: 10
 				expand: true
 				cwd: "src"
 				src: "*.scss"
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
:+1:

LaurentGoderre:
We need this for all theme @nschonni. Would you mind?

masterbee:
@nschonni: thanks for this! :+1: :+1:

PR #1015, opened 2015-08-11
CSS: Added missing unit measurement for H1 margin-top property.

EricDunsworth:
Prevents Firefox from sometimes misinterpreting 1.25 as 1.25px. That kind of unit-less measurement is normally ignored by browsers.

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -42,7 +42,7 @@ h1 {
 	border-bottom: 1px solid $border-red;
 	font-size: 34px;
 	margin-bottom: 0.2em;
-	margin-top: 1.25;
+	margin-top: 1.25em;
 	padding-bottom: 0.2em;
 }
 


EricDunsworth:
@LaurentGoderre @masterbee @lucas-hay Just wondering if there's anything holding this back from getting merged in? I could remove the property altogether if needed.

Currently, this property gets generated as "margin-top:1.25;" (without a unit measurement...) in GCWeb's CSS. Browsers ignore unit-less measurements that aren't 0. So the property in its current state just bloats the CSS and doesn't do anything in practice.

To make things even worse, on random occasions, I've noticed Firefox is misinterpreting the property as 1.25px instead of ignoring it... which causes the H1 to practically "stick" to the breadcrumb.

shawnthompson:
:+1: merging now sorry for the delay

EricDunsworth:
@shawnthompson Thanks :).

PR #1029, opened 2015-09-21
Try to fix build error by downgrading npm to what it was before?

andre-gh:


andre-gh:
Test build, wait to see if it passes travis

andre-gh:
Build passing, this is the fix

nschonni:
Should just apply the same change as https://github.com/wet-boew/wet-boew/commit/bdffe629a93667eee1c43ecde791988d63a4aa0e

andre-gh:
Ah, I assumed we were using npm@next for a recent reason rather than a legacy reason, I can PR that into all the branches.

nschonni:
You might need to do this version of the fix on any of the themes that haven't been updated to `grunt-sass` 1.0.0, since that was the thing that was keeping them on Node 0.10 for the longest time.
You might not need to even pin the NPM version at all for the other repos though, since 0.10.40 might have a new enough NPM for those node-sass issues.

andre-gh:
Ah yeah, some of those themes are still on old grunt-sass, I have pull requested a straight node version switchover, if they build fail I will fix them in the morning.

Thanks for your help

PR #1033, opened 2015-09-28
Removed Glyphicons CSS licenses

MaximMartel:


shawnthompson:
This is related to this issue wet-boew/wet-boew#6978

PR #1036, opened 2015-10-14
Menu: Replaced minus hyphens with en dashes in "– More/Autres" links.

EricDunsworth:
This ensures that en dashes are used consistently across all of those kinds of links. Previously, a mix of minus hyphens and en dash characters were used.

### DIFF
--- a/site/includes/fullmenu-en.hbs
+++ b/site/includes/fullmenu-en.hbs
@@ -40,7 +40,7 @@
 					<li><a href="http://travel.gc.ca/returning">Returning to Canada</a></li>
 					<li><a href="http://travel.gc.ca/assistance">Assistance abroad</a></li>
 					<li><a href="http://travel.gc.ca/stay-connected">Stay connected</a></li>
-					<li class="slflnk"><a href="http://travel.gc.ca">Travel - More</a></li>
+					<li class="slflnk"><a href="http://travel.gc.ca">Travel – More</a></li>
 				</ul>
 			</li>
 			<li><a href="#business" class="item">Business</a>
@@ -65,7 +65,7 @@
 					<li><a href="http://www.canada.ca/en/services/benefits/education/index.html">Education and training</a></li>
 					<li><a href="http://www.canada.ca/en/services/benefits/housing.html">Housing benefits</a></li>
 					<li><a href="http://www.canada.ca/en/services/benefits/disability/index.html">Disability benefits</a></li>
-					<li class="slflnk"><a href="http://www.canada.ca/en/services/benefits/index.html">Benefits - More</a></li>
+					<li class="slflnk"><a href="http://www.canada.ca/en/services/benefits/index.html">Benefits – More</a></li>
 				</ul>
 			</li>
 			<li><a href="#health" class="item">Health</a>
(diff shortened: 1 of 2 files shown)


andre-gh:
Please use `&mdash;` instead of the character directly, this helps avoid any confusion, especially when editing code.

andre-gh:
Or `&ndash;` etc, please just not a copy-pasted unicode character.

EricDunsworth:
@andre-gh Done :)! Ended up going with en dash numeric entities since there seems to be a precedent for using numeric characters in wet-boew/GCWeb (for example - &#160; is used all over the placed instead of &nbsp;).

EricDunsworth:
Would using numeric entities in this scenario go against [Appendix C of the Standard on Web Interoperability](http://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=25875#appC)?

It states that web pages and feeds should use UTF-8 characters. Technically the characters that compose the em dash's numeric entity are written in UTF-8, but I think the spirit behind that appendix is encourage the direct use of UTF-8 characters instead of named/numeric entities. I presume non-breaking spaces and ampersands are out of scope for obvious reasons, but aren't so sure about dashes...

@andre-gh @thomasgohard @pjackson28 @LaurentGoderre Thoughts?

andre-gh:
That just specifies that webpages should be encoded in UTF-8 (Eg. Not Latin1) it has nothing to do with HTML entities.  It mostly exists to tell folks to not use UTF16 or Latin1 or whatever else.  If you asked me about say, generated source, or a 1/2 character or a copyright symbol I would put a lot more thought into it but in the case of dashes where it is next to impossible for a reader in a text editor to notice that it is anything else (and that your intentions were indeed to do either one) it is best to do explicit entities for these characters.

andre-gh:
I would say this applies to any character which could be confused with a similar on-keyboard-ascii character or a whitespace character etc.

andre-gh:
If you are against the particular numbered encoding then please use the explicit named entity rather than the numbered character as it is much more easy for the developer to tell what you intended without looking anything up.

EricDunsworth:
@andre-gh My personal preference is either for the direct character or numeric entities. I'm open to modifying the PR as needed.

Some of the other people I addressed in my previous reply may have been involved with the creation of the Standard on Web Interoperability, so I'm interested in getting their perspectives too.

LaurentGoderre:
I prefer the actual character instead of entities, especially if editors of a CMS are not allowed in the HTML view, they better learn how to do the proper characters

andre-gh:
In this case the difference was between - and – there is also — and ―.  Can you tell at a glance which is which and which was intended?

pjackson28:
Numeric entities and actual characters are more interoperable than named entities. For instance if you have a named entity in web page content and you generate a web feed with that content, then that web feed will be invalid because of the presence of the named entity. You can find out more here: http://people.w3.org/~dom/archives/2005/04/named-versus-numeric-entities/

LaurentGoderre:
@andre-gh for someone who cares enough about this kind of stuff, they will

andre-gh:
@LaurentGoderre The CMS argument makes sense, although my worry of using very-similar-characters in source code remains (Is this CMS generated content or is it source code?  Is it both?)

@EricDunsworth Feel free to change this PR back to the raw unicode characters regardless, if things are being migrated to a visual CMS system folks are going to need to visually see the difference anyway.

EricDunsworth:
@LaurentGoderre @pjackson28 Thanks for your input :)!

@andre-gh I'll revise the PR later this week to use "raw" en dashes again. Haven't been able to get around to it due to being kinda swamped over the last couple of days.

EricDunsworth:
@andre-gh Updated again :).

PR #1068, opened 2016-02-22
Service initiation pages: Changed right panel's sections to asides.

EricDunsworth:


shawnthompson:
This comment was meant for here:
>@EricDunsworth we've been talking about this lately and we are trying to decide to use the `aside` or not...

EricDunsworth:
Umm here's a quote of my related response lol.

> @shawnthompson IMO it'd be wise since asides are more appropriate than sections for that kind of content.

shawnthompson:
@MaximMartel @lucas-hay I'd like to merge this PR, once @EricDunsworth fixes the conflic, but I think you should make sure the templates in AEM get updated with this change also.

EricDunsworth:
@shawnthompson Just updated this PR and fixed its conflicts :).

PR #1069, opened 2016-02-22
Simple service initiation page: Added reference to right panel include.

EricDunsworth:
Ensures the simple service initiation page's right panel is consistent with the advanced pages.

### DIFF
--- a/site/pages/service-en.hbs
+++ b/site/pages/service-en.hbs
@@ -9,7 +9,7 @@
 		{ "title": "Topic name", "link": "#" }
 	],
 	"secondlevel": false,
-	"dateModified": "2014-10-23",
+	"dateModified": "2016-02-22",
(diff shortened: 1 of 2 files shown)


shawnthompson:
@EricDunsworth we've been talking about this lately and we are trying to decide to use the `aside` or not...

EricDunsworth:
@shawnthompson IMO it'd be wise since asides are more appropriate than sections for that kind of content. Btw it's actually PR #1068 that adds asides to the include.

The include is currently only used by advanced service initiation pages. In the IA spec's mock-ups, the exact same right boxes are used by both the advanced and simple pages, so I figured it'd make sense to make the simple ones reference the same include. Merging this PR in before a decision is made about asides shouldn't cause any problems (since the include itself still uses sections at this time).

PR #1070, opened 2016-02-24
Theme: Removed body font-weight property from CSS.

### LINKED ISSUE #1040: TH elements do not bold in IE
Via wet-boew/wet-boew#7170

It appears that GCWeb sets font-weight to 300 on body which causes IE (and only IE) to not bold the TH elements by default.

EricDunsworth:
Makes GCWeb consistent with other WET themes in this respect. As a result, table THs now appear bolded in IE8-11 (in line with every other browser).

Fixes #1040 and resolves wet-boew/wet-boew#7170.

@masterbee I noticed you originally introduced ``font-weight: 300;`` via commit c2723ef. Would you have any concerns with my PR removing it? From what I can tell, its removal doesn't change anything at all in other browsers. Just makes tables less strange-looking in IE :).

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -5,7 +5,6 @@ body {
 	background: #f8f8f8;
 	font-family: Helvetica, Arial, sans-serif;
 	font-size: 16px;
-	font-weight: 300;
 }
 
 a {


shawnthompson:
@masterbee Can you answer @EricDunsworth and merge this PR if you agree?

EricDunsworth:
Btw this PR also closes #1075 (it's a duplicate of #1040).

PR #1082, opened 2016-03-09
Added the sign on/off examples

LaurentGoderre:
Depends on wet-boew/wet-boew#7446


shawnthompson:
Sample of page with login button

PR #1090, opened 2016-03-23
Theme: Hid GC activities and initiatives in print view.

EricDunsworth:
Reasoning:

* It's not main content (it's outside the ``<main>`` element). If anyone wants to print out a particular activity/initiative, in practice the odds are high that they'd click the link to it in their browser to open up that item's dedicated page, then print that out.
* It usually adds 1-2 extra pages when printing (depending on how many items are used).
* Its current print layout looks broken. If a given activity/initiative splits across pages, it looks really strange and creates large empty gaps of space.

CC @nicolasraymond @shawnthompson

### DIFF
--- a/src/ie8-theme.scss
+++ b/src/ie8-theme.scss
@@ -143,6 +143,7 @@
 	@import "footer/print";
 	@import "menu/print";
 	@import "promo/print";
+	@import "priorities/print";
 	@import "search/print";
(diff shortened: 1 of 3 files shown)


EricDunsworth:
Btw if it's decided to keep the activity/initiatives in print view instead of hiding them, maybe I could send over a replacement PR to make them look nicer in that view.

Issue #800, opened 2014-09-30
Fonts folder: _base.scss file


shawnthompson:
@nschonni is this file needed in the compiled GCWeb build?

\themes-dist-GCWeb\fonts_base.scss

PR #1113, opened 2016-04-22
Theme: Prevent fonts folder's .scss file from getting copied into dist.

EricDunsworth:
When running ``grunt`` or ``grunt dist`` in GCWeb, ``src/fonts/_base.scss`` currently gets copied into ``dist/fonts/_base.scss``. That file is also present in GCWeb 4.21's release package. This PR will prevent it from getting copied in the first place.

@LaurentGoderre @nschonni @shawnthompson FYI.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -203,7 +203,10 @@ module.exports = (grunt) ->
 			fonts:
 				expand: true
 				cwd: "src/fonts"
-				src: "**/*.*"
+				src: [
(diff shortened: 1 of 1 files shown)


shawnthompson:
weird I thought @LaurentGoderre fixed this in #800

LaurentGoderre:
I did in the minified, not the unminified....

shawnthompson:
let's get this merged than! Thanks!

PR #1178, opened 2016-10-03
Site: Fixed/Updated various broken/outdated links.

EricDunsworth:
Notes:
* Mega menu link updates put GCWeb ahead of Canada.ca and its CDN.
* One of the updates changed link text and a link destination in the report a problem page's privacy statement.

PR #1180, opened 2016-10-04
Site: Changed mega menu HBS files to use i18n variables.

EricDunsworth:
This should be merged in after #1178. Once that happens, this PR might run into merge conflicts. Should be super easy to resolve them if it comes to that.

### DIFF
--- a/site/includes/fullmenu-en.hbs
+++ b/site/includes/fullmenu-en.hbs
@@ -3,7 +3,7 @@
 	<h2 class="wb-inv">{{{i18n "tmpl-topics-menu"}}}</h2>
 	<div class="row">
 		<ul class="list-inline menu" role="menubar">
-			<li><a href="#jobs" class="item">Jobs</a>
+			<li><a href="#jobs" class="item">{{{i18n "jobs"}}}</a>
 				<ul class="sm list-unstyled" id="jobs" role="menu">
 					<li><a href="http://www.jobbank.gc.ca/home-eng.do?lang=eng">Job Bank</a></li>
 					<li><a href="http://www.esdc.gc.ca/en/jobs/opportunities/index.page">Find a job</a></li>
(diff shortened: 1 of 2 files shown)


nschonni:
You'd probably get more value out of this if everything was in the i18n, because then you'd be getting ride of one of the templates and just compiling it to 2 (or more) languages

EricDunsworth:
@nschonni I agree. My main goal with this PR was solely to eliminate the need to re-declare the same top-level mega menu links twice in [the "mini" mega menu](https://github.com/wet-boew/GCWeb/blob/master/site/includes/sitemenucontent.hbs) (via [/site/data/i18n/en.json](https://github.com/wet-boew/GCWeb/blob/master/site/data/i18n/en.json)) and [the "full" mega menu](https://github.com/wet-boew/GCWeb/blob/master/site/includes/fullmenu-en.hbs). That way, if those particular links change again the future, they'll only need to be updated in one spot for per language.

Doing it for the rest of the "full" mega menu's content might take a while (likely means introducing > 156 new variables per language), but I think I'll give it a shot since I see the value in what you're suggesting.

MichaelDalrymple:
Question: Does anyone know the process for informing departments when the mega menu is updated? I'm told that each department needs to hardcode links into their template for non-javascript browswing. Wondering how each dept is informed of their need to update this content. Thanks.

shawnthompson:
@MichaelDalrymple There is a CDN for the mega menu, just like the GoC Activities and Initiatives.
[Travel.gc.ca](https://travel.gc.ca/) is using it. It uses cross domain AJAX to pull it in.

rlambert27:
@shawnthompson I think @MichaelDalrymple means how do we know when to update the hard-coded links that appear when Javascript is disabled if we can't grab them from the CDN.

I know that for the CDTS solution we release new static files every release to account for any GCWeb changes including the mega menu links (http://www.gcpedia.gc.ca/wiki/Centrally_Deployed_Templates_Solution_(CDTS)).

shawnthompson:
@MichaelDalrymple presently there's no notification for departments. Even GCWeb needed to be updated and was notified of any changes.

cc/ @lucas-hay @MaximMartel @masterbee

MichaelDalrymple:
Thanks Shawn. So, this is a problem for us internally and has been raised to my attention a few times now. Any idea on how to get the notification process changed or who is responsible for notifying depts?

lucas-hay:
@MichaelDalrymple, the service desk from Principal Publisher (Service Canada) send out an email anytime there are mega menu updates. We have a list of contacts for each site/department (that we are aware of) that are using the GCWeb templates. We update the CDN, and send an email, as there are a few sites that are not using the CDN. Can you please look me up in GEDS and I will get your contact added to the list.

Issue #972, opened 2015-04-15
Use WET Core footer resources include


LaurentGoderre:
I'm trying to mininize the amount of changes we need to apply across multiple theme. The Core include should be flexible enough to meet GCWeb's need.

hsrudnicki:
and some additional suggestions included in https://github.com/wet-boew/wet-boew/issues/5413 - New Getting Started

hsrudnicki:
@LaurentGoderre : Can be this issue closed as resolved/fixed or if otherwise?  Either way, please update this issue's status, thank you

LaurentGoderre:
Still needed.

hsrudnicki:
Is this still an issue that needs to be resolved or otherwise? Please update this issue's status, thank you

PR #1191, opened 2016-10-24
Update headresources.hbs to support offline

stef52:
Grabs the local version of jquery when offline


LaurentGoderre:
@stef52 thanks for the PR but I would 👎 this based on #972. These file should be dropped in favor of the WET ones.

shawnthompson:
@LaurentGoderre Can you elaborate on this?

lucas-hay:
Discussed at PR review meeting. We will close this PR and not merge. We will leave it up to the users to change to offline resources if need be.

PR #1235, opened 2017-04-05
Update MegaMenu as per SD-2491

MaximMartel:


MaximMartel:
Working example : https://maximmartel.github.io/GCWeb-UpdateMegaMenu/home-en.html | https://maximmartel.github.io/GCWeb-UpdateMegaMenu/home-fr.html

PR #1241, opened 2017-04-28
Theme: Fixed header FIP image's incomplete alt text.

EricDunsworth:
A bilingual FIP image is situated in the top-right of all of GCWeb's pages. Until now, in all standard/server message page templates, the image's alt text was unilingual, which isn't an accurate representation of the text in the image.

This commit adds extra handlebars conditions to make the FIP's alt text bilingual depending on the current page language.

**PS:** The issue this PR fixes likely affects all pre-existing GCWeb implementations (e.g. Canada.ca, CDTS, web applications based on GCWeb, etc...).

### DIFF
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,6 @@
 {{#unless experimental}}
 	<div class="brand col-xs-8 col-sm-9 col-md-6">
-		<a href="https://www.canada.ca/{{language}}.html"><object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg"></object><span class="wb-inv"> {{{i18n "tmpl-gc-sig"}}}</span></a>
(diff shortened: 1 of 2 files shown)


site/includes/brand.hbs
```diff
@@ -1,6 +1,6 @@
 {{#unless experimental}}
 	<div class="brand col-xs-8 col-sm-9 col-md-6">
-		<a href="https://www.canada.ca/{{language}}.html"><object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg"></object><span class="wb-inv"> {{{i18n "tmpl-gc-sig"}}}</span></a>
+		<a href="https://www.canada.ca/{{language}}.html"><object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg"></object><span class="wb-inv"> {{{i18n "tmpl-gc-sig"}}} / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
```
LaurentGoderre: You don't need all this logic. The i18n defaults to English when a translation is not provided so you don't need all these ifs.

EricDunsworth: @LaurentGoderre I don't see how I can simplify that logic any further without producing undesirable results. Only one is/else block is currently in place. Brand.hbs is included in both French/English pages and the FIP differs based on the current page's language.

English pages use a FIP containing "Government of Canada / Gouvernement du Canada", whereas French ones use "Gouvernement du Canada / Government of Canada". So I need to account for the switched order in the text alternative.

If you still believe there's a simpler way of coding it, what would you recommend?

LaurentGoderre: Hmmm, maybe in the future we should have i18n for the combined strings but you're right.

PR #1245, opened 2017-05-01
Theme: Fixed incorrect FIP alt text in arm's length institution pages.

EricDunsworth:
The Government of Canada FIP is used as a sample image throughout arm's-length institution page templates. Until now, it was using "Department name"/"Nom du département" as alt text, which is incorrect.

This commit fixes it by removing the "arms-image-alt" i18n variable, replacing all references to it with "tmpl-gc-sig" and adding extra is/else logic to account for the fact that the FIP is bilingual.

### DIFF
--- a/site/data/i18n/en.json
+++ b/site/data/i18n/en.json
@@ -231,7 +231,6 @@
 	"service-right-panel-contact-heading": "Contact us",
 	"service-right-panel-contact-listitem": "Contact [Institution name] for help",
 	"arms-description": "We are a federal institution that is part of the [portfolio name] portfolio. We operate at arm’s length from the Government of Canada.",
-	"arms-image-alt": "Department name",
 	"comp-dwnld-title": "Document title",
 	"comp-dwnld-size-type": "MB",
 	"comp-dwnld-size-abbr": "Megabyte",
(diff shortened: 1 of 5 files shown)


LaurentGoderre:
@EricDunsworth it's using the FIP as a placeholder image. For example, a tribunal would use it's own FIP graphic and would need to use their deppartment name, no the FIP alt one.

EricDunsworth:
@LaurentGoderre I'm aware of that, but even though the image itself is a placeholder, IMO its alt text should still correspond to the text in the image.

If the placeholder image literally consisted of a Canada flag/coat of arms/etc symbol alongside "Department name", the pre-existing alt text would be perfectly fine. But since the bilingual GoC FIP was used instead, it should use corresponding alt text.

LaurentGoderre:
@EricDunsworth not since this is a template. The current alt text is accurately describing the purpose of the image rather than the text which in this case is irrelevant.

EricDunsworth:
@LaurentGoderre But in this case, the image itself doesn't contain any indication of its purpose. So even though an image of text is being used, IMO it wouldn't necessarily be obvious to sighted users that it's meant to be a placeholder for an institution-specific FIP image. Especially since the exact same GoC FIP image is situated in the top-right of all page templates with proper alt text. IMO it seems very strange to have two images of text on the same page, where one has accurate alt text and the other has different alt text describing an alternate purpose (rather than the text in the image).

Maybe what I'm proposing in this PR isn't the best way of resolving this situation. Maybe it'd be better to replace the image with something that's more in line with the pre-existing alt text and/or revise both the image and alt text.

lucas-hay:
Disscussed at PR review meeting. Taking a different approach for this. Closing for now.

duboisp:
@EricDunsworth for your info check the IA spec 7.9.5.1 page 113

PR #1252, opened 2017-05-12
Theme index: Revised link text to end with page languages.

EricDunsworth:
* Added hidden link text containing " - [Language]" to the end of all page template links. This makes each link's text unique, which can benefit screen reader users that may encounter them out of context.
* Revised the bilingual 404 error and server message page's links to end with " - [Language]/[Language]" (instead of " ([Language]/[Language])") for consistency.
* Related to wet-boew/wet-boew#7990.

### DIFF
--- a/site/pages/index-en.hbs
+++ b/site/pages/index-en.hbs
@@ -7,7 +7,7 @@
 		{ "title": "Home", "link": "https://www.canada.ca/en.html" }
 	],
 	"tag": "GCWeb",
-	"dateModified": "2017-05-04",
+	"dateModified": "2017-05-12",
 	"share": "true"
 }
 ---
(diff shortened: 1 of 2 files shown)


duboisp:
That introduce a lot a redundant content.

What about if the English page only contain English link and the French page only contain French link? 

If navigating between both language version is needed, we can simply add a paragraph under the h1 like

The above template and plugins are also [available in French](http://wet-boew.github.io/themes-dist/GCWeb/index-fr.html).
Tous les gabarit ainsi que les plugiciels ci-dessous son aussi [disponible en anglais](http://wet-boew.github.io/themes-dist/GCWeb/index-en.html).

EricDunsworth:
@duboisp I understand your standpoint, but FWIW, all pre-existing content is already somewhat redundant and potentially confusing to certain users.

Multiple links with exactly the same link text are currently pointing to different pages. If a screen reader user attempts to navigate GCWeb's index page by link (which will list out links out of context), they could potentially visit the first "Theme landing page" link that gets listed without realizing that the second "Theme landing page" link is in fact pointing to a different page.

By adding visually-hidden languages at the end of those links, their link text would become unique out of context. Only screen reader users would encounter the hidden link text. Sighted users wouldn't see it (since the headings/layout already provide visual context to what the links represent).

I should also note that wet-boew/wet-boew#7990 does the exact same thing and was recently merged in. My goal with this PR and similar ones I sent to WET's other theme repos was to bring everything in line with the main WET repo prior to v4.0.25's release. [WET 3.1](http://wet-boew.github.io/wet-boew-legacy/v3.1/demos/theme-wet-boew/index-en.html#en) and [3.0](http://wet-boew.github.io/wet-boew-legacy/v3.0/demos/theme-gcwu-fegc/index-en.html#en) also used to do something similar (but arguably more annoying since all of their links were prefixed with hidden languages).

IMO it's a better practice to keep all links for both languages on the same page, since it's more consistent with WET's other theme repos. In all of WET's other theme repos, theme index pages are also the only places that link to multilingual page templates.

duboisp:
@EricDunsworth I was only suggesting to have two unilingual pages instead of having two a bilingual page containing the same list of anchors with the only difference of having translated text. 

Currently, I am fine with either solution because it is currently manageable like that.

> IMO it's a better practice to keep all links for both languages on the same page, since it's more consistent with WET's other theme repos.

I don't agree about taking a decision, for this particular case, based on consistency. That kind of decision should be done to improve the usability of providing a theme demo. A theme, like GCWeb, do not have the same "user needs" as the [OGPL theme](https://github.com/wet-boew/theme-ogpl) simply because they don't have a comparable activity (GCWeb +1500 commits vs OGPL ~360 commits). :wink:

PR #1260, opened 2017-06-15
Footer: Made about list compatible with minified HTML.

EricDunsworth:
The "About this site" list's layout used to rely on space/line break/tab characters being present in-between li elements in the list's HTML markup. But that didn't lend itself to pages that use minified HTML. The horizontal spacing between the list's items looked even in unminified pages, but looked off in minified ones.

This commit changes the list's SCSS to always use an even amount of spacing, regardless of whether HTML has been minified.

**Screenshots:**
* **Before (minified):** before_minified
* **Before (unminified):**  before_unminified
* **After (minified and unminified):**
after_both

**PS:**
PR #1259 should be merged in before this one. Once that happens, this PR will most likely run into merge conflicts, so I'll rebase it afterwards.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -21,6 +21,9 @@ $wb-bc-background-color: #f5f5f5; // Outer background colour (WET override).
 //Heading links on landing pages
 $heading-link: 16px;
 
+// "About this site" footer links
+$footer-about-li-horizontal-spacing: .7em;
+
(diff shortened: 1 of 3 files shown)


EricDunsworth:
@lucas-hay FYI I've rebased, retested and linted this PR. It should be good to merge now.

PR #1269, opened 2017-07-12
Added RAP CSS fix - Hide textarea and hide Other checkbox

MaximMartel:
Moved the RAP stuff from defaults.scss to /template/_base.scss as well 
My fix is live 208 to line 222

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -303,39 +303,3 @@ aside {
 		}
 	}
 }
-
-/* Report a problem Thank you message*/
-.gc-rprt-prblm-thnk {
-	padding-bottom: 25px;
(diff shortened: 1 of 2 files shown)


MaximMartel:
@lucas-hay @masterbee

MaximMartel:
please see gh-pages : https://maximmartel.github.io/GCWeb-RAPFix/home-en.html to see the fix

shawnthompson:
Why are you hiding this CSS and not just removing the HTML from the source code?

MaximMartel:
We will remove the HTML later on.

shawnthompson:
Why not remove it now?

PR #1279, opened 2017-08-04
Fix constrast issue with the overlay close button

### LINKED ISSUE #1267: Overlay (Middle/Full screen and left right) - Default close button contrast issue
To Reproduce:

1. Go to [GCWeb Overlay working example](http://wet-boew.github.io/themes-dist/GCWeb/demos/overlay/overlay-en.html)
2. Open one of the following example
   * Middle screen overlay
   * Full-screen overlay
   * Right-panel
   * Left-panel
3. Remark that foreground and the background color of the close button is both white.

overlay-close-btn-contrast-issue p ng

duboisp:
fix #1267 

Working example: http://universallabs.org/wet/labs/fix-1267/unmin/demos/overlay/overlay-en.html

### DIFF
--- a/src/menu/_base.scss
+++ b/src/menu/_base.scss
@@ -139,6 +139,7 @@
 
 	.modal-body {
 		background: #0e4164;
+		margin-bottom: -3em;
 		margin-left: 0;
 		padding-bottom: 0;
 		padding-left: 0;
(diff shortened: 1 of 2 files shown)


EricDunsworth:
Looking into this further, it looks like the issue this PR is trying to fix was introduced via wet-boew/wet-boew#8087.

One of its changes added this into the lightbox plugin's base SCSS:
```
.modal-footer {
+	background: $clrWhite;
+	margin-top: 0;
+}
```

Since the overlay plugin also uses the ``modal-footer`` class, that change caused the mobile menu footers in WET/GCWeb's themes to become white. I believe it'll also affect every other WET theme (e.g. theme-base, theme-ogpl, etc...) the next time they're rebuilt. So it might be better to try fixing the root of the problem in WET itself rather than adding separate overrides into every theme.

PR #1282, opened 2017-08-14
Share this page link: Use 100% width in extra small view and under.

### LINKED ISSUE #1268: "Share this page" link no longer using 100% width in extra small views.
In most of GCWeb's page templates, the share widget plugin is used to inject a "Share this page" link at the bottom of the content area. In small view and over, that link is floated to the right. In GCWeb 4.0.25 and under, it uses a width of 100% in extra small views.

Unfortunately, the aforementioned 100% width behaviour relied on a CSS selector that was recently removed in wet-boew/wet-boew#8076. The specific selector GCWeb relied on was formerly situated on [line 554 of src/base/bootstrap-overrides/_base.scss](https://github.com/wet-boew/wet-boew/pull/8076/files#diff-93656052841a8c39fc075d98b607fb71L554).

I can think of 3 potential ways of resolving this, but all have their own drawbacks...

**Possible solutions:**
* Revert all or part of wet-boew/wet-boew#8076. The overrides it removed were first introduced in early 2014 via wet-boew/wet-boew#4206 to fix wet-boew/wet-boew#4191. So it can probably be safely assumed that in practically all WET 4.0.x implementations until now, content was coded and tested with those overrides in mind. The overrides probably shouldn't of been introduced in the first place, but by now they've been around for so long that suddenly removing them might cause unexpected issues across pre-existing content pages.
* Add the ``col-xs-12`` class to the share widget's parent ``div`` element. The downside of this would be that it would necessitate an HTML markup change in all pre-existing GCWeb implementations to retain a layout that was already working before.
* In [src/share/](https://github.com/wet-boew/GCWeb/blob/master/src/share), add a "_screen-xs-max.scss" file containing a selector that sets ``float: none !important;`` on any ``pull-right`` classes that are children of the ``pagedetails`` class. The drawbacks of this are that it's kind of hacky and can't explicitly target the share widget's parent ``div``.

@lucas-hay @thomasgohard Thoughts?

**Screenshots:**
* **GCWeb 4.0.25 and under:**
gcweb-share-before
* **GCWeb 4.0.26-development (as things stand right now):**
gcweb-share-after

EricDunsworth:
Fixes #1268.

### DIFF
--- a/src/share/_screen-xs-max.scss
+++ b/src/share/_screen-xs-max.scss
@@ -1,0 +1,10 @@
+/*
+ * Share/feedback-specific overrides (extra-small view and under)
+ */
+
+// Disable floating on the share button's container to maximise its width.
+.pagedetails {
+	.pull-right {
+		float: none !important;
+	}
+}
--- a/src/theme.scss
+++ b/src/theme.scss
@@ -67,6 +67,7 @@
 	@import "views/screen-xs-max";
 	@import "banner/screen-xs-max";
 	@import "pagination/screen-xs-max";
+	@import "share/screen-xs-max";
 	@import "toc/screen-xs-max";
 }
 


EricDunsworth:
@nschonni While prepping this PR I was torn on whether I should word it in a way that'd outright close the issue or keep it open. In the end I felt like since the layout issue itself is resolved by this PR, it'd make more sense to word it to outright close the issue. IMO the HTML markup change is a supplemental side issue.

Anyway, I was going to offer to modify the commit and OP to say something along the lines of "Fixes part of #1264" if you felt like that'd be for the best, but it seems that @lucas-hay merged in this PR before I got a chance to reply.

Either way, I'll send a subsequent PR soon to deal with the markup change. This PR was originally going to consist of both its fix and the markup change, but only the layout fix is critical for 4.0.26 (which is slated for release tomorrow). I decided to hold off until 4.0.27 for the markup change.

nschonni:
Yup, the other one you can use to link stuff without the auto-close is `Related`

PR #1302, opened 2017-10-19
Theme: Added nav/h2 elements to ToCs.

EricDunsworth:
Related to wet-boew/wet-boew#8177.

### DIFF
--- a/site/pages/institutional-service-performance-en.hbs
+++ b/site/pages/institutional-service-performance-en.hbs
@@ -12,14 +12,14 @@
 ---
 {{>byline}}
 <p>This report covers the 2015 to 2016 fiscal year. A fiscal year runs from April 1 to March 31.</p>
-<section>
+<nav role="navigation">
   <h2>On this page</h2>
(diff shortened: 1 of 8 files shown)


site/pages/institutional-service-performance-en.hbs
```diff
@@ -12,14 +12,14 @@
 ---
 {{>byline}}
 <p>This report covers the 2015 to 2016 fiscal year. A fiscal year runs from April 1 to March 31.</p>
-<section>
+<nav role="navigation">
```
duboisp: Is it required to duplicate the role? By definition the [```<nav>``` element](https://www.w3.org/TR/html52/sections.html#ref-for-represent-13) imply the aria role navigation.

Idem for all other instance.

EricDunsworth: Check out https://github.com/wet-boew/wet-boew/pull/8177#discussion_r146637014 for my response to this.

PR #1306, opened 2017-10-27
Theme: Set font style to normal on cite/dfn/em/i elements.

EricDunsworth:
Related to wet-boew/wet-boew#7631.

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -2,6 +2,10 @@
 	Global Default Body styling
 */
 
+%gcweb-font-style-normal {
+	font-style: normal;
+}
+
 %gcweb-pagedetails-children-margin-top {
 	margin-top: .5em;
 }
@@ -25,6 +29,22 @@ a {
 	&.no-undrln {
 		text-decoration: none;
 	}
+}
+
+cite {
+	@extend %gcweb-font-style-normal;
+}
+
+dfn {
+	@extend%gcweb-font-style-normal;
+}
+
+em {
+	@extend %gcweb-font-style-normal;
+}
+
+i {
+	@extend %gcweb-font-style-normal;
 }
 
 // scss-lint:disable QualifyingElement


EricDunsworth:
@duboisp This PR's solution is by no means perfect, but IMO it's a much better alternative than to not code those elements at all just because they visually-appear as italicized.

Btw I noticed your comment was in the form of a change request, but no specific changes were requested. Did you have any concerns with my SCSS markup? Even though you may disagree with end result of my changes, I believe what I did coding-wise should be fine if it's eventually decided to merge this PR into the repo.

CC @lucas-hay @masterbee @shawnthompson @stephendonnelly2 @DK-TBS

lucas-hay:
@DK-TBS this would solve [#7631](https://github.com/wet-boew/wet-boew/issues/7631), but we are looking for a decision from the C&IA group before we merge.

DK-TBS:
Don't merge it.  Talked to the group here, and we've realized that we can't just make the change without warning all of the departments first, somehow.  And we do have a mechanism for doing that, so we'll try and take that approach.  We'll give them some better guidance around how to deal with citations and emphasis etc, and then give them some time to prepare before we make any changes to the central stylesheets.  

It's too bad, I was really hoping to find a quick fix for this.  But clearly it's something that we'll have to approach more carefully.

In the meantime, authors should put the priority on proper semantic markup.  If using HTML properly causes them to have italics and that's against the style guide, then screw the style guide.

Sorry guys.  Thanks for your efforts on this.  Appreciate you trying to find fixes with me  :)

stephendonnelly2:
Sorry but "screw the style guide" is not something we can ever say to a client or follow. If you can provide an official statement and/or update the style guide, that is what we will go by.

MichaelDalrymple:
I've been watching this thread. I've long been a proponent that the Style Guide is misguided on their blanket statement to not allow any italicized text, as it breaks HTML rules and common brower behaviours.

That said, if the Style Guide stands, and if the Canada.ca CSS is to be changed to not display italicized text as italics, then it shouldn't be made to look normal. If anything, it should be bolded as the whole point of italicizing the text is to differentiate it from the text before/after it. Making it normal won't accomplish this. Has this been discussed?

DK-TBS:
I know, I'm being flippant.  Here's a clearer expression of the idea:  that rule was added to the style guide without proper thought having been put into the impact on the HTML and CSS on Canada.ca.  This has led to an untenable situation where the style guide and stylesheets do not comfortably coexist.  We are aware of the problem and intend to fix it.  Until we come up with a fix, we encourage people to put the priority on proper semantic HTML.  If coding HTML properly means that some elements render in italics against the style guide, that's okay for now.  The proper use of HTML is more important.  

We'll be taking another look at this problem soon.  And before that happens, I'll be working on getting the style guide gang to add a note in there on the style guide itself.

DK-TBS:
@MichaelDalrymple - It's been discussed, but no resolution yet.  Those conversations will have to happen as we come up with better guidance around how people should handle this.

MichaelDalrymple:
Dave...I've been pushing another look since version 1.5 of the Style Guide. If you make more headway than I did, here are some ideas to propose: The style guide should instead lay out some instructions on when and when not to use italics, rather than banning them completely (ie. don't use them in headings, linked text, in consecutive sentences, for styling purposes). Issuing instructions to avoid or use alternatives for the <em> and <i> tags would be much simpler than the blanket ban, especially since tags like <cite>, <address> and <dfn> are there (as per my message above) to differentiate the inner text from the surrounding text. This has been an ongoing debate within my dept, especially from our policy people who use <cite> in pretty much every paragraph when referencing other policies, since this rule has come out, one that is getting harder to defend as a best practice for all.

MichaelDalrymple:
One final note: If there's no resolution on this, then it should be removed from the Style Guide immediately until a resolution is found.

EricDunsworth:
@DK-TBS Thanks for clearing things up :)!

Btw should I go ahead and close this PR? The impression I'm getting now is that disabling italics outright without adding an alternate visual indicator (such as a different font/etc) is officially no longer being considered in any capacity. If that's the case, then this PR no longer serves any purpose.

@MichaelDalrymple I agree that it wasn't very ideal to only disable italics without implementing an alternate visual indicator. But my original intent with this PR was just to disable italics as a stop-gap measure to allow developers to resume coding their pages with the ``cite``, ``dfn``, ``em`` and ``i`` elements until a more concrete solution or decision came out of wet-boew/wet-boew#7631. I didn't personally agree with the change I was making, but I felt like it'd be the best compromise given the status quo at the time.

duboisp:
@EricDunsworth as this is currently a hot topic, I recommend you to wait until the next PR review meeting before to close it.

DK-TBS:
@EricDunsworth - Feel free to close the PR, or leave it open just to keep it on everybody's radar.  I'm not sure that everybody on our end completely understands the issue, so I'm trying to get the conversation going again in the hopes of getting some solid decisions out of my team.  We've got to either adjust the style guide or adjust the Canada.ca CSS.  I don't want to see this problem dangling any longer, so I'm going to see what I can do on this end.

@MichaelDalrymple - I hear you, brother.  I'm not sure that the team here is willing to consider allowing even limited italics, but I'll make sure they remember that it is an option for solving this problem.

duboisp:
Wait for the resolution in wet-boew/wet-boew#7631

EricDunsworth:
Going to close this out since #1306 has been resolved.

The [Canada.ca Content Style Guide's italics section](https://www.canada.ca/en/treasury-board-secretariat/services/government-communications/canada-content-style-guide.html#wp4-2) was recently updated to condone the use of italics under certain circumstances.

PR #1308, opened 2017-11-03
Exit link feature - initial commit

duboisp:
When you need to provide a "quick escape" or a "leave this website" feature at your user because of the sensitivity of your content.

Working example:

* http://universallabs.org/wet/labs/exitlink/demos/exitlink/exitlink-en.html
* http://universallabs.org/wet/labs/exitlink/demos/exitlink/exitlink-fr.html

Documentation:

* http://universallabs.org/wet/labs/exitlink/demos/exitlink/exitlink-doc-en.html


lucas-hay:
@DK-TBS, can you run the working example link above by the C&IA group for feedback and let us know what you think?

DK-TBS:
@lucas-hay

Ideally we'd like to test it, to ensure that people notice the buttons and understand what they mean, but that would hold things up pretty badly.  So instead we'll charge ahead with our best guesses and try to squeeze in a little usability testing later on.

Talked it over with the crew here, and we think the current approach is pretty decent, it could work well.  But we would like to change the label, to make it more descriptive:

"Leave this page and hide your visit"

"Quittez cette page et masquez votre visite"

Also, we'd recommend hard-coding it so that it always goes to weather.gc.ca.  We think weather is a good choice for two reasons:

* Nobody in the GoC has to wonder whether it's okay to drive traffic to some third-party commercial entity (Google vs Bing, for example)

* The weather.gc.ca pages look similar to other pages on Canada.ca, so when a visitor uses the button to make a quick exit, the transition isn't too jarring.  Somebody catching a glimpse over the user's shoulder would be less likely to notice that they've gone to a completely different website.

duboisp:
Todo:
* [ ] Find a way to better integrate with GCWeb for those rare use case. Like loading from a dependency.

shawnthompson:
I'd like to see more research on this case. More and more departments are using it and I think it could be done better. 

Would also like to see some type of promotion of the use of this somewhere. Maybe Women and Gender Equality Canada might want to do a small campaign on the use of this button on the websites.

PR #1312, opened 2017-11-14
Build: Replaced JSCS and JSHint with ESLint.

EricDunsworth:
* Updated package.json dependencies.
* Replaced the "jshint" and "jscs" tasks with an "eslint" task.
* Setup the "estlint" task to inherit its configuration settings from the wet-boew dependency's .eslintrc.json and .eslintrc.ci.json files.
* Updated src/theme.js to pass linting.
* Modified the "test" task to reference the "eslint" task.
* This is mostly a port of wet-boew/wet-boew#7456, along with some of wet-boew/wet-boew#6514 and wet-boew/wet-boew#7948's updates.


Gruntfile.coffee
```diff
@@ -740,19 +726,18 @@ module.exports = (grunt) ->
...
-				files: "<%= jshint.lib_test.src %>"
-				tasks: [
-					"build"
+			js:
+				files: "<%= eslint.all.src %>"
+				tasks: "js"
+
+		eslint:
+			options:
+				configFile: if process.env.CI == "true" then ".eslintrc.ci.json" else ".eslintrc.json"
```
EricDunsworth: @nschonni @LaurentGoderre Any thoughts on how best to go about referencing ESLint's settings?

Currently this line is referencing copies of wet-boew/wet-boew's .eslintrc.ci.json and .eslintrc.json files. But an alternate way would be to not include copies of those files in this PR and just reference the originals from the "lib/wet-boew" folder.

It seems like GCWeb and the other theme repos are currently using a mix of both ways:

* JSHint settings were previously being referenced from "lib/wet-boew/.jshintrc". But since that file was removed at some point from wet-boew/wet-boew, it's been over a year since the last time GCWeb was able to actually inherit those settings.
* Sass-lint is setup to use a copy of .sass-lint situated in the GCWeb repo. GCWeb and wet-boew/wet-boew's versions of that file are exactly identical.

.eslintrc.json
```diff
@@ -0,0 +1,32 @@
+{
+  "extends": "jquery",
```
nschonni: From your question below, maybe: `extends: lib/wet-boew/.eslintrc.json`
Other large projects sometimes just create a separate ESlint rules repo that can be referenced. EX https://github.com/jquery/eslint-config-jquery

Gruntfile.coffee
```diff
@@ -740,19 +726,18 @@ module.exports = (grunt) ->
...
-			lib_test:
-				files: "<%= jshint.lib_test.src %>"
-				tasks: [
-					"jshint:lib_test"
-				]
-			source:
-				files: "<%= jshint.lib_test.src %>"
-				tasks: [
-					"build"
+			js:
```
nschonni: Change is probably fine, but I don't recall if there was another reason for this split of tasks before

.eslintrc.ci.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ".eslintrc.json",
+  "rules": {
+    "linebreak-style": 2
+  }
+}
```
nschonni: Not sure if the line ending should be added

Gruntfile.coffee
LaurentGoderre: The split of task was also about line ending wasn't it? Or maybe whitespace validation?

nschonni: Actually I think it was for a faster turnaround when someone was adding/improving tests. That way the whole JS didn't recompile when only working on testing changes. If no ones using it then it can go away though

.eslintrc.ci.json
EricDunsworth: That file was copied from the wet-boew/wet-boew repo, so it should probably be addressed there. I'll likely be updating this PR soon to remove that file.

.eslintrc.json
EricDunsworth: Ok. I'll modify this PR to go in that direction. Btw that will result in this file's removal, since I initially copied it from wet-boew/wet-boew (without any changes).

Gruntfile.coffee
EricDunsworth: I'm not sure what it was being used for, but it wasn't for line endings (that's covered by .eslintrc.ci.json - which only Travis-CI is configured to use via wet-boew/wet-boew#7948).

In the wet-boew/wet-boew repo, it looks like it was originally introduced in wet-boew/wet-boew@fea3ff52c3721d273d4b9e172110565e7dd84eb0 (very first v4.0 commit) and eventually removed in wet-boew/wet-boew#6514. I guess the removals hadn't previously been ported to the theme repos.

Gruntfile.coffee
EricDunsworth: I'm gonna update this PR to modify this line to point to lib/wet-boew's versions of those files and remove my copies of those files.

EricDunsworth:
FYI I've updated this PR to remove its copies of .eslintrc.json/.eslintrc.ci.json and made it reference those files from lib/wet-boew. Also updated a few things in its commit message.

@LaurentGoderre @nschonni I'll port this to the other themes once both of you have approved it or once it's merged in. Just want to make sure its changes are finalized before I use it as a baseline to create the ports.

Btw there might also be value in modifying the theme repos to inherit their sass-lint settings from lib/wet-boew.

Gruntfile.coffee
EricDunsworth: Btw I could potentially revise this to make the watch task auto-run eslint when JS files get modified, but I don't think it'd be worthwhile to do as a part of this PR since wet-boew/wet-boew doesn't currently do that. So I don't want to inadvertently put the themes ahead of it in that respect.

Gruntfile.coffee
```diff
@@ -141,6 +141,7 @@ module.exports = (grunt) ->
 		"test"
 		"INTERNAL: Runs testing tasks except for SauceLabs testing"
 		[
+			"eslint"
```
nschonni: Oh, I guess it wasn't running JSHint at all before

EricDunsworth:
Just replicated this PR throughout the other themes (see above). They're mostly identical to this one.

Main differences are:

* All of those themes had their own ".jshintrc" files. I removed them.
* None of those themes use "watch" tasks. I didn't introduce that task into to any of them.
* Some of those themes' Gruntfiles didn't contain any "jshint" nor "jscs" tasks. So some of those PRs' diffs won't show any indication of those tasks being removed.

LaurentGoderre:
I think this id good to go? @nschonni

EricDunsworth:
FYI nothing's changed in my latest 2 pushes... I wasn't paying attention to the branch I was working in when I rebased my usability theme PRs. I initially inadvertently rebased and force-pushed this PR.

PR #1317, opened 2017-12-01
Cleaned up css and classes for margin in right column, added missing …

lucas-hay:
Cleaned up css and classes for margin in right column, added missing row to institution contact page, changed col definition on dept page


duboisp:
Looks good, assuming that eventually new changes will be added to remove any clearfixes.

PR #1320, opened 2017-12-05
Institutional template - Use proper grid when displaying Follow us content block

duboisp:
@lucas-hay see the code change

Also, the handlebar conditional need to be a negative instruction, then you will be able to remove the "else".

### DIFF
--- a/site/includes/follow.hbs
+++ b/site/includes/follow.hbs
@@ -1,8 +1,9 @@
 {{#contains "theme, topic, gcaudience" page.pageType }}
-<section class="followus">
 {{else}}
-<section class="followus pull-right">
+<div class="row">
+<div class="col-xs-12 col-md-auto pull-right">
 {{/contains}}
+<section class="followus">
 	<h2>{{{i18n "so-title"}}}</h2>
 	<ul>
 		<li><a href="#facebook" class="facebook wb-lbx"><span class="wb-inv">Facebook</span></a></li>
@@ -38,3 +39,7 @@
 		</ul>
 	</div>
 </section>
+{{#contains "theme, topic, gcaudience" page.pageType }}
+{{else}}
+</div></div>
+{{/contains}}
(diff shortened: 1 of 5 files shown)


duboisp:
see http://universallabs.org/wet/labs/pr1320/unmin/institution-arms-en.html

lucas-hay:
@duboisp 
Any idea how to deal with the float right (inside follow.hbs) in small view when we remove the following block of code from _base.css?

```xml

 +/* temporarily adding deleted WET styles back to help with right colum issues in GCWEB */
 +
 +/**
 +  *	Disable pull-right/left on grid columns that do not match the current
 +  *	breakpoint
 +  */
 +@media (max-width: $screen-xs-max) {
 +	.pull-left,
 +	.pull-right {
 +		&[class*=col-sm],
 +		&[class*=col-md],
 +		&[class*=col-lg],
 +		&[class*=followus] {
 +			float: none !important;
 +		}
 +	}
 +}
 +
 +@media (min-width: $screen-sm-min) and (max-width: $screen-sm-max) {
 +	.pull-left,
 +	.pull-right {
 +		&[class*=col-md],
 +		&[class*=col-lg],
 +		&[class*=followus] {
 +			float: none !important;
 +		}
 +	}
 +}
 +
 +@media (min-width: $screen-md-min) and (max-width: $screen-md-max) {
 +	.pull-left,
 +	.pull-right {
 +		&[class*=col-lg],
 +		&[class*=followus] {
 +			float: none !important;
 +		}
 +	}
 +} 
```

duboisp:
we can add ```col-xs-12```, that will remove the pull-right effect

lucas-hay:
I tested and it does not work on the inner pull right.

duboisp:
@lucas-hay there is an issue with the ```col-md-auto``` it look like it do get overwritten with the media query when using ```col-xs-12```.

I will take a look later this evening.

duboisp:
@lucas-hay see http://universallabs.org/wet/labs/pr1320/unmin/institution-arms-en.html 

FYI - that working example to do not include the work around that you mentioned previously on comments. I assume the removal of that work around will be made in a future version of gcweb.

PR #1321, opened 2017-12-05
replace clearfix with equal heights and cleaned up audience page

lucas-hay:
replace clearfix with equal heights and cleaned up audience page


lucas-hay:
Thanks @duboisp . I did this one separately as there was a couple more items that just the clear fix. I will be submitting one large PR with all clearfix's removed from all templates.

duboisp:
PR #1323 submitted to fix the clearfix in the JSON manager

PR #1322, opened 2017-12-06
removed all empty clearing divs (clearfix and clr-xxx-xxx classes)

lucas-hay:
The same end result of the empty clearing divs ("clearfix" class" was replicated and replaced by using equal heights. 
There were also various other instances of clr-xxx-xxxx classes that really served no purpose so were removed.
@duboisp please have a look and let me know what you think

### DIFF
--- a/site/includes/corp-info.hbs
+++ b/site/includes/corp-info.hbs
@@ -15,12 +15,11 @@
 			</ul>
 		</div>
 		<div class="col-md-8">
-			<div class="row">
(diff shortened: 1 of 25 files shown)


lucas-hay:
@duboisp OK sounds good. (I was only looking at gcweb templates in this PR)
JSON manager is your baby so I'll leave it to you.
If you want to get it in for this release, we can merge tonight if it's a quick fix for you.

PR #1325, opened 2017-12-06
removed temporary css for pull right fixes in right column

lucas-hay:
-removed temp styles
-added col-xs-12 to negate the pull right in small view
-removed some extra/unnecessary bottom margin on "service and information" section on certain pages


lucas-hay:
@duboisp please have a look and let me know if this is what you had in mind

lucas-hay:
@duboisp I just reviewed the code again and I did not specify col-xs-12 on the "service and info" column, only columns with "pull right".
Are you seeing something that I am not?

duboisp:
@lucas-hay I think that don't matter for those use case, but if you have "pull-left" and the block do not take the full-width, then the next content can be move to the right of that element. So in some circumstance, it could be valuable to add "col-xs-12" to column that have "pull-left".

As I said, I don't think that would have an impact as there is a minimum of 5 doormat link (as per IA spec 1.3.2) under the Service and info for the topic page.

lucas-hay:
@duboisp gotcha. Sorry I misunderstood at first. I thought you were saying I shouldn't specify it. I get it now. I think we are good in these scenarios because the content above and below "service and info" is specified as col-xs-12, so there should never be anything that can float up beside it.

duboisp:
FYI - we would need to find some time and update the wet-boew styleguide about reflowing content around element. The style guide are not showing the right method.

PR #1328, opened 2018-02-08
Datalist JSON suggestion - initial commit

duboisp:
Load suggestions from an array of string in a JSON file.

Working example: http://universallabs.org/wet/labs/datalist-json-suggest/demos/suggest/suggest-en.html


duboisp:
Research for other third party solution and evaluate it against the requirement.

Like: https://a11y.nicolas-hoffmann.net/autocomplete-list/

duboisp:
The concern that was raised at the technical review was:
* Why to have a custom solution instead of re-use an existing one.
* The risk it's API are not scalable and can prevent future improvement.

duboisp:
There is a new project created to have an improved datalist. The improved datalist created through the project might be released in wet-boew instead of being in this theme. Meanwhile, this plugin is needed for showing suggestion with the new GC search.

Project status and research are / would be available through the [wet-boew project documentations](http://wet-boew.github.io/wet-boew-documentation/projects-en.html)

PR #1329, opened 2018-02-14
Theme: Reference SVG logos with img element instead of object.

EricDunsworth:
* Port of wet-boew/wet-boew#8282.
* Related to wet-boew/wet-boew#8276.
* Modifies SCSS to increase the bottom margins on the FIP images to replicate the empty space that was previously appearing below their object element containers.

### DIFF
--- a/src/banner/_base.scss
+++ b/src/banner/_base.scss
@@ -27,10 +27,16 @@ header {
 			}
 		}
 
-		object,
-		img {
+		// Needed for backwards compatibility with GCWeb 4.0.27 and below's HTML markup.
+		// Remove object references from all of this theme's logo selectors in GCWeb 4.1+.
+		img,
+		object {
 			height: auto;
 			max-height: 40px;
+		}
+
+		img {
+			margin-bottom: .375em;
 		}
 	}
 }
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,6 @@
 {{#unless experimental}}
 	<div class="brand col-xs-8 col-sm-9 col-md-6">
-		<a href="https://www.canada.ca/{{language}}.html"><object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg"></object><span class="wb-inv"> {{{i18n "tmpl-gc-sig"}}} / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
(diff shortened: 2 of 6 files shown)


src/banner/_base.scss
```diff
@@ -27,11 +27,17 @@ header {
 			}
 		}
 
-		object,
-		img {
+		// Needed for backwards compatibility with GCWeb 4.0.27 and below's HTML markup.
+		// Remove object references from all of this theme's logo selectors in GCWeb 4.1+.
+		img,
```
nschonni: Minor, if you keep the order of tags the diff will be minimized

duboisp:
There is some difference, like anti-aliasing, that is different before and after. Review this change would remain valid against the content and IA specification

src/banner/_base.scss
EricDunsworth: @nschonni I had thought about that while prepping this and its sister PRs. But since I was already inserting a comment nearby that downplayed the use of the ``object`` element, I felt like it'd also be worthwhile to take the opportunity to place it below ``img`` in the selector to further de-emphasize it. Plus it's more alphabetical this way :).

EricDunsworth:
@duboisp I just did a comparison of [GCWeb's content page template](http://wet-boew.github.io/themes-dist/GCWeb/unmin/content-en.html) before and after this PR's changes:

* **Firefox:** FIP and symbol images render slightly differently. They seem to be off by a fraction of a pixel. So they may appear to be shifted to the left or right by 1 pixel. The images themselves are still rendering within the same dimensions as before, but blurring caused by anti-aliasing appears to be shifted slightly.
* **IE11:** The FIP looks exactly the same as before. But below the content area's H1 heading, all content appears to be shifted upwards/downwards by a pixel. As a result, the symbol renders a tiny bit differently. This seems to be caused by how IE11 is interpreting the ``em`` units in the FIP's bottom margin. If I increase the margin by ``.005em``, everything will look exactly the same as before in IE11 without affecting Firefox/Opera. So I might force push an updated commit to address that.
* **Opera:** Everything looks exactly the same as before.

**Screenshots:**
(not very obvious without opening the screenshots in separate tabs)

* **Firefox:** 
  * **FIP:** 
    * **Before:** 
firefox-fip-before
    * **After:** 
firefox-fip-after
  * **Symbol:** 
    * **Before:** 
firefox-symbol-before
    * **After:** 
firefox-symbol-after
* **IE11:**
  * **Symbol:** 
    * **Before:** 
ie11-symbol-before
    * **After:** 
ie11-symbol-after

EricDunsworth:
Just to follow-up regarding increasing bottom margins to better accommodate IE11... after some more testing I've decided against it. The increased margins become noticeable when zooming in other browsers. Plus the difference isn't obvious without doing direct head-to-head comparisons.

duboisp:
All good for the content and IA spec.

PR #1330, opened 2018-03-01
WIP: Replace bower with npm

LaurentGoderre:
Deprends on wet-boew/wet-boew#8305


package.json
```diff
@@ -29,13 +29,18 @@
   ],
   "author": "",
   "license": "MIT",
+  "dependencies": {
+    "fast-json-patch": "github:wet-boew/JSON-Patch#wb1.0+1.1.4",
+    "jsonpointer.js": "^0.4.0",
+    "wet-boew": "LaurentGoderre/wet-boew#themes-npm"
```
LaurentGoderre: Replace this with upstream WET when wet-boew/wet-boew#8305 is merged

Gruntfile.coffee
```diff
@@ -183,8 +176,10 @@ module.exports = (grunt) ->
 		# Metadata.
 		pkg: @file.readJSON "package.json"
 		themeDist: "dist/<%= pkg.name %>"
-		jqueryVersion: @file.readJSON "lib/jquery/bower.json"
-		jqueryOldIEVersion: @file.readJSON "lib/jquery-oldIE/bower.json"
+		jqueryVersion: grunt.file.readJSON(
```
nschonni: I did something "fancy" here in the upstream, but we could always just hardcode it like the OldIE one

package.json
```diff
@@ -29,13 +29,18 @@
...
+  },
   "devDependencies": {
-    "assemble-contrib-i18n": "~0.1.2",
+    "assemble-contrib-i18n": "0.1.*",
```
nschonni: Not sure I get this change. Wouldn't these be equivalant, or is the `~` a problem because it hasn't hit a major version yet?

script/setup
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm install -g bower grunt-cli && npm install
+npm install -g grunt-cli && npm install
```
nschonni: Is bower still needed for because of the `prepare` script in WET?

package.json
LaurentGoderre: I'm just making it consistent with WET core

script/setup
LaurentGoderre: Oh yeah it is for core.....But we can change core to not use it at all no?

Gruntfile.coffee
LaurentGoderre: This works without hardcoding so it makes it easier to maintain

LaurentGoderre:
@duboisp someone needs to a[pply this to all the other supported themes.

PR #1344, opened 2018-04-17
Build: Autoprefix to PostCSS conversion

EricDunsworth:
* Updated package-lock.json file
* Gruntfile:
    * Renamed "autoprefixer" task to "postcss"
    * Dropped autoprefixing support for older browsers (e.g. IE9-10, etc...)
* Dependencies:
    * Replaced "grunt-autoprefixer" with "grunt-postcss"
    * Added "autoprefixer"
    * Increased minimum "grunt-sass" version to 2.1.0

Port of wet-boew/wet-boew#8340. Credit goes to @nschonni for the original PR.


duboisp:
Test page: http://universallabs.org/wet/labs/gcweb-pr1344/index-en.html

EricDunsworth:
Looks like Travis-CI failed to build somewhere along the way...
```
Running "sass:all" (sass) task
>> Error: File to import not found or unreadable: magnific-popup/src/css/main.
>>         on line 7 of node_modules/wet-boew/src/plugins/lightbox/_base.scss
>>         from line 126 of node_modules/wet-boew/src/base/_wet-boew.scss
>>         from line 5 of node_modules/wet-boew/src/base/_ie8-wet-boew.scss
>>         from line 14 of src/ie8-theme-srv.scss
>> >> @import "magnific-popup/src/css/main";
>> 
>>    ^
Warning:  Use --force to continue.
Aborted due to warnings.
```

Gonna try restarting this PR's build since I didn't run into any issues locally while prepping this PR.

package.json
```diff
@@ -36,11 +36,11 @@
   },
   "devDependencies": {
     "assemble-contrib-i18n": "0.1.*",
+	"autoprefixer": "^8.2.0",
```
nschonni: Indent

EricDunsworth: Ugh, sorry. I'll fix it in a sec.

Recently got a new workstation and completely forgot to reinstall my editor's EditorConfig plugin.

EricDunsworth:
The Travis-CI error is still appearing after I initiated a manual rebuild and updated this PR. Really not sure what's going on with it. Haven't been able to reproduce it locally on Windows.

EricDunsworth:
I've force-pushed a rebased version of this PR with package-lock.json removed. Let's see how Travis-CI reacts...

nschonni:
Might need to redo the package.json and lock file now that the Sass part is fixed

EricDunsworth:
@nschonni Done :). I'll remove "[Do not merge]" from this issue's subject if Travis-CI builds successfully.

EricDunsworth:
Looks like it's working now :)!

The only minor remaining thing I noticed was this:

```
(node:4277) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
>> 0 passed! (0.00s)

Done, without errors.
```

FWIW this PR didn't introduce it. It's also present in the Travis-CI logs of other recently-merged GCWeb PRs.

duboisp:
@EricDunsworth please fix the conflict and then we will merge. This PR are pre-approved during the technical review meeting.

EricDunsworth:
Just updated this PR's package-lock.json file... but wanted to get your takes on some parts that changed in it before considering it to be merge-worthy.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "GCWeb",
-  "version": "4.0.28-development",
```
EricDunsworth: Would it be more appropriate to update this line in a separate PR? It probably should've happened as a part of #1352, but was probably missed.

package-lock.json
```diff
@@ -9021,7 +9098,7 @@
       "dev": true
     },
     "wet-boew": {
-      "version": "github:wet-boew/wet-boew#0295e93e2b8d8e0ed41a62923046d87d4ef56081",
```
EricDunsworth: Does this make sense? Running ``npm install`` seems to have bumped up the wet-boew dependency's commit hash from 4.0.28's final commit to 4.0.29-development's most recent commit.

package-lock.json
duboisp: Yes it was missed, I don't think there is a value to have a separate PR.

package-lock.json
duboisp: Yes, GCWeb should always use the latest version of wet-boew, but do you think we would need to update the GCWeb package.lock file every time we merge a PR in wet-boew?

package-lock.json
```diff
@@ -9034,11 +9111,6 @@
         "mathjax": "2.7.1",
         "proj4": "2.3.3",
         "unorm": "1.4.1"
-      },
```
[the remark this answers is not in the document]
EricDunsworth: Actually, nevermind this. I didn't review the diff properly. This dependency got moved to a different spot in package-lock.json.

package-lock.json
EricDunsworth: That might be a better question for @nschonni and @LaurentGoderre.

I think what happened here is that since package.json doesn't currently specify a specific version or commit hash for wet-boew, ``npm install`` always ends up installing the latest commit.

From my understanding, package.json is valued more than package-lock.json. So for example, if someone runs ``npm install`` in a project whose package.json file specifies a newer version of a dependency than package-lock.json, package.json will "win" and package-lock.json will get updated to what got installed.

EricDunsworth: Maybe package.json could be modified to specify the tag of the most recent major release of wet-boew and only be modified on an as-needed basis in-between releases. Like changing it to a specific commit hash from time to time if significant things change in wet-boew that could impact GCWeb.

package-lock.json
EricDunsworth: In the end I decided to bring this in via a separate PR (#1353). I'll remove that change from this PR.

package-lock.json
EricDunsworth: Going to revert this line to what it originally was since it's beyond the scope of this PR.

PR #1347, opened 2018-04-19
Fix Sass build issue

nschonni:
Add the package-lock.json to avoid transient build failures.
Added the nested node_modules of wet-boew, because some of the modules seem to get nested as part of the install.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -403,6 +403,7 @@ module.exports = (grunt) ->
 			options:
 				includePaths: [
 					"./node_modules"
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
The nesting happens when there is a version conflict for a package

nschonni:
Yup, just not sure why it seems to just be happening on Travis, and the package `magnific` really shouldn't be conflicting unless there is a weird cache thing going on their. I couldn't repo it one my work machine.

Either way, the fix doesn't really hurt anyway, even if the de-dupe isn't hit on some machines.

PR #1349, opened 2018-04-23
Build - Added Sub-resource basic integrity generation

duboisp:
The sub-resource integrity hash will be included with the release note.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -24,6 +24,7 @@ module.exports = (grunt) ->
 			"htmllint"
 			"bootlint"
 			"useMinAssets"
+			"sri"
 		]
 	)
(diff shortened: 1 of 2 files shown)


nschonni:
It looks like the lock file changes got missed here

duboisp:
@nschonni Should I get it fixed before the GCWeb release or can I get it fixed after?

nschonni:
I'm just looking now. It shouldn't affect the actual SRI, just possibly the build consistency if someone tries to go back and build. We only just added the lock to this repo, so it's not like they had that option before anyway :P

duboisp:
ok great, I will finish the release then I will submit a PR with the updated lock file

PR #1351, opened 2018-04-23
chore: Add missing lock file updates

nschonni:
Adds some missing grunt packages to the lock file

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3478,12 +3478,53 @@
       }
     },
     "grunt-contrib-clean": {
-      "version": "0.6.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -5523,9 +5592,6 @@
       "integrity": "sha1-bUUk6LlV+V1PW1iFHOId1y+06VI=",
       "dev": true
     },
-    "magnific-popup": {
```
nschonni: Not sure how this one got here in the first place, but this would have been why the build was failing before without changing the Sass import paths

PR #1359, opened 2018-05-25
Fieldflow - Add ifNone default action

duboisp:


src/plugins/fieldflow/fieldflow.js
```diff
@@ -1261,6 +1280,9 @@ $document.on( fieldflowActionsEvents, selector, function( event, data ) {
 		case "removeClass":
 			$( data.source ).removeClass( data.class );
 			break;
+		case "query":
```
duboisp: Split this block in a different commit

PR #1361, opened 2018-05-25
menu - Added colour to menubar css in Basic HTML mode

### LINKED ISSUE #1021: Canada.ca theme: Resize text/contrast minimum
When in Firefox, the text only is zoomed at 200% (or 170%), the menu text is white text on white background., esp for smaller resolution screens (1024+ width).  It fails for colour contrast and text resize. Here is a screenshot to explain what I mean.

snap1

zzucco:
Fixes #1021

Issue caused by menubar overflowing when page was too small or too zoomed in.
Only occurs when WET is not loaded.
Added background colour to menubar to preserve readability.

### DIFF
--- a/src/menu/_base.scss
+++ b/src/menu/_base.scss
@@ -5,6 +5,7 @@
 	background: $accent-blue;
 
 	.menu {
+        background: $accent-blue;
 		display: table;
 		margin-bottom: 0;
 		text-shadow: 1px 1px 1px #222;
--- a/src/menu/_base.scss
+++ b/src/menu/_base.scss
@@ -5,7 +5,6 @@
 	background: $accent-blue;
 
 	.menu {
-		background: $accent-blue;
 		display: table;
 		margin-bottom: 0;
 		text-shadow: 1px 1px 1px #222;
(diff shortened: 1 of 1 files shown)


duboisp:
@zzucco can you submit the same change to the wet-boew. Then it will be also included in GCweb

zzucco:
@duboisp I'm not sure where to submit the change in the wet-boew. Should I change the [base plugin css](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/menu/_base.scss) or the [wet-boew theme css](https://github.com/wet-boew/wet-boew/blob/master/theme/menu/_base.scss)? I don't know how it affects the GCWeb theme.

duboisp:
@zzucco it should be in the base plugin css because the goal is to have the same default behaviour for every theme.

zzucco:
I can add colour to the base plugin css, but we will need to override it in GCWeb because the WET theme and GCWeb theme use different colours for their menu (#0e4164 vs #335075).

I haven't put in a PR yet, but [here](https://github.com/zzucco/wet-boew/commit/219e25eb970e02e0f696a0b44d1f88595bc30d77) is the change on my fork.

duboisp:
@zzucco please send the PR in wet-boew. The colour variable is overwritten in this theme.

duboisp:
Closing as the fix in wet-boew fixed the issue in GCWeb

PR #1363, opened 2018-05-28
Changes to colouring of placeholder text

### LINKED ISSUE #1285: INPUT tag placeholder styling isn't working as I'd expect
I am currently working with WET v4.0.23 and the GCWeb theme it uses is at https://www.canada.ca/etc/designs/canada/cdts/gcweb/v4_0_23/css/theme.min.css. 

The styling I am having a problem with is on line 12 in v4.0.23 regarding:
```
    [placeholder], input[placeholder] {
        color: #5c5c5c!important;
    }
```

It seems the definition for the placeholder font color is overriding the color of the font even when the placeholder text isn't being displayed. 

I created a simple HTML that mimics what I am doing in my application to test how an INPUT tag with placeholder should work and how it works when the styling is defined as in GCWeb. When I use the page below outside WET (no styling applied whatsoever), the font changes to the default (full black in the absence of any other styling) from the placeholder's grey font. This is as I expect. 

However, using the GCWeb theme, the placeholder's grey font styling remains even when a value is set in the INPUT tag. I think something needs to be adjusted in the GCWeb styling (assuming this hasn't already been fixed). No idea what it should be adjusted to, though.

```
<html>
<body>
  <div id="myDiv">
    <script type="text/javascript">
    function fn(){
        var someValue = "Some value";
        var inputElement = document.getElementById('inputElem');

        if (inputElement.value == someValue){
            document.getElementById('buttonElem').textContent = "Update Value";
            inputElement.value = "";
            inputElement.labels[0].textContent = "Note the font for the 'placeholder' text.";
        }
        else {
            document.getElementById('buttonElem').textContent = "Remove Value";
            inputElement.value = someValue;
            inputElement.labels[0].textContent = "Note whether the font changes when the 'placeholder' text is no longer displayed.";
        }
    }
    </script>
    <label for="inputElem">Note the font for the 'placeholder' text.</label>
    <p />
    <input id="inputElem" type="text" placeholder="Click button ..." readonly="readonly">
    <p />
    <button id="buttonElem" onclick="fn()">Update Value</button>
  </div>
</body>
</html>
```

Hoping this is clear enough, 
jtm

zzucco:
Normally, a text field's placeholder text is a different, usually paler color than the user input. Currently, we are setting both of these to the same color: #5c5c5c

The first part of this PR makes it so only the placeholder text is being set to #5c5c5c. User input will be colored as #555555 by the theming for form-control elements.

The second part of this PR changes the placeholder colour to #636463. This is the palest value that satisfies the WCAG requirement 1.4.3 on contrast. (contrasting against the color of the search bar, #e0e0e0)

It should be noted that the difference between placeholder text and normal text is still quite minimal, but this is what is needed to satisfy WCAG.

Possible issues:
The changes to default.scss may cause unwanted changes to styling, although I cannot think of any situation where it would. If this is a concern, there is another fix.

Fixes #1285.

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -28,9 +28,7 @@ a {
 }
 
 // scss-lint:disable QualifyingElement
-input[placeholder],
-[placeholder],
-*[placeholder] {
+*[placeholder]::placeholder {
 	color: $input-color-placeholder;
 }
 
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -32,6 +32,6 @@ Left menu variables
 $menu-background: #fff;
 $menu-colour: #555;
 
-$input-color-placeholder: #646564 !important;
+$input-color-placeholder: #636463 !important;
 
 @import "wet-boew/src/variables";
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -32,6 +32,6 @@ Left menu variables
 $menu-background: #fff;
 $menu-colour: #555;
 
-$input-color-placeholder: #767676 !important;
+$input-color-placeholder: #646564 !important;
 
 @import "wet-boew/src/variables";
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -32,6 +32,6 @@ Left menu variables
 $menu-background: #fff;
 $menu-colour: #555;
 
-$input-color-placeholder: #5c5c5c !important;
+$input-color-placeholder: #767676 !important;
 
 @import "wet-boew/src/variables";


duboisp:
* [ ] Check what is the browser default color, [Bootstrap override it to # 999](https://github.com/twbs/bootstrap-sass/search?q=placeholder&unscoped_q=placeholder)
* [ ] We might be able to removed ```*[placeholder]::placeholder {``` Selector and just rely on how Bootstrap override it.

zzucco:
Different browsers have different default colours for placeholders. Bootstrap overrides that, and we override Bootstrap, since 999 is not WCAG compliant.

We can probably remove the `*[placeholder]::placeholder` bit of code, since Bootstrap is styling all form-control elements anyways. I left it in as a sort of a catch-all in case there's something Bootstrap misses.

If you think it wise, I'll update the PR.

duboisp:
It look like bootstrap did that change via a "rake convert" (Jul 21, 2014) commit and it was merged by the PR 1146 which seems to introduce mixin.

But you said that "different browser have different default colours". Can you list what default colour they use?

zzucco:
I'm currently doing some more research before making any changes to this.

I think I will open a PR in WET, since placeholder contrast is an issue everywhere.

zzucco:
Regardless of the solution in the above discussion, this PR will not be merged. If we decide to go ahead with changing the colour of placeholders, then an entirely different change will be made in wet-boew repo. As such, I recommend closing this.

PR #1297, opened 2017-10-04
Various updates to bring certain teamplates to C&IA spec 1.4

lucas-hay:
Various changes to content and layout of templates. Very few style changes.

PR #1375, opened 2018-07-09
Theme: Removed leftover references to to the "pagetag" class/property.

EricDunsworth:
References to the "pagetag" class, property and include were removed from all pages in #1155.

However, #1155 didn't remove the property from the French topic page template. Furthermore, #1297 inadvertently re-introduced the class into the departments and agencies page. The class was also present in #1372's long index pages.


duboisp:
Pre-approved at the technical review meeting once the conflict is fixed. The conflict it just because a files was moved.

PR #1395, opened 2018-08-22
WIP - action manager - Add table to CSV action

duboisp:


duboisp:
Now the "download" function is integrated in the wet-boew core, we only need to make the necessary change to the existing code.

@GormFrank if you have time, can you do the work?

PR #1399, opened 2018-09-11
Build: Added lintspaces task to check for .editorconfig compliance.

EricDunsworth:
* Added "lintspaces" at the end of the "test" task (which runs as part of the dist build).
* Inherits rules from the repo's .editorconfig file.
* Checks all of the repo's tracked files (minus comments, non-SVG images and external fonts).

**PS:**
This PR only introduces the linter itself. #1397 and #1398 should be merged-in beforehand, followed by a manual Travis-CI rebuild. The linter should start passing afterwards.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -137,6 +137,7 @@ module.exports = (grunt) ->
 		[
 			"eslint"
 			"sasslint"
+			"lintspaces"
 		]
 	)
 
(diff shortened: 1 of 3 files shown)


EricDunsworth:
FYI this has been rebased and is now building successfully... but I don't like some of what I'm seeing in package-lock.json's diff. So I'd like to take a closer look at that and possibly force-push a replacement commit before merging.

I'll follow-up once I've had a chance to do so.

PR #1460, opened 2018-12-13
Cleaner version of GC signature - English

patlaj:


patlaj:
Thanks @duboisp - re these:

Ensuring the all SVG parameter are accurate (Example the viewBox parameter)
Ensuring the new SVG can be scaled and use only relative width and height.

Can you let me know what's wrong with this file? I've run it through a validator (after making the other changes you specified) and get no issues. Note that the current sig does not display properly on the size we use on pages - the characters space together strangely and some blur. So if that is rendered correctly it makes me wonder if there's something wrong with how it's specified even if it's syntactically correct.

duboisp:
Can you provide some screen shot showing/explaining the issue?

I would like to check if it is really the SVG file that is faulty. May be the solution is simple as adjusting the HTML container (Grid) for the location where that SVG image is used.

When I look at both SVG image, after it is rendered by the browser, both look visually identical, except yours seems to have a fixed the positioning with a fixed top margin/padding.

The SVG is a XML file. It is expected to have it properly formatted and to contain meaningful content as defined by the [SVG specification](https://www.w3.org/TR/SVG11/). In addition, it is expected the file is optimized from a code point of view, like not having multiple consecutive ```<path>``` element when it's not needed and to avoid inline styling. Removing the redundancy will optimize the performance of the SVG when it is rendered.

patlaj:
That would be great...it was a pain to get this file together! Here you can see the issue:

screen shot 2018-12-13 at 2 10 54 pm

I zoomed in a bit to make it clearer:

screen shot 2018-12-13 at 2 06 25 pm

Problems are:

1. right side of ‘t’ in Gouvernment is cut off
2. 'e' shaped weirdly (too thin cross bar, top sticks up a bit)
3. lots of kerning issues, especially m-e and n-e: too much space, e-m: not enough space

When you view the SVG file on its own it doesn't have these issues, so we thought it was a result of how the shapes were defined.

duboisp:
I see.

Are you able to only modify the value of the [path data](https://www.w3.org/TR/SVG11/paths.html#PathData) (d attribute of the path element) : https://github.com/wet-boew/GCWeb/pull/1460/files#diff-7e84c6d2240afd3755dab832b266d3bfL12

Or can you modify the content of your SVG file to be structured similarly as it is currently? Like build with 2 ```path``` element, using CSS class and using relative width/height.

patlaj:
FYI just committed an updated version to use an XML header, proper svg attribute, and classes instead of inline styles. I didn't do anything with the paths or relative height/width because I don't know how to do that so it still works. Maybe with some experimentation...

duboisp:
@patlaj FYI - I am currently looking at the SVG file. I am trying to cleaning up a little like removing the unnecessary attribute ```transform="translate(0 -0.02)"``` added to every element and to merge a few path element together. I am also going to set the "preserveAspectRatio".

duboisp:
Related: https://github.com/gc-proto/GCWeb/pull/1

PR #1464, opened 2018-12-19
remove role attribute from updated header markup

shawnthompson:


shawnthompson:
Man this wasn't easy to do in the browser

PR #1467, opened 2018-12-21
Clean search templates for the design refresh

Borges-Porteous:


duboisp:
@Borges-Porteous The build script don't pass because you are using ```{{}}``` character in your code. As it is for a placeholder, you can use ```[[]]``` instead.

duboisp:
Single square brackets is fine too

Borges-Porteous:
Ah, thanks Pierre: just noticed your comment. I'll try with the single square bracket then. Hopefully will work this time!

site/pages/search-results-fr.hbs
```diff
@@ -0,0 +1,240 @@
...
+				<li class="[link]">[domain name of search result]</li>
+				<li class="[link]">[parent breadcrumb of page]</li>
+				<!--Note: this always has 2 list items. Does not include other breadcrumb items.-->
+			<ol>
+			<p><time datetime="[dateModified of search result]">[DD MMM. YYYY]</time> - [description/highlight of search result. Search terms in <strong>]</p>
+		</section>
+		<section>
+			<h3><a href="[link]">[Search result page title (not including '- Canada.ca' or other institutional suffix)]</a></h3>
+			<ul>
+				<li class="label">[institution name]</li>
```
duboisp: It will be best to set the CSS class on the ```<ul>``` instead of each individual ```<li>```. Because that unordered list would only contains labels.

site/pages/search-results-fr.hbs
```diff
@@ -0,0 +1,240 @@
...
+				<li class="label">[institution name]</li>
+				<li class="label">[institution name]</li>
+				<!--Note: if labels go past 1 row, then option a) cut off at end of row, option b) do not show any labels beyond the first row. Text-overflow of the cut off label should be with an ellipsis. Expand/collapse to show all is not a requirement for Jan 2018 MVP-->
+				<!--Note 2: First label should always be the label of the context the search is in (if contextualised). Not necessarily order of tags in page metadata.-->
+			<ul>
+			<ol class="breadcrumb">
+				<li class="[link]">[domain name of search result]</li>
```
duboisp: What is the use of the CSS class ```class="[link]"```. Is it needed?

>  This should be shown as the domain name of the page with a chevron “>” followed by the parent breadcrumb of the page. 

Is the intention to have each individual item to be clickable and pointing to different link, like a breadcrumb? If so, the markup would look like ```<li><a href="[link]">[breadcrumb name]</a></li>```

> When this is not possible, the breadcrumb can be replaced by the page URL.

Do you have a example of that? It might require a different markup, like a paragraph or a cite element.

site/pages/search-results-fr.hbs
```diff
@@ -0,0 +1,240 @@
...
+				<li class="[link]">[parent breadcrumb of page]</li>
+				<!--Note: this always has 2 list items. Does not include other breadcrumb items.-->
+			<ol>
+			<p><time datetime="[dateModified of search result]">[DD MMM. YYYY]</time> - [description/highlight of search result. Search terms in <strong>]</p>
```
duboisp: This is an invalid ```datetime``` attribute. The information you have included can be moved in a page dedicated for technical spec of the search page.

site/pages/search-results-fr.hbs
Borges-Porteous: Individual item shouldn't be clickable. I'll remove class there and add the example URL markup

site/pages/search-results-fr.hbs
Borges-Porteous: Should I just put a random date then?

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,111 @@
...
+	],
+	"dateModified": "2018-12-21",
+	"share": "false",
+	"deptfeature": false
+}
+---
+<!--Note: visual example with styles applied: https://gc-proto.github.io/student-loans/scenarioA/services/search-results-contextual-visual.html-->
+<!--Note 2: this template should only be used for contextual search results from a single institution-->
+<!--Note 3: until radio buttons pattern works, please simply include 'All government of Canada websites' link in the well, under the search box-->
+<div class="row profile">
```
duboisp: What is the use of the CSS ```profile```?

That ```div.row``` followed by ```div.col-md-12``` are probably useless in that design

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,111 @@
...
+				<div class="input-group mrgn-tp-lg">
+					<input value="[value]" spellcheck="false" autocomplete="off" dir="ltr" id="sch-inp" name="q" class="form-control" type="search">
+					<span class="input-group-btn">
+						<button type="submit" id="wb-srch-sub" class="btn btn-primary btn-small" name="wb-srch-sub"> <span class="glyphicon-search glyphicon"></span><span class="wb-inv">Search</span></button>
+					</span>
+				</div>
+			</form>
+			<div class="row mrgn-tp-md">
+				<div class="col-md-4 nowrap">
+					<input type="radio" name="searchscope" id="radio-allresults" value="allresults"  /> <label id="label-alldepartments" for="radio-alldepartments">All Government of Canada websites</label>
```
duboisp: Did you try with an inline unordered list?

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,111 @@
...
+				</div>
+				<div class="col-md-8 nowrap">
+					<input type="radio" name="searchscope" id="radio-context" value="context" checked="checked" /> <label id="label-singledepartment" for="radio-singledepartment"><strong>[context label]</strong></label>
+				</div>
+			</div>
+		</div>
+	</div>
+</div>
+<div class="row">
+	<div class="col-xs-12">
```
duboisp: the ```div.row``` followed by ```div.col-xs-12``` are might be useless

site/pages/search-results-contextual-en.hbs
Borges-Porteous: No idea will remove row 19 and the closing div

site/pages/search-results-contextual-en.hbs
Borges-Porteous: will remove

site/pages/search-results-contextual-en.hbs
Borges-Porteous: Not sure if I got the syntax right - commiting it

site/pages/search-results-fr.hbs
duboisp: May be January 1 2018 ```datetime="2018-01-01"``` and in your technical spec you explain what that value correspond to

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,103 @@
...
+			<input type="hidden" name="cdn" value="canada" />
+			<input type="hidden" name="hq" value="" />
+			<div class="input-group mrgn-tp-lg">
+				<input value="[value]" spellcheck="false" autocomplete="off" dir="ltr" id="sch-inp" name="q" class="form-control" type="search">
+				<span class="input-group-btn">
+					<button type="submit" class="btn btn-primary btn-small" name="wb-srch-sub"> <span class="glyphicon-search glyphicon"></span><span class="wb-inv">Search</span></button>
+				</span>
+			</div>
+		</form>
+		<ul class="inline">
```
duboisp: That markup seems to be fine. I would may add a paragraph before explaining what it is about. At least for screen reader. ```<p class="wb-inv">....```

site/pages/search-results-fr.hbs
Borges-Porteous: I'll add that to the tech spec then

site/pages/search-results-contextual-en.hbs
duboisp: The markup change you did seems ok.

site/pages/search-results-contextual-en.hbs
duboisp: ok

duboisp: don't forget to remove the div with the column that is after ```<div class="col-md-12">```

Borges-Porteous: Can't see anything with a column inside the <div class="col-md-12">

Borges-Porteous:
The only thing it doesn't seem to like at this point is use of the `<url>` tag in the search results. Is that something I should change?

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,104 @@
...
+			<!--Note: if labels go past 1 row, then option a) cut off at end of row, option b) do not show any labels beyond the first row. Text-overflow of the cut off label should be with an ellipsis. Expand/collapse to show all is not a requirement for Jan 2018 MVP-->
+			<!--Note 2: First label should always be the label of the context the search is in (if contextualised). Not necessarily order of tags in page metadata.-->
+		</ul>
+		<ol class="breadcrumb">
+			<li>[domain name of search result]</li>
+			<li>[parent breadcrumb of page]</li>
+			<!--Note: this always has 2 list items. Does not include other breadcrumb items.-->
+		</ol>
+		<!--when the breadcrumb is not available to the search engine, use the URL instead-->
+		<url>[url of the search result]</url>
```
duboisp: ```<url>``` is not a valid element. You can use ```<p class="url">``` or ```<p><cite><a href="[link]">[link]</a></cite></p>```
Using the ```<cite>``` element is might be the best option as per it's definition in the HTML 5 spec: https://www.w3.org/TR/html53/textlevel-semantics.html#the-cite-element

Borges-Porteous: K, I'll add that, and should pass Travis, finally! Thanks Pierre!

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,104 @@
...
+	<div class="well">
+		<form name="cse-search-box" method="GET" action="/en/sr/srb.html" id="cse-search-box">
+			<label class="wb-inv" for="sch-inp">Search Government of Canada websites</label>
+			<input type="hidden" name="st" value="s" />
+			<input type="hidden" name="s5bm3ts21rch" value="x" />
+			<input type="hidden" name="num" value="10" />
+			<input type="hidden" name="st1rt" value="0" />
+			<input type="hidden" name="langs" value="eng" />
+			<input type="hidden" name="cdn" value="canada" />
+			<input type="hidden" name="hq" value="" />
```
duboisp: Do I am right to assume those ```input[type=hidden]``` are very specific to our search tool implementation. If so, I suggest to remove it from this template.

site/pages/search-results-contextual-en.hbs
```diff
@@ -0,0 +1,104 @@
...
+	],
+	"dateModified": "2018-12-21",
+	"share": "false",
+	"deptfeature": false
+}
+---
+<!--Note: visual example with styles applied: https://gc-proto.github.io/student-loans/scenarioA/services/search-results-contextual-visual.html-->
+<!--Note 2: this template should only be used for contextual search results from a single institution-->
+<!--Note 3: until radio buttons pattern works, please simply include 'All government of Canada websites' link in the well, under the search box-->
+<div class="col-md-12">
```
duboisp: Remove this column ```div.col-md-12```, it's a bootlint issue and it's useless in this design.

site/pages/search-results-contextual-en.hbs
Borges-Porteous: No idea. More of a question for Cristian.

I assume you're right though! I'll remove them from the template

site/pages/search-results-contextual-en.hbs
Borges-Porteous: Okay

duboisp:
@Borges-Porteous Can you add a link this new search template from those page:

* https://github.com/wet-boew/GCWeb/blob/master/site/pages/index-en.hbs
* https://github.com/wet-boew/GCWeb/blob/master/site/pages/index-fr.hbs

Thanks

Borges-Porteous:
> @Borges-Porteous Can you add a link this new search template from those page:
> 
> * https://github.com/wet-boew/GCWeb/blob/master/site/pages/index-en.hbs
> * https://github.com/wet-boew/GCWeb/blob/master/site/pages/index-fr.hbs
> 
> Thanks

Where do I add the link? In the template?

duboisp:
> Where do I add the link? In the template?

Add a new table row for each template if they are not already in that list. That list is the same and ordered the same as the list of template in the Content and IA spec.

duboisp:
The first column will contains a link to the template (the pages that you created with this PR). the text link is the name of the template.
The second column will be today date, at least to match up the last modified date set on the new template included in this PR.
The third column will be text "Template/Search" and optionally linked to a new page for the technical spec in a new folder named "search". (See the example for the home page and for the index pages)
The fourth is a short description, one sentence.
The fifth column is the status. All status meaning is defined in an expand/collapse after the third paragraph on that page. 
The sixth column is the name of the specification with a corresponding direct link if exists.
The seventh column is the version of the specification the current template match.

Borges-Porteous:
> The first column will contains a link to the template (the pages that you created with this PR). the text link is the name of the template.
> The second column will be today date, at least to match up the last modified date set on the new template included in this PR.
> The third column will be text "Template/Search" and optionally linked to a new page for the technical spec in a new folder named "search". (See the example for the home page and for the index pages)
> The fourth is a short description, one sentence.
> The fifth column is the status. All status meaning is defined in an expand/collapse after the third paragraph on that page.
> The sixth column is the name of the specification with a corresponding direct link if exists.
> The seventh column is the version of the specification the current template match.

Thank you - think I got it right this time!

duboisp:
@Borges-Porteous did you already have the CSS (SASS) to style the new search result template? The style/code will must likely goes in the src/search/_base.scss file.

Borges-Porteous:
> @Borges-Porteous did you already have the CSS (SASS) to style the new search result template? The style/code will must likely goes in the src/search/_base.scss file.

Hmm, never worked with SASS before. BAsed on the _base.scss file in the 'search' folder under 'src', it looks like any other css? What's the difference?

Also, [the prototype to show how it should look](https://github.com/gc-proto/student-loans/blob/master/scenarioA/services/search-results-visual.html) was just done with in-line styles.

I could try making a css file based on the prototype, but I'll need significant amounts of help... Maybe you could send an example of how the structure would look like? Or what the new topic page uses?

How could I preview the effects? Not sure how to preview an hbs file.

And, there were a few pieces that I had never quite figured out, that I was hoping someone with more css experience could do, specifically the overflow cut off when context labels would go past 1 row.

Borges-Porteous:
@duboisp Took a stab at it and have tried to add the new base.scss file.

It's going to need your review for sure. Please advise!

duboisp:
duplicate of #1480

PR #1472, opened 2019-01-07
Fix doormat font size, body line-height and typo

jmealing:
- Fixed open gov links
- Doormats: paragraph fonts resized and line-height corrected
- Doormats: add mobile size specs for paragraph
- Change line-height of main copy

### DIFF
--- a/src/views/_screen-sm-max.scss
+++ b/src/views/_screen-sm-max.scss
@@ -15,6 +15,13 @@ h2,
 	font-size: 1.6em; // expected 342x;
 }
 
+// Increase doormat text size for mobile
+.gc-srvinfo, .gc-drmt {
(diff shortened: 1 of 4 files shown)


src/views/_screen-sm-max.scss
```diff
@@ -15,6 +15,13 @@ h2,
 	font-size: 1.6em; // expected 342x;
 }
 
+// Increase doormat text size for mobile
+.gc-srvinfo, .gc-drmt {
+	p {
+		font-size: 19px;
+		line-height: 1.35em;
+	}
+
```
duboisp: Missing a closing brace. 
Not sure if targeting the ```p``` element is required.

PR #1477, opened 2019-01-15
V2 - Form validation error label in legend - Item 89

GormFrank:


src/_defaults.scss
```diff
@@ -159,6 +159,9 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 form .btn:not(.btn-call-to-action), .form-control {
 	padding: 6px 12px;
 }
+legend {
+	line-height: 1.6275em;
```
duboisp: Can you change it for ```1.65em```? so it will be same value as line height for the main content.

GormFrank: Sure, no problem!

PR #1478, opened 2019-01-16
Some label and description edits

patlaj:


patlaj:
Ah thanks Pierre, committed now. The menu and home page don't necessarily have to match, it is transitional now from the old menu (with the short name only).

site/pages/home-en.hbs
```diff
@@ -60,12 +60,12 @@
 			<p>Find a job, training, hiring programs, work permits, Social Insurance Number (SIN)</p>
 		</div>
 		<div class="col-lg-4 col-md-6">
-			<h3><a href="https://www.canada.ca/en/services/immigration-citizenship.html">Immigration</a></h3>
+			<h3><a href="https://www.canada.ca/en/services/immigration-citizenship.html">Immigration and citizenship</a></h3>
```
duboisp: FYI - I would apply the same update at the French equivalent.

PR #1480, opened 2019-01-16
Searchbranch

Borges-Porteous:
The search template updates from a new branch of the gc-proto GCWeb repo.


src/search/_base.scss
```diff
@@ -49,4 +49,72 @@
...
 		}
 	}
+
+	//new styles from the update
+	h2 {
+		font-family: "Noto sans", sans-serif;
+		font-size: 20px;
+		line-height: 34px;
+		padding-bottom: 8px;
+		padding-top: 8px;
```
duboisp: Is there a rational to style the heading differently from the default? like using "Noto sans" instead of "Lato",....

Heading need to be differentiated from normal text

site/pages/search-results-contextual-fr.hbs
```diff
@@ -0,0 +1,96 @@
...
+			</span>
+		</div>
+	</form>
+	<p class="wb-inv">Voire les résultats de:</p>
+	<ul class="inline">
+		<li><input type="radio" name="searchscope" id="radio-allresults" value="allresults"  /> <label id="label-alldepartments">Tous les sites Web du gouvernement du Canada</label></li>
+		<li><input type="radio" name="searchscope" id="radio-context" value="context" checked="checked" /> <label id="label-singledepartment"><strong>[context label]</strong></label></li>
+	</ul>
+</div>
+<div class="col-xs-12">
```
duboisp: This ```div.col-xs-12``` can be removed. Also a column always need to be contained in a ```div.row``` as per bootlint.

src/search/_base.scss
```diff
@@ -49,4 +49,72 @@
...
+	cite a {
+		color: #006621;
+		display: inline-block;
+		font-size: 20px;
+		margin-top: 10px;
+	}
+
+	p {
+		font-family: "Noto sans", sans-serif;
+		font-size: 20px;
```
duboisp: It is unnecessary to overwrite the default font-size + family to the same value.

src/search/_base.scss
```diff
@@ -49,4 +49,72 @@
...
+		line-height: 34px;
+		padding-bottom: 8px;
+		padding-top: 8px;
+	}
+
+	time {
+		color: grey;
+	}
+
+	section {
```
duboisp: Styling the section element like that might be not optimal and could cause some undesirable effect. It would worth it to target that style with a CSS class.

src/search/_base.scss
```diff
@@ -49,4 +49,72 @@
...
+
+	p {
+		font-family: "Noto sans", sans-serif;
+		font-size: 20px;
+		line-height: 34px;
+		padding-bottom: 8px;
+		padding-top: 8px;
+	}
+
+	time {
```
duboisp: Styling the section element like that might be not optimal and could cause some undesirable effect. It would worth it to target that style with a CSS class or to ensure it is in-scope.

Borges-Porteous:
> Is this a duplicate of PR #1467?
> 
> We should use relative font-size where possible. The base font size is 20px.

Yes - there was some confusion on our end and the search pages were deleted from the master branch of the GC-Proto fork, which the pull request was from. So this pull request is just getting them back into the repo in a search updates branch to get them merged.

I'll update to em where possible

src/search/_base.scss
Borges-Porteous: I've changed the 'h2' to simply a paragraph tag, since on further research, the number of results isn't really a heading, and isn't used as such by other search engines like Google or Bing.

I've added a 'results-count' class specifically for this one, and faded it a bit. We should also probably remove all the padding around it, but can't seem to...

I've changed the search result link to an h2 class="h3", and we can just use the standard Canada.ca h2 styling, which is lato 24px, if I saw correctly.

Hope that works!

site/pages/search-results-contextual-fr.hbs
Borges-Porteous: It is removed.

src/search/_base.scss
Borges-Porteous: I've removed those two and just added a note so it isn't missed.

src/search/_base.scss
Borges-Porteous: made it a class and put it on the time tags

src/search/_base.scss
Borges-Porteous: added a search-description class

src/search/_base.scss
```diff
@@ -49,4 +49,63 @@
 			margin-right: auto;
 		}
 	}
+
```
duboisp: The following CSS need to be scoped by the page type CSS ```.page-type-search```.
FYI - I am going to do that update and I am working on it to help you to finalize this search template

duboisp:
@Borges-Porteous you can see my progress here: https://github.com/duboisp/GCWeb/tree/search-template

FYI - I am splitting the search template into individual component and documenting them individually. After I am going to review your documentation (google docs) compared to your code sample to create a static example page rendered in that template. Then I would adjust/apply the styling needed for that design.

src/search/_base.scss
Borges-Porteous: Thank you Pierre!

Borges-Porteous:
> I am splitting the search template into individual component and documenting them individually. After I am going to review your documentation (google docs) compared to your code sample to create a static example page rendered in that template. Then I would adjust/apply the styling needed for that design.

Glad to hear you were able to make some progress while I was away! I look forward to seeing what it looks like when it's all brought together!

Once theirs a working version, I'll finalize the documentation and work with David to add it to the Canada.ca design system.

Please tell me what's missing from my end on the coding, or if it would help if I go to the Code Sprint on Tuesday to discuss the next steps.

site/pages/search-results-contextual-fr.hbs
```diff
@@ -0,0 +1,94 @@
...
+			<span class="input-group-btn">
+				<button type="submit" class="btn btn-primary btn-small" name="wb-srch-sub"> <span class="glyphicon-search glyphicon"></span><span class="wb-inv">Recherche</span></button>
+			</span>
+		</div>
+	</form>
+	<p class="wb-inv">Voire les résultats de:</p>
+	<ul class="inline">
+		<li><input type="radio" name="searchscope" id="radio-allresults" value="allresults"  /> <label id="label-alldepartments">Tous les sites Web du gouvernement du Canada</label></li>
+		<li><input type="radio" name="searchscope" id="radio-context" value="context" checked="checked" /> <label id="label-singledepartment"><strong>[context label]</strong></label></li>
+	</ul>
```
duboisp: Having those input radio after the submit button could create an info and relationship/meaningful sequence issue. An alternative would be to have the submit button after those radio button. I know we have discussed to replace the input by a link which could technically "avoid" that issue but it may create some confusion as it will be still grouped together. I will draft some design alternative for further discussion, like using the tabbed interface or having the submit button after the input radio.

duboisp:
@Borges-Porteous Can the color of the time in the search result item be #6f6f6f which is the color of "text-muted" CSS instead of using the color "#656565" as specified in the draft spec?

PR #1502, opened 2019-02-07
Search template

duboisp:
The compiled/test version corresponding at this PR: https://universallabs.org/wet/labs/search-template/search/

@Borges-Porteous can you review?

Close #1480


Borges-Porteous:
Ah, this is where it is! Was confused trying to find out which things to review...

This looks great Pierre! You've done a lot to document all those pieces! Thank you. 

All three versions seem to be good. Is it alright now to get Cristian involved in setting a demo up with live search results then?

I know you had a couple comments on the contextual version radio buttons and colour of date. Thanks for looking into the radio buttons. Would be great if that could work. We didn't test a tabbed version, but that is a design that has been used successfully on other sites, from what I can tell, so it's a good idea for an alternative. It might not be worth putting too much effort into this if we go with the filter box approach though. The filter box is still more flexible than either radio or tabbed. So maybe we should focus on that and if it can't get implemented, we can revisit tabs.

No issue with the colour you suggested for date #6f6f6f. I have made the update in the draft specs in google docs.

Borges-Porteous:
One other thing - the Canada.ca search input field in the header (top right) should not appear on the search results pages. Is it possible to remove that?

duboisp:
What about the "advanced search" and the "search tips" link? Do they still need to be included the template?

duboisp:
@Borges-Porteous have you seen my previous comment regarding if we still have the "advanced search" and the "search tips" link? If so, where they should be included in the template?

Borges-Porteous:
Ah, I had not noticed. Sorry about the delay.

No search tips/advanced search links. We can remove those.

duboisp:
I updated the documentation to add in the backlog the advanced search and the starting search page. Also I squash all commit into one.

The test version was updated: https://universallabs.org/wet/labs/search-template/search/

@Borges-Porteous Before I go forward and merge this, can you quickly review the docs and the working example as illustrated in the test version? Thanks

Borges-Porteous:
> I updated the documentation to add in the backlog the advanced search and the starting search page. Also I squash all commit into one.
> 
> The test version was updated: https://universallabs.org/wet/labs/search-template/search/
> 
> @Borges-Porteous Before I go forward and merge this, can you quickly review the docs and the working example as illustrated in the test version? Thanks

Just reviewed the examples and they look great. 

The only thing to mention would be that in the contextual search, if both the 'All Government of Canada' and 'Contextual filter' are displayed, either as links or radio buttons, the 'active' one, should be bolded.

Please go ahead!

duboisp:
Pre approved during the tehcnical review meeting. I am going to do the small change requested and also and in the "search template roadmap" an item to add a instruction block for assistive technology to explain how the search result item are structured.

PR #1511, opened 2019-02-18
Menu fix when navigating with a screen reader

duboisp:
This fix an issue that prevent to operate the main menu in IE11 by using the NVDA screen reader.

Closes https://github.com/wet-boew/wet-boew/issues/8556

Remaining test:
* Chrome with NVDA
* Chrome with Jaws18
* Re-test with Firefox, IE and Edge

Test page: http://universallabs.org/wet/labs/gcwebmenu/content-en.html


duboisp:
I received the test result from WAWG, which I am going to publish it later.

PR #1519, opened 2019-03-18
Add fontawesome

patlaj:
Stylesheet link and related documentation to add Font Awesome to the GCWeb Wet variant.


EricDunsworth:
@patlaj @duboisp @GormFrank Just a few more things to think about:
* Font awesome's CDN is setup to be able to provide individual subsets of it. According to [its start page](https://fontawesome.com/start), "all", "solid", "regular" and "brands" variants of of it exist. [Only "solid" and "brands" are free to use](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use). To avoid getting a bunch of extra unusable icons, maybe there could be value in going specifically with the "solid" (and optionally "brands") variants to reduce file size.
* Font awesome's CDN URLs contain specific version numbers...
* Would there potentially be value in setting up font awesome as a GCWeb dependency and have it directly imported into GCWeb's main CSS in a similar manner to what's done for Bootstrap? FWIW, [font awesome has its own GitHub repo](https://github.com/FortAwesome/Font-Awesome).

patlaj:
Still failing due to the "integrity" attribute of course. 

I also just noticed after committing that it's already at 5.8.1. I don't know if we should try to deal with that (so we don't need a new version of GCWeb to update the Font Awesome version), or just accept that there will be a bunch of FontAwesome updates in between. I'd be inclined to just accept it.

@EricDunsworth it would be a good idea to not import icons we can't use, but there are several of the "regular" icons that actually are available for free. For example, there's the super-useful open circle:

https://fontawesome.com/icons/circle?style=regular

So I think we'd need to get "All" anyway.

duboisp:
It will better if we go directly with the latest one as today. Just update your PR.

I sent a PR that fix the "integrity" attribute error. Now the build pass

duboisp:
As discussed at the technical review meeting it was suggested to:
* Integrate font-awsome CSS inside the theme.css. That means the GCWeb build script would need to be modified to take a copy of the font-awsome SASS github and ensure it is compiled with the all the SASS file we have for the GCWeb theme.
* We should limit only to the icon that are free to use. 
* We will need to sort the icon that are free when updating the build script because some regular icon, as @EricDunsworth mentionned during the meeting, is part of the pro package according to https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use and there is no light icons free

We will need the list of free icons and SASS file that are free to include

patlaj:
There is an advantage to using the FontAwesome CDN, because anyone who already downloaded the fonts wouldn't have to again.

It'd be good to only load the sets that are free, but there are some very useful icons in the regular set that are free... https://fontawesome.com/icons/circle?style=regular not sure if it's possible to only get some icons out of a set?

cfarquharson:
I agree with @patlaj there are some useful free icons that are in the "far" set (some are free while others are subscription).  The "fas" is all free but some good options in "far" too.  If we are going to add icons, I recommend getting as many as we can.

EricDunsworth:
Sorry, when I made my previous comment, it was based on what I saw on [font awesome's basic use page](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use)... which seems to misrepresent the entire set of regular icons as being pro (paid)-only. Didn't realize that a subset of regular icons were free.

From what I can see on the [start a new project page](https://fontawesome.com/start), the "all" CSS file is 53.8 KB, whereas the 4 individual CSS files for solid+regular+brands make for a total of 54.3 KB.

So given that, it would technically be more efficient to use the "all" CSS file despite the fact that many of its regular icons and all of its "light" icons would be unusable.

GormFrank:
@EricDunsworth  you got it. The all.css file is totally free (under the MIT license) and contains everything we need to work with, from regular to solid... and the brands. Plus, its size is equivalent if not less than a combination of the other package. So in my opinion, the link tag to the all file is the way to go, also considering the fact that it is in addition to the WET and not necessarily part of its core (unlike glyphicon). By having that link tag, we also ensure that our icons remain up-to-date, and "protected" by the integrity attribute.

duboisp:
@patlaj The integrity hash are incorrect, can you submit a new PR with the correct integrity hash?

patlaj:
Hey @duboisp - I submitted something the other day with the correct hash but am not sure it's all you need: https://github.com/wet-boew/GCWeb/pull/1527

PR #1529, opened 2019-04-11
update-megamenu-20190411

thangle2013:


duboisp:
https://github.com/wet-boew/GCWeb/blob/master/site/pages/ajax/sitemenu-v2-en.hbs
https://github.com/wet-boew/GCWeb/blob/master/site/pages/ajax/sitemenu-v2-fr.hbs

PR #1539, opened 2019-05-14
Menu - Prevent menu closing for touch screen and basic instruction menu

duboisp:
Fix visual update temporary issue
* Item 73
* Item 74

Related

https://github.com/wet-boew/GCWeb/issues?q=is%3Aopen+is%3Aissue+label%3A%22Feature%3A+Menu+-+Plugin%22

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -20,7 +20,7 @@ var componentName = "gcweb-menu",
 	isMediumView,
 	preventFocusIn,
 	i18nInstruction = {
-		en: "Press the SPACEBAR to expand or the escape key to collapse this menu. Use the up and Down arrow keys to choose a submenu item. Press the Enter or Right arrow key to expand it, or the Left arrow or Escape key to collapse it. Use the up and Down arrow keys to choose an item on that level and the Enter key to access it.",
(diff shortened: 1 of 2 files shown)


src/plugins/menu/menu.js
```diff
@@ -18,6 +18,11 @@ var componentName = "gcweb-menu",
 	justOpened,
 	isMobileMode, // Mobile vs Desktop
 	isMediumView,
+	preventFocusIn,
+	i18nInstruction = {
+		en: "Press the SPACEBAR to expand or the escape key to collapse this menu. Use the up and Down arrow keys to choose a submenu item. Press the Enter or Right arrow key to expand it, or the Left arrow or Escape key to collapse it. Use the up and Down arrow keys to choose an item on that level and the Enter key to access it.",
```
GormFrank: For consistency purposes, I would suggest to put the letter "u" from "Up" in uppercase.

```suggestion
		en: "Press the SPACEBAR to expand or the escape key to collapse this menu. Use the Up and Down arrow keys to choose a submenu item. Press the Enter or Right arrow key to expand it, or the Left arrow or Escape key to collapse it. Use the Up and Down arrow keys to choose an item on that level and the Enter key to access it.",
```

PR #1555, opened 2019-06-20
Make primary buttons look the same everywhere on a page

jkshapiro:
Fixes a bug introduced by https://github.com/wet-boew/GCWeb/commit/6044d706d41a5fb9501e2ef9029f19945a493c34


duboisp:
To be tested with IE11.

PR #1568, opened 2019-07-25
Build: Update grunt to ~1.0.4

EricDunsworth:
Decreases build times due to an updated glob sub-dependency. Windows gets a substantial speed boost. Linux probably also benefits.

From a cold boot in Windows, grunt dist used to take ~4.75 minutes... now it only takes ~3.75 minutes.

This PR's generated dist folder is identical to before.

The newer version of CoffeeScript used in grunt 1.0.4 doesn't appear to introduce any breaking changes to Gruntfile.coffee.

Port of wet-boew/wet-boew#8690

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1105,9 +1105,9 @@
       "dev": true
     },
     "colors": {
-      "version": "0.6.2",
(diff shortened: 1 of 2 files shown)


GormFrank:
Pre-approved during technical review.

PR #1572, opened 2019-07-29
Form test cases: Remove stray </div> tag

EricDunsworth:


GormFrank:
Pre-approved during technical review as long as conflicts are resolved.

PR #1582, opened 2019-08-20
Fieldflow updates

jmealing:
To populate the new classes and layout for new checkboxes and radios


duboisp:
I think you text editor have replaced all tabs character by a the space character. That sometime happen when the text editor are not configured or are missing a plugin to follow the rule in the .editorconfig files. 

Also, I saw a few error related to the [jQuery style guide](https://contribute.jquery.org/style-guide/js/), like always using curly brackets for conditional and add some spacing between some character.

PR #1584, opened 2019-08-26
Checkboxes

jmealing:
Have fixed the tab/space issue, and put the sass into alphabetical order

PR #1585, opened 2019-08-26
documentation for checkboxes and radio buttons

Borges-Porteous:


site/pages/gc-form-en.hbs
```diff
@@ -0,0 +1,197 @@
...
+							<label for="checkbox_3">Option 3</label>
+						</li>
+						<li class="checkbox">
+							<input type="checkbox" id="checkbox_4">
+							<label for="checkbox_4">Option 4</label>
+						</li>
+					</ul>
+					<hr>
+					<h6>Small checkboxes</h6>
+					<ul class="list-unstyled gc-form">
```
duboisp: Is the CSS class ```gc-form``` need to be set on the unordered list or it can be set on a parent form element? 

See my related comment https://github.com/wet-boew/GCWeb/pull/1584#pullrequestreview-283837546

site/pages/gc-form-en.hbs
```diff
@@ -0,0 +1,197 @@
...
+	&#x3C;/li&#x3E;
+&#x3C;/ul&#x3E;</code></pre>
+		</div>
+	</div>
+	<h3>Radio buttons</h3>
+	<div class="row">
+		<div class="col-md-4">
+			<div class="panel panel-default">
+				<div class="panel-body">
+					<h5 class="mrgn-tp-0">Appearance</h5>
```
duboisp: The ```h4``` heading is missing here. 

You might want to put the heading inside the ```.panel-heading``` section instead of being inside the ```.panel-body```. That way you wont need to set the margin top to 0.  Like in this example: https://wet-boew.github.io/wet-boew-styleguide/design/panels-en.html#default

site/pages/gc-form-en.hbs
```diff
@@ -0,0 +1,197 @@
...
+		&#x3C;input id=&#x22;radio-option-3&#x22; type=&#x22;radio&#x22; name=&#x22;radio-example&#x22;&#x3E;
+		&#x3C;label for=&#x22;radio-option-3&#x22;&#x3E;Option 3&#x3C;/label&#x3E;
+	&#x3C;/li&#x3E;
+&#x3C;/ul&#x3E;</code></pre>
+		</div>
+	</div>
+</section>
+<section>
+	<h2>Configuration options</h2>
+	<table class="table">
```
duboisp: This could be a simple definition list instead of table where you list all the new CSS class with a small description. I was thinking something like the "CSS class" documentation for the Service and Information component under the "API" section. See: https://wet-boew.github.io/themes-dist/GCWeb/component/gc-srvinfo.html

Btw it is not an issue to use a table to add more info.

site/pages/gc-form-en.hbs
```diff
@@ -0,0 +1,197 @@
...
+			</ul>
+			<h5 class="mrgn-tp-0 text-danger"><span class="glyphicon glyphicon-remove-circle"></span> Incorrect use</h5>
+			<p>Compliance point(s):</p>
+			<ul>
+				<li>Do not use if there is a need for users to select more than one option</li>
+			</ul>
+		</div>
+		<div class="col-md-4">
+			<h5 class="mrgn-tp-0">Code</h5>
+			<pre class="prettyprint prettyprinted" style=""><code><span class="com">// Radio buttons:</span>&#x3C;ul class=&#x22;list-unstyled gc-form&#x22;&#x3E;
```
duboisp: Please use the wet-boew prettify plugin, instead of adding the prettyprint CSS class. 
https://wet-boew.github.io/wet-boew/demos/prettify/prettify-en.html 

Remove the ```style``` empty attribute and I don't think the ```<span class="com">``` is needed. You can put that comment into a paragraph before the preformated text.

site/pages/gc-form-en.hbs
```diff
@@ -0,0 +1,197 @@
...
+	&#x3C;/li&#x3E;
+	&#x3C;li class=&#x22;radio&#x22;&#x3E;
+		&#x3C;input id=&#x22;radio-option-2&#x22; type=&#x22;radio&#x22; name=&#x22;radio-example&#x22;&#x3E;
+		&#x3C;label for=&#x22;radio-option-2&#x22;&#x3E;Option 2&#x3C;/label&#x3E;
+	&#x3C;/li&#x3E;
+	&#x3C;li class=&#x22;radio&#x22;&#x3E;
+		&#x3C;input id=&#x22;radio-option-3&#x22; type=&#x22;radio&#x22; name=&#x22;radio-example&#x22;&#x3E;
+		&#x3C;label for=&#x22;radio-option-3&#x22;&#x3E;Option 3&#x3C;/label&#x3E;
+	&#x3C;/li&#x3E;
+&#x3C;/ul&#x3E;</code></pre>
```
duboisp: Not necessary an issue, more a comment. Usually we encode the character "<" ```&lt;``` and the character ">" ```&gt;``` instead of using the hexadecimal entities number. Double quote don't need to be encoded, but when we do it is like ```&quot;```

site/pages/gc-form-en.hbs
Borges-Porteous: I tried changing prettyprint to wb-prettify but it didn't appear to work.

The example you gave the link for uses class="prettyprint" as well, so I am not sure how to change this one to use wb-prettify

(<section>
--
  | <h3>HTML</h3>
  | <pre class="prettyprint"><code>&lt;form action="#" method="post"&gt;
  | &lt;div&gt;
  | &lt;label for="data1"&gt;Form input default appearance:&lt;/label&gt;
  | &lt;input name="data1" type="text" id="data1" /&gt;
  | &lt;/div&gt;) 

- code from wb-prettify example documentation

I have removed the style and spans and placed it in a <p>

site/pages/gc-form-en.hbs
Borges-Porteous: Can do this

site/pages/gc-form-en.hbs
Borges-Porteous: Fixed heading hierarchy and put appearance in panel heading

site/pages/gc-form-en.hbs
Borges-Porteous: Based on the html entity encoding software I was using. If there's a recommended tool for this, I'd be happy to use that instead. Suggestions?

site/pages/gc-form-en.hbs
Borges-Porteous: Should work on parent elements. I will check with Jennifer and add a description.

site/pages/gc-form-en.hbs
duboisp: Sometime that could be just a setting in your editing tools. But as I said that is not an issue. It is just uncommon to encode the character greater than and lower than by using their corresponding hexadecimal entities.

site/pages/gc-form-en.hbs
duboisp: You can simply add the CSS class name ```wb-prettify all-pre``` to any, usually the first, element or on a div once into your page. Then the plugin will apply the syntax colouring to any ````<pre>``` element

```<div class="wb-prettify all-pre"></div>```

site/pages/gc-form-en.hbs
Borges-Porteous: I'm using a browser based software: mothereff.in/html-entities. Was the first thing I found when searching about how to do those code snippets...

Anyway, if it's not much of an issue, then I'll just leave this one as is for convenience.

site/pages/gc-form-en.hbs
Borges-Porteous: Thanks! I'll try that out,

duboisp:
Replaced by #1709

PR #1586, opened 2019-08-26
Fieldflow

jmealing:
Have fixed the tab/space issue, and run a code beautifier to catch format issues/errors.

If there are still problems, could you point some examples out specifically so I understand what needs to be fixed.


src/plugins/fieldflow/fieldflow.js
```diff
@@ -685,6 +685,11 @@ var componentName = "wb-fieldflow",
 				}
 			}
 		}
+		if ( isInline ) {
+			radCheckOut = "<div class='gc-form'>" + radCheckOut;
```
duboisp: I though the initial intend was to manually add the CSS class in a parent container of fieldflow

src/plugins/fieldflow/fieldflow.js
```diff
@@ -685,6 +685,11 @@ var componentName = "wb-fieldflow",
...
+		} else {
+			radCheckOut = "<ul class='list-unstyled gc-form'>" + radCheckOut;
```
duboisp: Rendering as a list items have some impact on the alignment of the first checkbox with the label. 

Working example with your code: https://universallabs.org/wet/labs/pr-1586/demos/fieldflow/basic-en.html
Downloadable zip file: https://universallabs.org/wet/labs/pr-1586.zip

Screen Shot 2019-09-04 at 2 40 38 PM

jmealing: Whoops - definitely not as intended, will see about fixing it.

src/plugins/fieldflow/fieldflow.js
jmealing: The intention is any pages using fieldflow will be automatically applied so that those pages will not require updating.

GormFrank:
This was completed in #1824

PR #1587, opened 2019-08-27
CSS - Change accent blue to match gcweb menu

StdGit:
Is it ok to change all accent-blue or should I change only .wb-sl ?

The original fix is for
https://gccode.ssc-spc.gc.ca/iitb-dgiit/nw-ws/sgdc-cdts/issues/99


duboisp:
@StdGit what is the reason of changing the the accent blue? Is is a WCAG issue?

duboisp:
@StdGit the gcweb menu follow some color specific requirement, we will check what could be the best solution.

StdGit:
It was to match the color of gc-web-menu with the "skip menu"

duboisp:
Personally I am ok with the change of updating the colour of the skip links to match the blue in the GCWeb theme.

The way this change is suggested, it might impact the colour of a [few other components](https://github.com/wet-boew/GCWeb/search?q=%24accent-blue&unscoped_q=%24accent-blue).

delisma:
@duboisp You can go ahead and merge the PR. We're fine with every other components to adopt the same accent color.

We did notice that the change won't affect the pager buttons at the bottom of the DataTables. Do we have to submit a different issue to have it changed or could it be done within the scope of this PR?
cc @patlaj

PR #1593, opened 2019-09-05
Video example: Use repo's copies of poster images

EricDunsworth:
This is another take on what was done in #1576.

GCWeb contains copies of most (all?) of the WET repo's plugin demo pages, documentation and supporting files. So it makes more sense for the video example page's poster image references to be relative to GCWeb instead of referencing an external repo (wet-boew).

### DIFF
--- a/site/pages/video-en.hbs
+++ b/site/pages/video-en.hbs
@@ -6,7 +6,7 @@
 	"breadcrumb": [
 		{ "title": "Home", "link": "https://www.canada.ca/en.html" }
 	],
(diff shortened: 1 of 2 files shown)


duboisp:
> GCWeb contains copies of most (all?) of the WET repo's plugin demo pages, documentation and supporting files.

Yes, a copy of all. That is how it was initially setup.

---

FYI - There is some conflicting files to resolve

EricDunsworth:
@duboisp Just rebased :)!

PR #1598, opened 2019-09-27
Build: Update grunt-html to ^11.1.1 and revise htmllint ignore rules

EricDunsworth:
Updates grunt-html to ^11.1.1 and adjusts several ignore rules in its associated task (htmllint). A few new ignore rules have been added, whereas many outdated ones have been removed.

This is a port of wet-boew/wet-boew#8728.

**PS:**
I copied the following line comment from wet-boew/wet-boew#8373 for the sake of consistency:
> ``# TODO: Should be removed and fixed now that HTML5 specs updated``

Thanks @nschonni :)

PR #1600, opened 2019-09-27
DO NOT MERGE: Splash page: Remove French meta description var

EricDunsworth:
Fixes the following error in the htmllint:all task (and official HTML validator):
"A document must not include more than one “meta” element with its “name” attribute set to the value “description”."

The WHATWG's HTML spec (which is now the W3C's canonical spec) currently disallows multiple meta descriptions on the same page:
https://html.spec.whatwg.org/multipage/semantics.html#meta-description

The W3C's archived HTML spec used to allow it (via w3c/html#1161). A similar change has yet to be proposed for the WHATWG's spec.

Related to #1598 and #1599.

### DIFF
--- a/site/layouts/splashpage.hbs
+++ b/site/layouts/splashpage.hbs
@@ -7,8 +7,7 @@
 		<meta charset="utf-8"/>
 		<title>{{title}}</title>
 		<meta content="width=device-width, initial-scale=1" name="viewport"/>
-		<meta name="description" content="{{decription-en}}" />
-		<meta name="description" lang="fr" content="{{decription-fr}}" />
+		<meta name="description" content="{{decription}}" />
 		<meta name="dcterms.creator" content="{{creator-en}}" />
 		<meta name="dcterms.creator" lang="fr" content="{{creator-en}}" />
 		<meta name="dcterms.title" content="{{title}}" />
--- a/site/pages/splashpage.hbs
+++ b/site/pages/splashpage.hbs
@@ -2,15 +2,14 @@
 {
 	"title": "Canada.ca",
 	"pageclass": "splash",
-	"description-en": "The Government of Canada website is a single point of access to all programs, services, departments, ministries and organizations of the Government of Canada.",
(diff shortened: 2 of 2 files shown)


EricDunsworth:
Btw this PR will need to be rebased after merging #1599.

EricDunsworth:
I marked this PR as DNM for the time being until I get around to opening a related issue or PR in whatwg/html to get the WHATWG's stance on this.

duboisp:
@bsouster what do you think?

duboisp:
@IsabelleWaltzing

duboisp:
The other language information can be provided by using RDFa and language

GormFrank:
@EricDunsworth FYI, this was completed in #1811

I am closing this PR. Thanks!

EricDunsworth:
@GormFrank Even though #1811 does remove the French description from the splash page template, IMO that doesn't really get to the bottom of the underlying issue that caused me to mark this PR as a draft.

The W3C's obsolete HTML 5.x spec officially allowed pages to have multiple meta descriptions in different languages.

What I wanted to do before bringing this PR out of draft was to open an issue in the WHATWG HTML repo to to get their stance. At the time I opened this PR, no one had ever previously asked the WHATWG about it. IMO from a semantics point of view, it would make sense for the landing page of a multilingual site like Canada.ca to provide multiple sets of metadata in the languages the site supports.

Some older browsers (like Opera and Firefox) were apparently even able to store meta descriptions in their bookmarks and "pick" the right one based on the user's language. So for example, if my browser's language was set to French and I bookmarked Canada.ca's splash page, my bookmarks viewer might show a description column for the Canada.ca site that's written in French. Whereas someone else using an English browser would get it in English. I don't believe any current browsers still store meta descriptions in their bookmarks and aren't sure if anything else is capable of making use of it.

Thoughts? Should this be re-opened?

GormFrank:
@EricDunsworth I don't see a problem in reopening this one or creating a new PR in the future, but I would say first step is to open an issue in the WHATWG HTML repo (like you said). If you have feedback from them and it is a positive one, I would say only then we could tackle it in WET-BOEW.

PR #1603, opened 2019-10-11
Build: Fail on bootlint warnings

EricDunsworth:
This aborts the build if the bootlint task runs into any warnings (in line with how the other linters work). It'll show all issues it found before failing.

Should make it much more obvious if any future changes introduce bootlint warnings.

Partial port of wet-boew/wet-boew#8739 (GCWeb already fails on bootlint errors).

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -778,6 +778,8 @@ module.exports = (grunt) ->
 			all:
 				options:
 					stoponerror: true
+					stoponwarning: true
+					showallerrors: true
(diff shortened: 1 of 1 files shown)


EricDunsworth:
This depends on #1605 to fix pre-existing bootlint errors.

Once that's merged in, I'll revise this PR to add ``stoponwarning: true`` and ``showallerrors: true`` to the ``bootlint:bootstrap4`` task. Then I'll remove DNM from the title.

PR #1607, opened 2019-10-16
Theme: Add hreflang attribute to language toggle links

EricDunsworth:
Brings GCWeb in line with wet-boew (which has been using ``hreflang`` since wet-boew/wet-boew#4288).


duboisp:
For the release note, this change is a minor change.

PR #1608, opened 2019-10-29
Build: Update grunt-bootlint to ^0.12.3

EricDunsworth:
Port of wet-boew/wet-boew#8747

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,22 +19,6 @@
       "integrity": "sha512-41qEJgBH/TWgo5NFSvBCJ1qkoi3Q6ONSF2avrHq1LVEZfYpdHmj0y9SuTK+u9ZhG1sYQKBL1AWXKyLWP4RaUoQ==",
       "dev": true
     },
-    "CSSselect": {
-      "version": "0.4.1",
-      "resolved": "https://registry.npmjs.org/CSSselect/-/CSSselect-0.4.1.tgz",
-      "integrity": "sha1-+Kt+H4QYzmPNput713ioXX7EkrI=",
-      "dev": true,
-      "requires": {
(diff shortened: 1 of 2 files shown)


GormFrank:
At worst maybe we'd need to update to the latest version of bootlint that was still based on bootstrap 3.4.1.

EricDunsworth:
Looks like this was superseded by #1784 (which bumped this dependency to ``^1.1.0``).

But... the main point of this PR was to demonstrate that GCWeb had throngs of ``bootlint`` failures. Those failures are almost certainly still out there. They're not currently causing the build to fail due to the following:
* GCWeb transitioned from Handlebars to Jekyll at some point after this PR was sent in. That left ``bootlint`` with no pre-built HTML files to scan for in the "/dist/" folder.
* The "test" task no longer calls the ``bootlint`` / ``htmllint`` tasks.
* Nothing else in GCWeb's build system calls the ``bootlint`` / ``htmllint`` tasks.

If something gets done to bring GCWeb's versions of the ``bootlint`` / ``htmllint`` tasks back in working order (i.e. scanning fully-generated Jekyll pages), they'll probably report plenty of failures.

EricDunsworth:
[Bootlint was marked as archived](https://github.com/twbs/bootlint) in late 2021 or early 2022. I couldn't find any explanation nor discussions about it anywhere.

As for how that impacts this issue... it's kind of in the eye of the beholder. WET/GCWeb still use Bootstrap 3.x and a Bootlint dependency to this day.

So we'll need to pick between these options:
* Keep the dependency in place for now and fix this issue
* **Or** drop the dependency altogether

Garneauma:
Closing as grunt-bootlint has now been removed from the project with #2528.

PR #1609, opened 2019-11-01
Theme: Reset the .label class' line height

EricDunsworth:
The form font resizing exclusions that were added in #1449 didn't cover the .label class. This adds it in.

As a result, this prevents the form validation plugin's field error messages from overflowing on top of labels and legends under certain circumstances. If a class was set on a label or legend to manipulate its font (such as .h1-.h6), it was causing a significant line-height mismatch between the label/legend and its associated form validation error message. That behaviour was inadvertently introduced in #1449.


duboisp:
You can use ```chkbxrdio-grp``` css class

EricDunsworth:
Thanks for pointing that out :)! Had no idea that class existed. From what I can tell, its existence was never documented anywhere in WET/GCWeb nor the WET style guide.

I might send over a PR to add that class to the form validation plugin's example page at some point...

PR #1631, opened 2020-01-14
Experimental changed to Provisional

namjohn920:
* "Moved to WET-BOEW" is added for bg-cover and bgimg.
* Changed experimental to provisional across all pages.

### DIFF
--- a/site/pages/provisional-en.hbs
+++ b/site/pages/provisional-en.hbs
@@ -1,13 +1,13 @@
 ---
 {
-	"title": "Experimental functionality - GCWeb theme",
+	"title": "Provisional functionality - GCWeb theme",
 	"language": "en",
-	"altLangPrefix": "experimental",
+	"altLangPrefix": "provisional",
 	"breadcrumb": [
 		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/themes-dist/GCWeb/index-en.html" }
 	],
 	"secondlevel": false,
-	"dateModified": "2019-06-10",
+	"dateModified": "2020-01-14",
 	"share": "true"
 }
 ---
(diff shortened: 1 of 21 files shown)


site/pages/provisional-en.hbs
```diff
@@ -105,13 +105,13 @@
...
 		<td>v5.1</td>
-		<td></td>
+		<td>Moved to WET-BOEW</td>
 	</tr>
 	<tr>
 		<th>Plugin <code>[data-bgimg]</code></th>
 		<td>Take the URL value and set it as the background image. To be replaced by CSS4 selector.</td>
 		<td>v5.1</td>
-		<td></td>
+		<td>Moved to WET-BOEW</td>
```
duboisp: Need a note about this deprecated markup  that the onw in wet-boew require the use of "provisional"

src/_provisional.scss
```diff
@@ -1,10 +1,10 @@
 /*
-	Global Experimental styling
+	Global Provisional styling
 
-	- Experimental  CSSspecific for view port are defined in the "views/[view port]" theme CSS
+	- Provisional  CSSspecific for view port are defined in the "views/[view port]" theme CSS
 */
 
-.experimental {
+.provisional {
```
duboisp: We still need to support ".experimental" for a period as defined by the DD-7

site/pages/provisional-fr.hbs
```diff
@@ -105,13 +105,13 @@
 		<th>CSS <code>.bg-cover</code></th>
```
duboisp: Before this table, we will need a info box that explain the deprecated "experimental" CSS and they need to be changed for "provisional"

namjohn920:
> Add a redirect from experimental html page to the provisional page. Use the meta redirect, like DTO did https://github.com/canada-ca/design-system/blob/master/index.html

I changed the experimental.html to provisional.html. So there is no experimental.html in this commit. Should I change it back to experimental.html?

duboisp:
@namjohn920 please fix Travis-ci errors

duboisp:
@GormFrank fyi - we would need to prioritize your PR that are going to fix the build issue because it is become to be a blocker for other pr, like this one.

GormFrank:
We should support both `.expermiental` and `.provisional` for a while before we just move to `.provisional` only.

site/pages/provisional-fr.hbs
namjohn920: we already have this. Do you want something specific for css? I assume this comment was written before i added the code below. 
<div class="alert alert-info">
	<h2><code>.experimental</code> depricated</h2>
	<p><code>.experimental</code> is now deprecated and it will not be supported in 6 months.
	It has to be replaced by <code>.provisional</code></p>
</div>

src/plugins/bgimg/bgimg.js
```diff
@@ -15,7 +15,7 @@
  */
 var $document = wb.doc,
 	componentName = "wb-bgimg",
-	selector = "[data-bgimg]",
+	selector = ".provisioanl[data-bgimg], .experimental[data-bgimg]",
```
GormFrank, line 18:
```suggestion
	selector = ".provisional[data-bgimg], .experimental[data-bgimg]",
```

GormFrank:
Pre-approved during technical review as long as the last change is made.

PR #1638, opened 2020-02-12
Provisional alert

Borges-Porteous:
First step:(:) see if it compiles!

### DIFF
--- a/site/pages/experimental-en.hbs
+++ b/site/pages/experimental-en.hbs
@@ -125,6 +125,19 @@
 		<td>v5.1</td>
 		<td>Supported via wet-boew since v6.0</td>
 	</tr>
+	<tr>
+        <th>CSS:
+            <ul>
+                <li>Danger alert: <code>.alert-danger</code></li>
+                <li>Warning alert: <code>.alert-warning</code></li>
+                <li>Success alert: <code>.alert-success</code></li>
+                <li>Info alert: <code>.alert-info</code></li>
+            </ul>
+        </th>
+        <td>Adopt the beta design of the alert, without a background colour.</td>
+        <td>v6.1</td>
+        <td></td>
+    </tr>
 </tbody>
 </table>
 
@@ -243,3 +256,42 @@ if ( s.getTime() &lt; msT && msT &lt; e.getTime() ) {
 		</div>
 	</div>
 </div>
+
+<h3 id="beta-alert">Beta alert designs</h3>
+<div class="provisional">
+    <div class="alert alert-danger">
+        <h2>Danger alert</h2>
+        <p>Optional alert details.</p>
+    </div>
+    <div class="alert alert-warning">
+        <h2>Warning alert</h2>
+        <p>Optional alert details.</p>
+    </div>
+    <div class="alert alert-success">
+        <h2>Success alert</h2>
+        <p>Optional alert details.</p>
+    </div>
+    <div class="alert alert-info">
+        <h2>Info alert</h2>
+        <p>Optional alert details.</p>
+    </div>
+</div>
+
+<pre><code>&lt;div class="provisional"&gt;
+    &lt;div class="alert alert-danger"&gt;
+        &lt;h2&gt;Danger alert&lt;/h2&gt;
+        &lt;p&gt;Optional alert details.&lt;/p&gt;
+    &lt;/div&gt;
+    &lt;div class="alert alert-warning"&gt;
+        &lt;h2&gt;Warning alert&lt;/h2&gt;
+        &lt;p&gt;Optional alert details.&lt;/p&gt;
+    &lt;/div&gt;
+    &lt;div class="alert alert-success"&gt;
+        &lt;h2&gt;Success alert&lt;/h2&gt;
+        &lt;p&gt;Optional alert details.&lt;/p&gt;
+    &lt;/div&gt;
+    &lt;div class="alert alert-info"&gt;
+        &lt;h2&gt;Info alert&lt;/h2&gt;
+        &lt;p&gt;Optional alert details.&lt;/p&gt;
+    &lt;/div&gt;
+&lt;/div&gt;</code></pre>
--- a/site/pages/experimental-en.hbs
+++ b/site/pages/experimental-en.hbs
@@ -38,107 +38,107 @@
 </details>
 
 <table class="table table-bordered">
-<thead>
-	<tr>
-		<th>Name</th>
-		<th>Description</th>
-		<th class="nowrap">Available in</th>
-		<th>Status</th>
-	</tr>
-</thead>
-<tbody>
-	<tr>
-		<th>CSS <code>.p-0</code></th>
-		<td>Set a padding of 0px</td>
-		<td>v5.1</td>
-		<td>Supported via wet-boew since v6.0</td>
-	</tr>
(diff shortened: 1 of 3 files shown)


src/_experimental.scss
```diff
@@ -28,14 +28,14 @@
 		}
 	}
 
-	// PinkDay color
-	.bg-pnkDy {
-		background-color: #AD005F;
+	// // PinkDay color NA
+	// .bg-pnkDy {
```
duboisp: This should not be in comment, If there a removal it should be submited in a different pull request

src/_experimental.scss
```diff
@@ -28,14 +28,14 @@
...
+	// 	background-color: #AD005F;
 
-		&.well.header-rwd {
-			background-color: #570030;
-		}
-	}
+	// 	&.well.header-rwd {
+	// 		background-color: #570030;
```
duboisp: This should not be in comment, If there a removal it should be submited in a different pull request

src/_experimental.scss
```diff
@@ -159,3 +159,138 @@
...
+			color: #d3080c;
+			content: "\e101";
+		}
+
+		&.alert-bg {
+			background-color: #f3e9e8;
+
+			details > summary {
+				color: #843534;
+				font-weight: 700;
```
duboisp: It could be preferable to use mixins instead of copying the same rule. 

Here an example:
How to define a mixin https://github.com/wet-boew/wet-boew/blob/master/src/plugins/multimedia/_base.scss#L5 
How to use it: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/multimedia/_base.scss#L104

Borges-Porteous: Should I put the mixin variables at the top of the experimental.scss file? Or above the alerts section in the experimental.scss file? Or in the variables.scss/somewhere else in GCWeb?

Borges-Porteous: @duboisp I haven't finished this - would be great if you could advise where to put them and I'll update this tomorrow (Thursday)

Borges-Porteous: I've gone ahead and defined them at the top of the _experimental.scss sheet. We can move them elsewhere if needed...

Should be ready now.

duboisp: The mixin is only used by SASS during the compilation into CSS. So yes it is correct at the top of the file.

src/_experimental.scss
```diff
@@ -159,3 +169,102 @@
 		}
 	}
 }
+
+
+.provisional {
+	.alert-info > :first-child::before,
```
duboisp: I think this block is to undo the old style. Can you add a comment before that said that?

src/_experimental.scss
```diff
@@ -4,6 +4,16 @@
 	- Experimental  CSSspecific for view port are defined in the "views/[view port]" theme CSS
 */
 
+// $alert-green--color: #2b542c;
+$danger-color: #d3080c;
```
duboisp: Add a comment before to mention those new color is for the provisional alert.
For each those new SASS variable, can you prefix them with "provisional" ? It is just to avoid possible conflict with existing code.

src/_experimental.scss
```diff
@@ -4,6 +4,16 @@
...
+$warning-color: #f90;
+$info-color: #269abc;
+$success-color: #278400;
+
+%alert-details-summary {
+	font-weight: 700;
+}
+
 .experimental {
```
duboisp: Do you mind to update this line for:
```
.experimental, .provisional {
```

src/_experimental.scss
```diff
@@ -159,3 +169,102 @@
 		}
 	}
 }
+
+
+.provisional {
```
duboisp: This is perfect, keep it as is

👍

src/_experimental.scss
```diff
@@ -159,3 +169,102 @@
...
+	.alert-info > :first-child::before,
+	.alert-success > :first-child::before,
+	.alert-warning > :first-child::before,
+	.alert-danger > :first-child::before {
+		content: none;
+	}
+
+	.alert {
```
duboisp: Along with the markup change proposed, you can use the parent selector ```&:```

fyi ref. https://sass-lang.com/documentation/style-rules/parent-selector

Borges-Porteous: Not sure how to action this one. 

Did you mean on the .alert-info > :first-child::before, .alert-success > :first-child::before, etc?

Should that be .alert-info, .alert-success, .alert-warning, .alert-danger { &> :first-child::before { ?

site/pages/experimental-en.hbs
```diff
@@ -243,3 +256,46 @@ if ( s.getTime() &lt; msT && msT &lt; e.getTime() ) {
 		</div>
 	</div>
 </div>
+
+<h3 id="beta-alert">Beta alert designs</h3>
+<div class="provisional">
+	<details class="alert alert-danger">
```
duboisp: provisional css class name need to be sibling CSS of alert and not be the parent

site/pages/experimental-fr.hbs
```diff
@@ -242,3 +255,42 @@ if ( s.getTime() &lt; msT && msT &lt; e.getTime() ) {
 		</div>
 	</div>
 </div>
+
+<h3 id="beta-alert">Beta alert designs</h3>
```
delisma: @Borges-Porteous Can we remove the beta mention? Since it's already in a provisional process

src/_experimental.scss
duboisp: The code will look like this
```
<div class="provisional alert alert-warning">
	<h2>Alerte d’avertissement</h2>
	<p>Détails optionnel d'alerte.</p>
</div>
```

And yes add it to the other CSS class name, like "alert-warning, ..."

src/_experimental.scss
```diff
@@ -159,3 +165,103 @@
...
+
+		&::before {
+			color: $provisional-warning-color;
+			content: "\e107";
+		}
+	}
+
+	.alert-danger {
+		border-color: $provisional-danger-color;
+		border-image: linear-gradient(to bottom, $provisional-danger-color 16px, $provisional-danger-color 16px, rgba(0, 0, 0, 0) 16px, rgba(0, 0, 0, 0) 48px, $provisional-danger-color 48px, $provisional-danger-color 48px) 1 100%;
```
duboisp: This would be a nice to have, it will be to try to remove CSS code duplication.

SASS already have a few feature to help to avoid code duplication like
* `@ function` ref.https://sass-lang.com/documentation/at-rules/function
* The share plugin has something similar too. ref. https://github.com/wet-boew/wet-boew/blob/master/src/plugins/share/_base.scss#L7

src/_experimental.scss
EricDunsworth: Addressed in #1650.

site/pages/experimental-fr.hbs
EricDunsworth: Addressed in #1650. I removed all mentions of "beta" with regards to the alerts.

src/_experimental.scss
EricDunsworth: Addressed in #1650. Went with a mixin.

duboisp:
Provisional feature added via #1661

PR #1646, opened 2020-03-03
color for warning icon added

namjohn920:
This is the design pattern that is for 'in case of crisis'

PR #1650, opened 2020-03-13
Provisional: Add new alert designs

EricDunsworth:
Supersedes and closes #1638.

This PR is an enhanced version of #1638. It contains all of the original's commits, plus several additional commits that fix various issues and address leftover feedback.

I created a new PR since I was unable to push extra commits into #1638's branch (``gc-proto:master``).

CC @Borges-Porteous @delisma @duboisp @GormFrank

### DIFF
--- a/site/pages/provisional-en.hbs
+++ b/site/pages/provisional-en.hbs
@@ -283,9 +283,14 @@ if ( s.getTime() &lt; msT && msT &lt; e.getTime() ) {
 
 <h3 id="beta-alert">Beta alert designs</h3>
 <div class="provisional">
+	<details class="alert alert-danger">
+		<summary>Testing danger alert summary</summary>
+		<p>Some text</p>
+	</details>
     <div class="alert alert-danger">
         <h2>Danger alert</h2>
         <p>Optional alert details.</p>
+
     </div>
     <div class="alert alert-warning">
         <h2>Warning alert</h2>
--- a/site/pages/provisional-en.hbs
+++ b/site/pages/provisional-en.hbs
@@ -137,7 +137,19 @@
 		<td>v5.1</td>
 		<td></td>
 	</tr>
-</tbody>
+	<tr>
+        <th>CSS:
+            <ul>
+                <li>Danger alert: <code>.alert-danger</code></li>
+                <li>Warning alert: <code>.alert-warning</code></li>
+                <li>Success alert: <code>.alert-success</code></li>
+                <li>Info alert: <code>.alert-info</code></li>
+            </ul>
+        </th>
+        <td>Adopt the beta design of the alert, without a background colour.</td>
+        <td>v6.1</td>
+        <td></td>
+    </tr></tbody>
 </table>
 
 <h2>Provisional feature documentation</h2>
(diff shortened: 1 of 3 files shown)


src/_experimental.scss
```diff
@@ -4,6 +4,37 @@
 	- Experimental  CSSspecific for view port are defined in the "views/[view port]" theme CSS
 */
 
+// Colours for the provisional alert styles
+// TODO: Rename and migrate these into _variables.scss if the provisional alert styles become official in the future
+$provisional-alert-danger-color: #d3080c;
+$provisional-alert-warning-color: #ee7100;
+$provisional-alert-info-color: #269abc;
+$provisional-alert-success-color: #278400;
```
EricDunsworth: @duboisp Should I scrap these and just directly pass the HEX colours as parameters for the ``provisional-alert-type()`` mixin? These colour variables are no longer being used multiple times.

site/pages/experimental-en.hbs
```diff
@@ -38,94 +39,107 @@
...
+			<td>v5.1</td>
+			<td>Supported via wet-boew since v6.0</td>
+		</tr>
+		<tr>
+			<th>CSS:
+				<ul>
+					<li>Danger alert: <code>.alert-danger</code></li>
+					<li>Warning alert: <code>.alert-warning</code></li>
+					<li>Success alert: <code>.alert-success</code></li>
+					<li>Info alert: <code>.alert-info</code></li>
```
EricDunsworth: @duboisp Should these mention the ``.provisional`` class anywhere? Or is that already implied for all classes this page covers?

src/_experimental.scss
duboisp: it's up to you, but having those variable written here will ease the move when that provisional would be official.

site/pages/experimental-en.hbs
duboisp: It is implied.
Can you do 4 row instead of one?

EricDunsworth: Updated in latest commit.

src/_experimental.scss
EricDunsworth: Ok, will keep them as-is.

duboisp:
@EricDunsworth Dammit we caused a merge conflict when we merged another PR. I might get all the changes and create a new PR later

EricDunsworth:
@duboisp I'll see if I can quickly rebase it again.

duboisp:
ok I will wait

EricDunsworth:
@duboisp @GormFrank Just spent awhile rebasing again and tried another force-push, but it looks like even more merge conflicts have been introduced via other PRs within the past 30-60 mins. Not gonna bother making any further attempts.

duboisp:
ok, I will take care of it. Thanks @EricDunsworth . I know that rebasing become complicated when there is a lot of commits.

EricDunsworth:
@duboisp What happened in this case was that on Thursday COB, #1658 was merged into ``master`` while I was in the middle of fixing the conflicts that were caused by #1646. So even though I fixed the merge conflicts I set out to resolve, my rebased commits had become outdated yet again by the time I force-pushed them onto GitHub.

I could've just squashed all the commits into one to save myself the trouble of manually resolving conflicts spread across 28 commits, but I wanted to retain the original commits in this PR for posterity.

In the future I'd recommend being more careful about merging PRs in the order they were sent in. Particularly when PRs like this one (which has 28 commits in it) are in play.

duboisp:
@EricDunsworth 
There is no actual benefit (from a global/highest project view) to have those 28 commits where it only related to the same thing (add a new alert provisional feature). I understand this is collaborative effort and I ensured the squashed commit message provide credit at both of you. I kept Miguel as the author and I added your name as the co-author.

I can understand there is posterity value regarding to show the evolution of this collaborative effort between two department and all those commits would remain listed in this PR to illustrate that effort and the work involved.

At the end, it was going to be squashed anyway.

PR #1654, opened 2020-04-14
Fix #1634

### LINKED ISSUE #1634: GCWeb forms use a signficiantly smaller font size than its standard font, can't be overridden

```scss
 */
	font-size: 16px;
	line-height: 23px;
}
	padding: 6px 12px;
}
legend {
	line-height: 1.65em;
}
```

jkshapiro:
No obvious reason to format text differently just because it's inside a <form> element.

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -128,20 +128,6 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 }
 
 
-/* Exclude temporary the form element from the font update, keep same style as before
- * - Rational: More testing and various adjustment need to be completed before to apply it
- */
-:not(#wb-srch) form, .checkbox, .checkbox-inline, .radio, .radio-inline, form .btn:not(.btn-call-to-action) {
-	font-size: 16px;
-	line-height: 23px;
-}
-form .btn:not(.btn-call-to-action), .form-control {
-	padding: 6px 12px;
-}
-legend {
-	line-height: 1.65em;
-}
-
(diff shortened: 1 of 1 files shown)


duboisp:
Also we will need an interim solution to let author/developper to use old font-size for existing `form`. The font change would impact some existing form design and the new font might require to redesign those form.

/cc @delisma

jkshapiro:
We are currently in contravention of the Canada.ca design system. This PR fixes that.

From https://design.canada.ca/styles/typography.html:
> The typography guidance is mandatory on all pages.

There is no exception for forms.

delisma:
@jkshapiro In these trying times, I don't think any hard consequences will fall on anybody. I think what @duboisp wants is to unearth the edge cases where the font size change breaks the layout and offer a way to stay on the former text size until these layouts are fixed.

EricDunsworth:
I agree that there could potentially be risks in retroactively increasing the font size of pre-existing forms. But having said that... similar risks existed for increasing the baseline font size on everything else, yet that change was still pushed through. Although I presume that involved a lot of advance testing.

At this point, it's been more than a year since GCWeb's baseline font size was dramatically increased. Anyone that has tried creating new forms on Canada.ca over the past year has had to deal with their forms looking really small compared to "normal" content. No special class is currently available to change form fonts to match the baseline font "on demand".

If there's no appetite or capacity to do any advance research/coordination/etc that might be necessary before merging this PR, could a special class at least be introduced to allow developers to force their forms to use the baseline font?

Related: #1634

Issue #1670, opened 2020-04-22
GC logo gets listed as an WCAG SC 1.1.1 error


bsouster:
@duboisp @GormFrank 

This code is creating a false positive in an automatic accessibility testing tool being used by GC:

    <a href="/en.html" property="url">
        <img src="/etc/designs/canada/wet-boew/assets/sig-blk-en.svg" alt="" property="logo">
        <span id="" class="wb-inv" property="name"> Government of Canada</span>
    </a>

The technique was necessary at the time it was coded but can now be updated with techniques automatic testing tools recognize:

    <a href="/en.html" property="url">
        <img src="/etc/designs/canada/wet-boew/assets/sig-blk-en.svg" alt="Government of Canada" property="logo">
    </a>

There are suggestions that role="img" is necessary for some adaptive technologies but I've yet to find out which AT those are but I'll look around.

bsouster:
There is discussion whether role="img" should be applied to images with SVG src. The origin of this as a recommendation comes from [Deque's research on accessible SVG patterns](https://codepen.io/cariefisher/pen/bOKEVQ).

The standard (img+src+alt) pattern with SVGs does not work with Mobile (iOS): Safari + VoiceOver
(v12.1.2).

@duboisp : you said you want more research on this before applying role fixes

PR #1668, opened 2020-04-21
Slim header

namjohn920:
Issues regarding header are now fixed.
Needs testing.

PR #1673, opened 2020-04-28
Moving GC brand text to img alt

bsouster:
- removed the span with GC text and RDFa name
- placed GC text in logo img alt
- supplemented RDFa name with a meta tag

Note: the img alt text is reflecting the brand text of the chosen language of page only. Logos are exempt from verbosity.

Related: https://github.com/wet-boew/GCWeb/issues/1670

### DIFF
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,7 @@
 {{#unless provisional}}
 	<div class="brand col-xs-5 col-md-4" property="publisher" typeof="GovernmentOrganization">
 		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /></a>
+		<meta property="name" content="{{{i18n "tmpl-gc-sig"}}}" />
 		<meta property="areaServed" typeOf="Country" content="Canada" />
 		<link property="logo" href="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" />
 	</div>
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,6 @@
 {{#unless provisional}}
 	<div class="brand col-xs-5 col-md-4" property="publisher" typeof="GovernmentOrganization">
-		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /></a>
+		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /><span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
 		<meta property="name" content="{{{i18n "tmpl-gc-sig"}}}" />
 		<meta property="areaServed" typeOf="Country" content="Canada" />
 		<link property="logo" href="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" />
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,6 @@
 {{#unless provisional}}
 	<div class="brand col-xs-5 col-md-4" property="publisher" typeof="GovernmentOrganization">
-		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="" property="logo" /><span class="wb-inv" property="name"> {{{i18n "tmpl-gc-sig"}}} / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
+		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /></a>
 		<meta property="areaServed" typeOf="Country" content="Canada" />
 		<link property="logo" href="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" />
 	</div>
--- a/site/layouts/servermessage.hbs
+++ b/site/layouts/servermessage.hbs
@@ -15,8 +15,10 @@
 	<body vocab="http://schema.org/" typeof="WebPage">
 		<header role="banner" id="wb-bnr" class="container">
 			<div class="row">
-				<div class="col-sm-6">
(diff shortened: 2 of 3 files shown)


duboisp:
We are going to sync this change with the PR #1668

EricDunsworth:
@bsouster @duboisp I have some concerns with this PR's changes...

**Questions:**
1. What's the rationale for removing the opposite language from the FIP's textual alternative? If a bilingual FIP image was deemed to be important-enough to present to sighted users, why wouldn't it be equally important to AT users? If the FIP's opposite language is superfluous, why isn't GCWeb using unilingual FIPs?
2. Will these changes really improve accessibility and/or AT support? Or is it only to benefit #1670's unnamed testing tool? If it's solely the latter, IMO it'd be better to adapt the tool to support GCWeb rather than modify the theme to accommodate it.
3. If it's decided to proceed with these changes, they should also be ported over to  the splash page and server message page templates.

**PS:**
The FIP's textual alternative [didn't originally contain the opposite language](https://web.archive.org/web/20140807045240/http://wet-boew.github.io/themes-dist/GCWeb/index-en.html) in most of GCWeb's page templates. I added it in #1241. FWIW at the time @nschonni approved it and @LaurentGoderre didn't seem to be against it.

It looks like the splash page [originally used bilingual ``aria-label`` attributes](https://web.archive.org/web/20150215031211/https://wet-boew.github.io/themes-dist/GCWeb/splashpage.html). Nowadays it uses bilingual ``alt`` attributes (which deviates from all the other page templates).

bsouster:
@EricDunsworth 

1. Bilingualism is a brand value. It is unnecessarily verbose for the context of a web page to communicate brand values. Just as we do not have alternative text for brand personality and attributes (the Maple Leaf, the colour red), we do not need to communicate brand values. 

2. Using the alt for alt text is clearly cleaner and more consistent with regular image alt text practices, and the owner of the tool that was used to report error was written with, and in direct contribution to, the Accessibility Conformance Testing (ACT) rules format 1.0 from W3C. The code this replaces was written as a band-aid. Are we going to suggest their write their tool to include our band-aid? If this is not the right time to clean our code, when will be the right time?

3. Agreed

EricDunsworth:
@bsouster
1. I personally agree that it's verbose and arguably unnecessary. But I think that would be TBS' call.

   If I strictly look at the FIP from a WCAG 2.x perspective, I'd deconstruct the FIP into 3 "pieces":
    * Decorative flag image (Canada flag)
    * Image of English text ("Government of Canada")
    * Image of French text ("Gouvernment du Canada")

   Both the primary and opposite languages are presented equally side-by-side. Neither language is de-emphasized. Sighted users will see both languages no matter what, whether they like it or not. So why shouldn't screen reader users also be presented with bilingual equivalents? I don't see how the opposite language would be any less important to screen reader users vs sighted users. IMO the bilinguism is equally important (or useless) to both kinds of users.

   [WCAG 2.0 success criterion 1.1.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/text-equiv-all.html) and its techniques talk about using text alternatives that serve an equivalent purpose to non-text content. So it could potentially be argued that if the FIP's purpose is solely to convey that the site belongs to the GC, a unilingual text alternative might be enough to fulfill it. But IMO that would be kind of a stretch given that we're talking about plain-looking bilingual text.

   If the bilingual text had been coded as a couple of HTML paragraphs and stylized with CSS, I strongly doubt any extra markup like ``aria-hidden`` would've been used to hide the opposite language from screen readers. So I don't see why there seems to be a double-standard for that content being presented as part of an SVG image.

   Btw I wasn't advocating for literally describing everything in the FIP in its textual alternative. I don't consider its maple leaf flag to be meaningful and would treat it as purely decorative (as is currently being done).

2. I agree that it's generally a better practice to use ``alt`` attributes. But IMO the status quo arguably has an advantage over ``alt`` attributes (although it would no longer apply if the opposite language was removed).

   What's currently being done with the hidden ``<span>`` elements started out is a leftover from back when the FIP image used an ``<object>`` element (for IE8 support). But since the FIP's text alternative is bilingual, the ``<span>`` way allows the opposite language's language to be programmatically-identified. So screen readers can announce each "part" of the FIP in their respective languages. That wouldn't be possible with an ``alt`` attribute.

   IMO it isn't realistic to expect an automated tool to be able to fully evaluate a page for WCAG 2.x compliance. It's practical for some things, but not everything. Especially when it comes to grey zones. If what GCWeb is currently doing with its FIP is WCAG 2.0-compliant and the aforementioned tool is unable to cope with it, IMO that's the tool's problem - not GCWeb's.

EricDunsworth:
Just throwing this out there, but here's another potential way that could be used to markup those kinds of FIP images. Could serve as a compromise of sorts :).

**English HTML pitch:**
```html
<img src="./GCWeb/assets/sig-blk-en.svg" alt="Government of Canada" />
<span class="wb-inv"> / <span lang="fr">Gouvernement du Canada</span></span>
```

That would give the FIP image a meaningful ``alt`` attribute, which would probably please automated testing tools as well as do a better job of acting as a fallback if the image fails or gets disabled in a browser. The visually-hidden ``<span>``s would serve to present the opposite language's content in a programmatically-defined way.

Not sure what the best way of dealing with microdata would be with that kind of structure though...

EricDunsworth:
FYI I've pushed some new commits into this PR's branch. They implemented the code I pitched in my previous reply and port the revised code (including microdata) into the splash and server message page templates.

Btw I left the wordmark images alone for the time being... IMO it might be better to send a separate PR for those. Plus I think more significant underlying changes will be necessary to account for unilingual vs bilingual server message pages (wordmark's text alternative should probably be unilingual in the former and bilingual in the latter... but the server message page layout file isn't setup for that).

EricDunsworth:
Related: #1677

site/includes/brand.hbs
```diff
@@ -1,6 +1,7 @@
 {{#unless provisional}}
 	<div class="brand col-xs-5 col-md-4" property="publisher" typeof="GovernmentOrganization">
-		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="" property="logo" /><span class="wb-inv" property="name"> {{{i18n "tmpl-gc-sig"}}} / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
+		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /><span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
```
duboisp, line 3: Is the "\" need to be excluded from the language change? can we simply have 1 span instead of 2?
Like:
```
<span class="wb-inv" lang="{{#is page.language "en"}}fr"> / {{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></a>
```
The slash is only use as a language change text marker.

What is your thought?

delisma: #noslah

duboisp: @bsouster what do you think if we remove the "/" character in the screen reader only text and keep the language change? like
```
 property="logo" /><span class="wb-inv" lang="{{#is page.language "en"}}fr"> {{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></a>
```

EricDunsworth: @bsouster Any thoughts on this?

PR #1676, opened 2020-05-04
Fix #1634

### LINKED ISSUE #1634: GCWeb forms use a signficiantly smaller font size than its standard font, can't be overridden

jkshapiro:
Temporary fix, until testing is completed on the change of 2018-11-28.

This PR introduces the .force-style-gcweb-current class which content authors can apply to forms, in order to maintain compliance with the Canada.ca design system.

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -136,7 +136,7 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 	font-size: 16px;
 	line-height: 23px;
 }
-form:not(.force-style-gcweb-current) .btn:not(.btn-call-to-action), :not(.force-style-gcweb-current) .form-control {
+form:not(.force-style-gcweb-latest) .btn:not(.btn-call-to-action), :not(.force-style-gcweb-latest) .form-control {
 	padding: 6px 12px;
 }
 :not(.force-style-gcweb-current) legend {
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -130,15 +130,16 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 
 /* Exclude temporary the form element from the font update, keep same style as before
  * - Rational: More testing and various adjustment need to be completed before to apply it
+ * - In the mean time, users can apply the class "force-style-gcweb-current" to the form
  */
-:not(#wb-srch) form, .checkbox, .checkbox-inline, .radio, .radio-inline, form .btn:not(.btn-call-to-action) {
+:not(#wb-srch) form:not(.force-style-gcweb-current), :not(.force-style-gcweb-current) .checkbox, :not(.force-style-gcweb-current) .checkbox-inline, :not(.force-style-gcweb-current) .radio, :not(.force-style-gcweb-current) .radio-inline, form:not(.force-style-gcweb-current) .btn:not(.btn-call-to-action) {
 	font-size: 16px;
 	line-height: 23px;
 }
-form .btn:not(.btn-call-to-action), .form-control {
+form:not(.force-style-gcweb-current) .btn:not(.btn-call-to-action), :not(.force-style-gcweb-current) .form-control {
 	padding: 6px 12px;
 }
-legend {
+:not(.force-style-gcweb-current) legend {
 	line-height: 1.65em;
 }
 
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -132,7 +132,7 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
  * - Rational: More testing and various adjustment need to be completed before to apply it
  * - In the mean time, users can apply the class "force-style-gcweb-current" to the form
  */
-:not(#wb-srch) form:not(.force-style-gcweb-current), :not(.force-style-gcweb-current) .checkbox, :not(.force-style-gcweb-current) .checkbox-inline, :not(.force-style-gcweb-current) .radio, :not(.force-style-gcweb-current) .radio-inline, form:not(.force-style-gcweb-current) .btn:not(.btn-call-to-action) {
+:not(#wb-srch) form:not(.force-style-gcweb-latest), :not(.force-style-gcweb-latest) .checkbox, :not(.force-style-gcweb-latest) .checkbox-inline, :not(.force-style-gcweb-latest) .radio, :not(.force-style-gcweb-latest) .radio-inline, form:not(.force-style-gcweb-latest) .btn:not(.btn-call-to-action) {
 	font-size: 16px;
 	line-height: 23px;
 }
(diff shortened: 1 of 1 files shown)


EricDunsworth:
@jkshapiro Btw you might want to consider adding "Fix #1634" and "Close #1654" to the end of the OP. That'll link the PR to the issue it's fixing and cause it to automatically close your other PR upon merging.

What you have right now would close #1634 upon merging, but wouldn't of linked this PR to it and wouldn't of auto-closed #1654.

**More info:**
[GitHub Help: Linking a pull request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue)

EricDunsworth:
@jkshapiro Oh actually nevermind what I was saying about close syntax. Didn't realize you planned to leave #1654 open and make this PR a stop-gap measure. My bad :(.

duboisp:
Thanks @jkshapiro 
I like the approach, I just have a some concern about the actual CSS class named used because it will create a conflict if the form style change again (size or new flatten bigger checkbox for example). I am will look at it and propose something at the next Technical Review (next week)

duboisp:
Probably we can leverage the provisional feature to test it out and get early feedback on possible design layout issue or design challenge by authors.

jkshapiro:
@EricDunsworth thanks for the tip! You're right about my intention, but you're also right that I didn't know about the magic words. I appreciate the intel.

delisma:
> Thanks @jkshapiro
> I like the approach, I just have a some concern about the actual CSS class named used because it will create a conflict if the form style change again (size or new flatten bigger checkbox for example). I am will look at it and propose something at the next Technical Review (next week)

@jkshapiro Another comes to mind with the class name; when you mean -current, do you mean the latest styles? Which mean, should it be an evergreen class or would you rather have it frozen in time?

jkshapiro:
@delisma evergreen, until #1654 is implemented :-)

delisma:
> @delisma evergreen, until #1654 is implemented :-)

@jkshapiro So we should call it `.force-style-gcweb-latest` instead. It would convey that the styles are always at the cutting edge.

jkshapiro:
Works for me

GormFrank:
Just tested and everything looks good with the class, except for the fact that the file input looks a bit silly with the bigger font:
file-input

delisma:
> Just tested and everything looks good with the class, except for the fact that the file input looks a bit silly with the bigger font:
> file-input
> (Taken from the [Form valid demo](https://wet-boew.github.io/themes-dist/GCWeb/demos/formvalid/formvalid-en.html) page with the class `.force-style-gcweb-current` applied)

The WET-BOEW definitely needs an upload file plugin or custom style.

src/_defaults.scss
```diff
@@ -130,15 +130,16 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 
 /* Exclude temporary the form element from the font update, keep same style as before
  * - Rational: More testing and various adjustment need to be completed before to apply it
+ * - In the mean time, users can apply the class "force-style-gcweb-current" to the form
  */
-:not(#wb-srch) form, .checkbox, .checkbox-inline, .radio, .radio-inline, form .btn:not(.btn-call-to-action) {
+:not(#wb-srch) form:not(.force-style-gcweb-current), :not(.force-style-gcweb-current) .checkbox, :not(.force-style-gcweb-current) .checkbox-inline, :not(.force-style-gcweb-current) .radio, :not(.force-style-gcweb-current) .radio-inline, form:not(.force-style-gcweb-current) .btn:not(.btn-call-to-action) {
```
delisma, line 135:
```suggestion
:not(#wb-srch) form:not(.force-style-gcweb-latest), :not(.force-style-gcweb-latest) .checkbox, :not(.force-style-gcweb-latest) .checkbox-inline, :not(.force-style-gcweb-latest) .radio, :not(.force-style-gcweb-latest) .radio-inline, form:not(.force-style-gcweb-latest) .btn:not(.btn-call-to-action) {
```

src/_defaults.scss
```diff
@@ -130,15 +130,16 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
...
 	font-size: 16px;
 	line-height: 23px;
 }
-form .btn:not(.btn-call-to-action), .form-control {
+form:not(.force-style-gcweb-current) .btn:not(.btn-call-to-action), :not(.force-style-gcweb-current) .form-control {
```
delisma, line 139:
```suggestion
form:not(.force-style-gcweb-latest) .btn:not(.btn-call-to-action), :not(.force-style-gcweb-latest) .form-control {
```

src/_defaults.scss
```diff
@@ -130,15 +130,16 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
...
 	padding: 6px 12px;
 }
-legend {
+:not(.force-style-gcweb-current) legend {
```
delisma, line 142:
```suggestion
:not(.force-style-gcweb-latest) legend {
```

jkshapiro:
Thank you for the review @delisma

src/_defaults.scss
```diff
@@ -130,15 +130,16 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
...
 	line-height: 23px;
 }
-form .btn:not(.btn-call-to-action), .form-control {
+form:not(.force-style-gcweb-latest) .btn:not(.btn-call-to-action), :not(.force-style-gcweb-latest) .form-control {
 	padding: 6px 12px;
 }
-legend {
+:not(.force-style-gcweb-latest) legend {
 	line-height: 1.65em;
 }
```
EricDunsworth, line 144: @duboisp Should this selector be dropped?

GCWeb already contains a selector that sets the ``main`` element's default ``line-height`` to 1.65em:
https://github.com/wet-boew/GCWeb/blob/7b2164a252d4313b00be3d4964668fcf6e0c06f5/src/_defaults.scss#L22

So assuming GCWeb doesn't use any legends outside of its ``main`` element, this selector (both the old and updated versions) is redundant.

EricDunsworth, line 144: @duboisp Should this selector be dropped?

GCWeb already contains a selector that sets the ``main`` element's default ``line-height`` to 1.65em:
https://github.com/wet-boew/GCWeb/blob/7b2164a252d4313b00be3d4964668fcf6e0c06f5/src/_defaults.scss#L22

So assuming GCWeb doesn't use any legends outside of its ``main`` element, this selector (both the old and updated versions) is redundant.

EricDunsworth:
@duboisp What would you think of a selector along these lines?
```css
```

duboisp:
FYI - as discussed at the technical review, @EricDunsworth will check and try a similar but different approach where the need would be still met.

jkshapiro:
Thank you very much @duboisp and @EricDunsworth

PR #1677, opened 2020-05-06
Theme: Update wordmark code and add isBilingual var

EricDunsworth:
This change updates wordmark code for consistency with #1673's FIP changes.

Specifically:
- The splash page's wordmark now uses an English alt attribute and hidden span elements for the French version's text alternative
- Bilingual server message pages now use bilingual textual equivalents for their wordmarks (they previously only used unilingual alt attributes)
- Unilingual server message pages remain as-is (unilingual alt attributes)
- All bilingual pages now contain an extra isBilingual indicator variable
  - Splash page's isBilingual variable is currently unused

CC @bsouster @duboisp

### DIFF
--- a/site/layouts/servermessage.hbs
+++ b/site/layouts/servermessage.hbs
@@ -21,7 +21,7 @@
 					<meta property="areaServed" typeOf="Country" content="Canada" />
 				</div>
 				<div class="col-sm-6">
-					<img id="wmms" src="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" alt="{{{i18n "tmpl-gc-wmms"}}}" />
+					<img id="wmms" src="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" alt="{{{i18n "tmpl-gc-wmms"}}}" />{{#if page.isBilingual}}<span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-wmms" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-wmms" language="en"}}}{{/is}}</span></span>{{/if}}
 				</div>
 			</div>
 		</header>
--- a/site/pages/404-en-fr.hbs
+++ b/site/pages/404-en-fr.hbs
@@ -2,6 +2,7 @@
 {
 	"title": "We couldn't find that Web page (Error 404) - Canada.ca theme / Nous ne pouvons trouver cette page Web (Erreur 404) - Thème Canada.ca",
 	"language": "en",
+	"isBilingual": true,
 	"layout": "servermessage.hbs"
 }
 ---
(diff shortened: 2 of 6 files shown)


site/layouts/servermessage.hbs
```diff
@@ -21,7 +21,7 @@
 					<meta property="areaServed" typeOf="Country" content="Canada" />
 				</div>
 				<div class="col-sm-6">
-					<img id="wmms" src="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" alt="{{{i18n "tmpl-gc-wmms"}}}" />
+					<img id="wmms" src="{{assets}}/../{{site.theme}}/assets/wmms-blk.svg" alt="{{{i18n "tmpl-gc-wmms"}}}" />{{#is page.bilingual true}}<span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-wmms" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-wmms" language="en"}}}{{/is}}</span></span>{{/is}}
```
duboisp, line 24: Change "bilingual" for "isBilingual"

EricDunsworth:
@duboisp Regarding this:
https://github.com/wet-boew/GCWeb/blob/089f406e5c591360ceb67a0fb840c72dbaeb4178/site/includes/brand.hbs#L8-L11

That snippet of code doesn't use an ```` element for its wordmark image. It uses a ``<span>`` element with a ``cndwrdmrk`` class. None of GCWeb's standard page templates use it. The class and the logic that uses it are over 7 years old.

From what I can tell, the only pages that use it are copies of all of GCWeb's standard pages that are outputted to the /provisional/ folder.

**Example:**
https://wet-boew.github.io/themes-dist/GCWeb/provisional/content-en.html

As you can see, the provisional template's header seems to be complete broken. So I don't see any point in porting this PR's logic to something that'll only affect those templates. IMO it'd be better to scrap anything to do with the /provisional/ folder since it no longer seems to serve any purpose.

site/layouts/servermessage.hbs
EricDunsworth: Will force-push an updated PR soon with this change.

EricDunsworth:
Just force-pushed an update with the following changes:
- Renamed the ``bilingual`` variable to ``isBlingual``.
- Replaced ``#is`` with ``#if`` (to avoid having to literally write``true`` in one of the conditions).

As for the unilingual page, I'll create a separate PR for it since IMO it goes beyond this one's scope.

EricDunsworth:
Sorry for the spam, but just wanted to note one other thing.

Even though this PR adds a ``isBilingual`` variable to the splash page, I decided against leveraging it in the handlebars condition for that page's wordmark. That page is bilingual by design. So there isn't any value in its wordmark conditions checking a variable to confirm what's already known.

PR #1682, opened 2020-05-11
feat(Checkboxes): Update input targets on checkboxes and radio-buttons

delisma:
Updating the input type radio and checkbox with bigger target input
radio-buttons-specs


src/plugins/navseq/_base.scss
```diff
@@ -0,0 +1,145 @@
+/*
+	Local experimental styling
+
+	- Experimental CSS specific for view port are defined in the "views/[view port]" theme CSS
+*/
```
delisma, line 5: File shouldn't be in this commit

Gruntfile.coffee
```diff
@@ -793,6 +793,8 @@ module.exports = (grunt) ->
 						"E032" # `.modal-content` must be a child of `.modal-dialog`
 						"W009" # Using empty spacer columns isn't necessary with Bootstrap's grid. So instead of having an empty grid column with `class="col-xs-12"` , just add `class="col-xs-offset-12"` to the next grid column.
 						"E012" # `.input-group` and `.col-*-*` cannot be used directly on the same element. Instead, nest the `.input-group` within the `.col-*-*`
+						"E017" # GCWeb wants to support explicit labels when using checkbox inputs under the .provisional > .checkbox classes
+						"E018" # GCWeb wants to support explicit labels when using radio inputs under the .provisional > .radio classes
```
delisma, line 797: Is there a way to add a bootlint rule for explicit `labels`, to have a `for` attributes that match an `id` on an `input` element?

duboisp: You can check the bootlint website/tool: https://github.com/twbs/bootlint

delisma: @duboisp After reading the Bootlint doc, I need to modify the `bootlint_test.js`, but can't seem to find it in `wet-boew/geweb`. Is this file located in `wet-boew/wet-boew` instead?

StdGit: Is it just a typo perhaps ?
gcweb not geweb

duboisp: Je pense que c'est cette section du gruntfile que tu recherche car c'est les paramètre envoyé à bootlint. Un fichier "coffee" est à la base un fichier javascript.

delisma: @duboisp the coffee file adds exception the bootlint_test.js file adds rules. The coffee file in this PR already adds the exception to make sure travis doesn't complain. But I'm looking into a way to make GCWeb specific bootlint rules too when building plugins or components for Canada.ca.

src/plugins/checkboxes/checkbox.scss
```diff
@@ -0,0 +1,137 @@
...
+			label::before {
+				/* (0,2,2) */
+				border-radius: 50%;
+			}
+		}
+
+		&:hover + span::before,
+		+ label:hover::before {
+			/* (0,3,1) */
+			background-image: linear-gradient(to bottom, hsl(10, 0%, 90%), hsl(10, 0%, 100%) 50%);
```
GormFrank, line 84: You should try out mixins to avoid duplication.

src/plugins/checkboxes/checkbox.scss
```diff
@@ -0,0 +1,137 @@
+.provisional {
```
GormFrank, line 1: This could potentially affect other provisional features related to forms, like wb-postback. Since wb-postback is provisional the form will get those new styles and it could be unwanted.

delisma: I will scope it a little bit more:
```suggestion
.provisional {
   .chckbxrdio-grp label {
```
Would that make sense?

site/pages/xprmntl/checkboxes-en.hbs
```diff
@@ -0,0 +1,194 @@
...
+	<h2>Example 2</h2>
+	<fieldset class="chkbxrdio-grp">
+		<legend>
+		Which is your favorite pet: <small class="text-muted">(checkboxes + implicit label, large box)</small>
+		</legend>
+		<ul class="list-unstyled lst-spcd-2">
+			<li class="radio">
+				<label>
+					<input name="cond13" type="radio" id="cond13-a">
+					<span>Dog</span>
```
duboisp, line 76: General comment,
we try to not have the need to add accessory "span" only for style. If that is needed, it is preferred to do it via a plugin. In the future plugin might get transform to use shadow dom.
If the new style is only feasible for implicit label and not explicit then make that new style only available for implicit label.

src/plugins/checkboxes/checkbox.scss
```diff
@@ -0,0 +1,137 @@
...
+	[type="radio"]:checked + {
+		span::before,
+		label::before {
+			/* (0,3,2) */
+			background: #444;
+		}
+	}
+
+	[type="checkbox"]:checked + {
+		span::after,
```
GormFrank, line 121: Can you try something like this to avoid using a span?

EXPLICIT
`[type="checkbox"]:checked + label::after`

IMPLICIT
`label [type="checkbox"]:checked::after`

src/plugins/checkboxes/checkboxes-en.hbs
```diff
@@ -0,0 +1,232 @@
...
+			<input type="checkbox"
+					id="cond11">
+				<span>You are a resident of Canada</span>
+			</label>
+		</li>
+		<li class="checkbox">
+			<label>
+			<input type="checkbox"
+					id="cond12">
+				<span>You are 15 years of age or older when you apply</span>
```
GormFrank, line 39: Do you need the span here even though it is implicit?

src/plugins/checkboxes/checkbox.scss
GormFrank: Still too broad

src/plugins/checkboxes/checkbox.scss
delisma: Unfortunately, only explicit label can be supported with that approach, because we cannot target a parent element of a checked element with `CSS`

src/plugins/checkboxes/checkboxes-en.hbs
delisma: The `span` is required for implicit labels not for explicit label. It makes it easier to style a child element of a checked element. Will probably use a plugin approach to support implicit labels for checkboxes and radio buttons.

PR #1684, opened 2020-05-20
Theme: Add class to force latest font styles

### LINKED ISSUE #1634: GCWeb forms use a signficiantly smaller font size than its standard font, can't be overridden

EricDunsworth:
- Fixes #1634
- Supersedes #1676
- Related to #1654


delisma:
@EricDunsworth What do you need to remove it from its draft status? Should we include the button sizing issue with this PR too?

EricDunsworth:
@delisma Pretty much just need to find the time to get back to it. The files changed tab has some random todo comments and the like about things that still need to be looked into (including buttons).

Will try to see if I can finally get this sorted out later this week.

delisma:
@EricDunsworth I'll try to give you a hand too and go thru your TODOs. This issue is becoming prominent at TBS.

EricDunsworth:
FWIW I'm gonna try looking into this some more for the next 1-2 hours.

src/_provisional.scss
```diff
@@ -149,12 +149,110 @@
...
+			@extend %gcweb-default-font-size;
+		}
+
+		&.btn:not(.btn-call-to-action), &.form-control,
+		.btn:not(.btn-call-to-action), .form-control {
+			padding: 6px 12px;
+			/*TODO: This conflicts with the below selector that's based on the call to action button's paddinge*/
+		}
+
+		/*Ripoff of call to action button properties to use as baseline for normal buttons*/
```
delisma, line 187: @EricDunsworth Wouldn't it not be better to the `btn-call-to-action` to behave more like WET/Bootstrap button? Maybe resizing the WET/Bootstrap buttons to match the `btn-call-to-action` font-size would prevent conflicts and tons of overrides.

src/_provisional.scss
```diff
@@ -149,12 +149,110 @@
...
+		}
+
+		/*
+		TODOS:
+		-Could any mixins or placeholder selectors from Bootstrap's SCSS be leveraged for provisional button styles?
+		-Come up with a better class name?
+		-Use consistent line breaks in selector rules (e.g. placing each rule on its own line)
+		-Determine whether to base button font sizes and padding on pixels or pixels (call to action uses em, Bootstrap/WET use pixels)
+		-Rework this into two setups: the provisional styles and an opposite "current" class (for anyone that wants to force a tiny font size)
+		-Get rid of the need for having to use ":not(.btn-call-to-action)" everywhere
```
delisma, line 248: I agree

src/_provisional.scss
EricDunsworth: I'm currently experimenting a bit more with button font sizes/padding/etc to see how that might work out in practice. Like how things would be like if call to action didn't exist.

It's trickier than just making each button size use the next highest size's properties though. Bootstrap uses a sliding scale of sorts when increasing font size, padding, etc for bigger button sizes. So it isn't as straightforward as adding like 4px extra to each property for each size increment. I think part of it will be a matter of perception/opinion.

What I find weirdest about call to action buttons is that their font size is 22px... GCWeb's baseline font is 20px (WET is 16px, Bootstrap is 14px). Both WET and Bootstrap re-use their baseline font size for buttons (so normal button text isn't any larger in those).

The one thing about call to action buttons that's kind of grown onto me is text-shadow... it gives off the perception of slightly higher contrast around white text. So I might sneak it into normal buttons that use white text. Unfortunately it doesn't really work out for darker text though (causes the text to get kind of blurry).

delisma: Yes, I've seen sliding scale in the wild recently. Do you think that we could leverage the Bootstrap sliding all the way up to GCWeb? I would much prefer to architect the buttons in GCWeb in a way that we can recreate `btn-call-to-action` from a common architecture, even if at end some design elements are exclusive the CTA button.

EricDunsworth: Whoops - meant to say curve instead of sliding scale :S. What I had in mind was how Bootstrap doesn't seem to increment fonts/padding/etc by a preset amount when going from extra-small to large buttons. Which makes increasing button sizes trickier than it'd of otherwise been. For example, Bootstrap's ``btn-xs`` class uses 1px top/bottom padding, then increases to 5px for ``btn-sm``, 6px for ``btn`` and 10px for ``btn-lg``. There isn't really a preset pattern for how the increments are done. So I have to add my opinion into the mix to try determining what looks best in practice.

Haven't forced-pushed an update yet, but what I'm currently looking into is using ``@include button-size()`` to redeclare all button styles to what I'd consider to be appropriate for GCWeb. For call to action buttons, I'm undoing the non-standard properties (like margins/padding) to basically turn it into a colour class. Also using ``text-shadow`` on all buttons that contain white text (borrowed from call to action).

All of those alternations will require the ``provisional`` class to be used in some way or another. Either directly alongside the current element's classes (e.g. ``<button class="provisional btn btn-default">``) or alongside another provisional class on a parent element (e.g. ``<div class="provisional force-style-gcweb-latest">``).

delisma: I was thinking of [curve](https://type-scale.com/) too. I'll wait for your push. For what I understand you're going in the direction that makes total sense.

src/_provisional.scss
```diff
@@ -149,6 +149,528 @@
...
+TODOS:
+-IIRC disabled inputs are looking darker than expected due to WET/GCWeb overrides
+-I'd like to maybe adjust small/large line heights to be different from GCWeb's default
+-Separate temp Bootstrap selectors from GCWeb-specific ones (e.g. call to action, 11px margin-top for radio buttons, etc...)
+
+*/
+
+
+%gcweb-btn {
+	@include button-size($gcweb-padding-base-vertical, $gcweb-padding-base-horizontal, $gcweb-font-size-base, $gcweb-line-height-base, $gcweb-btn-border-radius-base);
```
GormFrank, line 225: New default (or "md" if you wish) button padding should match the [current CTA pattern](https://wet-boew.github.io/GCWeb/templates/content-en.html#call-to-action). However, I noticed that .54em was better looking than .58em when zooming in on a button used in pair with inline text boxes, so I went with .54em for fielflow buttons.

src/_provisional.scss
```diff
@@ -149,6 +149,528 @@
...
+
+	&.btn-xs {
+		@extend %gcweb-btn-xs;
+	}
+
+	&.btn-call-to-action {
+		@extend %gcweb-btn-call-to-action-margin;
+	}
+
+	&.gc-form-font-new {
```
GormFrank, line 662: It was decided that the class name required to get the increased font size in forms would temporarily be `.gc-font-2019` until it becomes stable, and will not be used in pair with the ".provisional" class.

GormFrank (changes requested):
@EricDunsworth I know it's a draft, sorry for adding Request changes :/

src/_provisional.scss
```diff
@@ -149,6 +149,528 @@
...
+	/*keep post-finalization*/
+	.checkbox,
+	.checkbox-inline,
+	.radio,
+	.radio-inline {
+		@extend %gcweb-main-font-size-line-height;
+	}
+
+	legend {
+		font-size: ($gcweb-font-size-base * 1.5);
```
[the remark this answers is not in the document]
EricDunsworth: My rationale is that's what Bootstrap does lol. Legends only look like regular text when the ``chkbxrdio-grp`` class is used on their parent ``<fieldset>``.

This is a copy of Bootstrap's ``legend`` selector that uses a GCWeb-specific variable instead of Bootstrap's equivalent one:
https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_forms.scss#L25

FWIW that's how most of this PR works. I basically copied all of Bootstrap's form selectors, gutted most of their properties and only kept ones that needed to be re-declared as overrides (primarily those that use variables). Plus some extra selectors to deal with quirks in WET/GCWeb.

As for the design of legends in general, IMO it is strange how Bootstrap decided to make its legends look so big :S. But it should be kept in mind that Bootstrap's base font size is 14px. In that context the legend's font size (21px) is bigger than normal text, but still looks clearly smaller than headings. By comparison, WET increased the base font size to 16px while keeping heading sizes as-is. As a result, WET's legends look almost like high-level headings. GCWeb's legends currently look the same as in WET. But with this PR, they'd balloon to 30px since GCWeb's base font is 20px.

src/_provisional.scss
```diff
@@ -149,6 +149,528 @@
...
+		margin-bottom: $gcweb-line-height-computed;
```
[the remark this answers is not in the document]
EricDunsworth: This selector also comes from Bootstrap and can be suppressed with the ``chkbxrdio-grp`` class:
https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_forms.scss#L24

src/_provisional.scss
EricDunsworth: Ok cool :). I'll update this PR to match in my next set of updates.

Btw is the ``gc-font-2019`` class meant to be used in any situation where the font size needs to be increased? For example, would all of the following coding samples be considered ok to use that class in?
* ``<div class="gc-font-2019">...</div>`` (or any other container element)
* ``<form class="gc-font-2019">...</form>``
* ``<a class="btn btn-default gc-font-2019">...</a>`` (for one-off-buttons... buttons with a parent element that uses ``gc-font-2019`` wouldn't need this)

IIRC the last time I discussed it with @duboisp a couple of months ago, his stance was that he wanted to scrap the dedicated class and make it possible to just throw the ``provisional`` class onto any ``<form>`` element or in combination with the ``btn`` class for one-off buttons. But as you can see in my notes, I had some concerns since I felt like there were still valid use cases where a non-``<form>`` container element might need to be used.

src/_provisional.scss
EricDunsworth: IIRC the button sizing changes this PR makes are meant to be based on the premise of increasing each button style's font size to more or less match one size above what they'd normally be (e.g. GCWeb's ``btn`` class would sorta resemble WET's ``btn-lg`` class, etc...).

The current call to action (CTA) button styles are inconsistent with that. It's not based on the same font sizing scheme or design principles (like margins) as Bootstrap/WET's button styles. So I can't really turn GCWeb's medium button styles into an exact match for CTA without causing a whole lot of other headaches.

For example:
* GCWeb's default font size is 20px whereas the CTA button font size is 22px.
* Generic button classes don't use outer margins, whereas CTA does.
* CTA buttons use text shadows, whereas Bootstrap/WET's buttons don't.

IIRC past discussions I've had about this indicated that I'd be able to pitch changes to the CTA button styles to match whatever generic "medium" button style comes out of this PR. With the only exception being the CTA button's margins (although I was thinking of allowing the margins to be undone if a ``provisional`` class is used on a CTA button). @duboisp and @delisma might need to chime in on this.

Btw I also have many more button size variations to contend with and need to ensure their sizing differences are easy to perceive. I'd of liked to post a screenshot of all my side-by-side button size examples but can't do so atm since I'd need to reset my GCWeb build system to do it lol.

delisma:
@EricDunsworth I will `Reset hard` and apply all your changes on top and `force push` everything back in the PR.

EricDunsworth:
Btw I believe #1976 is (was?) slated to supersede this PR at some point.

That's why I never bothered rebasing this one when GCWeb's repo structure got reworked. Felt like this PR had reached a dead end and was just waiting for the newer PR to close-it off.

PR #1686, opened 2020-05-25
Home page: Have a smaller image for smaller device on the home page with bgimg-srcset

GormFrank:
At 479px wide and down, you get a lighter, more vertical than horizontal image as the background on the home page. This is better adapted for mobile user and it consumes less data for them.

### DIFF
--- a/site/pages/home-en.hbs
+++ b/site/pages/home-en.hbs
@@ -7,14 +7,14 @@
 	"breadcrumb": [
 		{ "title": "Canada.ca", "link": "https://www.canada.ca/en.html" }
 	],
-	"dateModified": "2019-07-02",
+	"dateModified": "2020-05-25",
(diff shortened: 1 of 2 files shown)


GormFrank:
Suggestion to eventually have pixel density as part of the configs, just like the srcset attribute have:
e.g.
```
<img srcset="elva-fairy-320w.jpg,
             elva-fairy-480w.jpg 1.5x,
             elva-fairy-640w.jpg 2x"
```

GormFrank:
Also, recommended widths are the same as our breakpoints, as described here: https://github.com/twbs/bootstrap-sass/blob/bb7dbf8af72b455b51936bc07e51efcaf6220bcc/assets/stylesheets/bootstrap/_variables.scss#L290

PR #1690, opened 2020-06-03
Header element: Fix CSS for Geomap panel to show X button

GormFrank:
As described in #1682 

for minimal impact, I simply removed the `.panel-heading` from the relative positioning equation.
Doesn't seem to break the layout in any way.

### DIFF
--- a/src/banner/_base.scss
+++ b/src/banner/_base.scss
@@ -3,7 +3,9 @@
  */
 
 header {
-	position: relative;
+	&:not(.panel-heading) {
+		position: relative;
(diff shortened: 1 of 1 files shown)


duboisp:
We will need to check the impact, If it is just geomap, just fix it for geomap. Is the "X" added by geomap? Do we need it? Can geomap we leverage overlay instead.

GormFrank:
Opened a PR in WET-BOEW instead: https://github.com/wet-boew/wet-boew/pull/8881

PR #1693, opened 2020-06-04
Build: Remove provisional task/rules/conditions/CSS

EricDunsworth:
Reasoning:
- This task generated modified copies of all top-level pages within a /provisional/ folder.
- The only obvious difference in affected pages was that their headers appeared to be broken.
- Building 78 fewer pages (156 if unmin is counted) is likely to slightly speed up GCWeb's build system.
- The cndwrdmrk class from the removed handlebars condition was only used in provisional page templates. Looks like an old experiment whose goal was to inject the Canada wordmark via CSS.

Note:
- None of these changes affect the /xprmntl/ folder's pages.


EricDunsworth:
Marked this as a draft since the wordmark image the ``cndwrdmrk `` class was referencing is still out there...

site/includes/brand.hbs
```diff
@@ -1,12 +1,6 @@
-{{#unless provisional}}
 	<div class="brand col-xs-5 col-md-4" property="publisher" typeof="GovernmentOrganization">
 		<a href="https://www.canada.ca/{{language}}.html" property="url"><img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /><span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span></a>
 		<meta property="name" content="{{{i18n "tmpl-gc-sig"}}}" />
-		<meta property="areaServed" typeOf="Country" content="Canada" />
+		<meta property="areaServed" typeOf="Countryok" content="Canada" />
```
duboisp, line 4: Countryok is not a valid typeof

duboisp:
So after looking at it, the custom font is not used. We can go forward and remove the custom font CSS and the corresponding asset. Regarding the custom icon that was used for the breadcrumb, it was replaced by an "greater than" symbol when we did the major review GCWeb v5.

src/fonts/_base.scss
```diff
@@ -70,29 +70,3 @@ header {
 		content: "\f103";
 	}
 }
-
```
GormFrank, line 73: I tested the look of the header brand on mobile and desktop with the latest version of GCweb and with the previous major version before that (v7.0.1 to be more precise), with zoom 300% and without, and no perceivable changes what so ever. The styles are in this file at line 31. The file can therefore be removed.

PR #1725, opened 2020-09-15
extract bubble from chatwizard

ctopCanada:
add open action in action manager
add corresponding css for bubble
add tutorial for open attribute.

### DIFF
--- a/src/plugins/actionmng/actionmng.js
+++ b/src/plugins/actionmng/actionmng.js
@@ -2,8 +2,9 @@
  * @title WET-BOEW Action Manager
  * @overview API that coordinate actions with other wet-boew plugin
  * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
- * @author @duboisp
+ * @author @duboisp @wyuan
  */
+
 ( function( $, wb, document ) {
 "use strict";
 
@@ -33,6 +34,7 @@ var $document = wb.doc,
 		"removeClass",
 		"tblfilter",
 		"withInput",
+		"open",
 		"run"
 	].join( "." + actionEvent + " " ) + "." + actionEvent,
 
@@ -163,6 +165,16 @@ var $document = wb.doc,
 		}
 		$elm.addClass( data.class );
 	},
+
+	/* open action start
+	Open overlay action
+	data-wb-doaction='{ "action": "open", "source": "#id" }'
+	*/
+	openAct = function( event, data ) {
+		var $elm = $( data.source || event.target );
+		$elm.trigger( "open.wb-overlay" );
+	},
+
 	remClassAct = function( event, data ) {
 		var $elm = $( data.source || event.target );
 		if ( !data.class ) {
@@ -560,6 +572,9 @@ $document.on( actionMngEvent, selector, function( event, data ) {
 			break;
 		case "addClass":
 			addClassAct( event, data );
+			break;
+		case "open":
+			openAct( event, data );
 			break;
 		case "removeClass":
 			remClassAct( event, data );
--- a/src/plugins/bubble/_base.scss
+++ b/src/plugins/bubble/_base.scss
@@ -1,0 +1,426 @@
+
+/*
+ * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
+ * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * Bubble wizard - Provisional CSS
+ *
+ * ++++++++
+ * This is a Provisional CSS, once stabilized it's include should be done from the theme.scss instead of _provisional.scss
+ * ++++++++
+ *
+ */
+/* Animations */
+@-webkit-keyframes slideInFromRight {
+	0% {
+		-webkit-transform: scale(0, 1);
+		transform: scale(0, 1); }
+	95% {
+		-webkit-transform: scale(0, 1);
+		transform: scale(0, 1); }
+	100% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); } }
+@keyframes slideInFromRight {
+	0% {
+		-webkit-transform: scale(0, 1);
+		transform: scale(0, 1); }
+	95% {
+		-webkit-transform: scale(0, 1);
+		transform: scale(0, 1); }
+	100% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); } }
+
+@-webkit-keyframes pulseIn {
+	0% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); }
+	15% {
+		-webkit-transform: scale(1.15, 1.15);
+		transform: scale(1.15, 1.15); }
+	30% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); }
+	65% {
+		-webkit-transform: scale(1.3, 1.3);
+		transform: scale(1.3, 1.3); }
+	100% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); } }
+
+@keyframes pulseIn {
+	0% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); }
+	15% {
+		-webkit-transform: scale(1.15, 1.15);
+		transform: scale(1.15, 1.15); }
+	30% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); }
+	65% {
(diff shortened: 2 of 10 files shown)


src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+			// Correct bubble positionning on load, on resize an on Y scroll if necessary
+			stickyUntilFooter( $selector );
+
+			$( window ).on( "resize scroll", function() {
+				stickyUntilFooter( $selector );
+			} );
+		} else {
+
+			$( "main" ).after( "<footer id=\"wb-info\"></footer>" );
+			$( "#wb-info" ).addClass( componentName + "-mrgn" );
```
duboisp, line 68: Do line 67 and 68 in 1 call instead of 2. You can set CSS class when creating the footer element at line 67

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		// Close notification aside bubble
+		$( ".notif-close", $selector ).on( "click", function( event ) {
+			event.preventDefault();
+			$( this ).parent().hide();
+			$selector.focus();
+
+			// Do not show notification on next load
+			localStorage.setItem( "wb-chtwzrd-notif", 1 );
```
duboisp, line 79: Rename this for the componentName + "notif" instead of "wb-chtwzrd". I think the intend is to be in the sessionStorage, not in the localStorage

Also, when this is read? It is set but never used. @GormFrank what is the intend? Is it the same situation with the current chat wizard? If so we might need to do some cleanup.

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+
+		// Start initialization
+		// returns DOM object = proceed with init
+		// returns undefined = do not proceed with init (e.g., already initialized)
+		var elm = wb.init( event, componentName, selector ),
+			$elm;
+		if ( elm ) {
+			$elm = $( elm );
+
+			$elm.wrap( "<div class='" + componentName + "-bubble-wrap' id=\"adafds\"></div>" );
```
duboisp, line 98: Remove the id "adafds". If an id is requirement, we will need to use the `wb.getId()` which return an unique identifier

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+
+			var li = document.createElement( "li" );
+			li.className = "wb-slc";
+
+			// Append the Basic HTML version link version
+			var data_wb_doaction_json = JSON.parse( $elm.attr( "data-wb-doaction" ) );
```
duboisp, line 104: Do we need to parse it in JSON in order to re-transform it into string? I think we can take the string value as-is and simply copy it over the other element.

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+
+			if ( data_wb_doaction_json ) {
+				li.innerHTML = "<button  data-wb-doaction='{ \"action\": \"open\", \"source\": " + "\"" + data_wb_doaction_json.source + "\"" + " }' class=\"wb-sl\" >" + $elm.text() + "</button>";
+			}
+
+			// Add link to disable WET plugins and polyfills
```
duboisp, line 110: Please update this comment. It is not exactly right, we are adding it to the top page skip link.

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+			var list = document.getElementById( "wb-tphp" );
+			list.insertBefore( li, list.childNodes[ 0 ] );
+
+
+			var $bubble = $( selector + "-bubble-wrap" );
```
duboisp, line 115: Dont' do a jQuery call when we can get it by getting the parent of the $elm. see: https://api.jquery.com/parent/ or the `parentElement` DOM property

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+	selector = "." + componentName,
+	initEvent = "wb-init" + selector,
+	$document = wb.doc,
+
+	/**
+	 * Initiate chat wizard bubble
+	 * @method initiateBubble
+	 * @param {jQuery DOM element} $selector Element which is the actual bubble
+	 */
+	initiateBubble = function( $selector ) {
```
duboisp, line 26: The function name need to be more descriptive because it is confusing with the other `init` functions. Both are initiatlizing the bubble. You can create multiple sub function with a descriptive name (this include also the code in the init function).

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+			// Initiate chat wizard bubble
+			initiateBubble( $bubble );
+
+			// Identify that initialization has completed
+			wb.ready( $elm, componentName );
+		}
+	};
+
+// Open Chat from the notification message
+$( ".notif", selector ).on( "click", function() {
```
duboisp, line 126: Is this function needed for the bubble? considering that we are going to use the doaction plugin to open up the overlay.

src/plugins/chatwizard/chatwizard.js
```diff
@@ -52,13 +52,13 @@ var componentName = "wb-chtwzrd",
 			"chtwzrd-toggle": "Basculer vers l&apos;assistant",
 			"chtwzrd-notification": "Fermer la notification de discussion",
 			"chtwzrd-open": "Ouvrir l&apos;assistant de discussion",
-			"chtwzrd-minimize": "Réduire l&apos;assistant de discussion",
+			"chtwzrd-minimize": "RÃ©duire l&apos;assistant de discussion",
```
duboisp, line 55: The change in this file must be revert

duboisp (changes requested):
Ensure only the CSS for the bubble is there and the rest is removed

src/plugins/bubble/_base.scss
```diff
@@ -0,0 +1,426 @@
...
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); }
+	65% {
+		-webkit-transform: scale(1.3, 1.3);
+		transform: scale(1.3, 1.3); }
+	100% {
+		-webkit-transform: scale(1, 1);
+		transform: scale(1, 1); } }
+
+@-webkit-keyframes grow {
```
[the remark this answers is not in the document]
ctopCanada: deleted

src/plugins/bubble/bubble.js
duboisp: @ctopCanada as discussed, please use the session storage and take it in consideration when the plugin initiate.

src/plugins/bubble/_base.scss
```diff
@@ -0,0 +1,426 @@
...
+	right: 0; }
+
+button.wb-sl {
+	background: #26374a;
+	border: none;
+	cursor: pointer;
+	font-weight: 700;
+	padding: 5px;
+	width: auto !important;
+	z-index: 501;}
```
duboisp, line 142: FYI - after this PR is done, we will do it permanently in wet-boew core here: https://github.com/wet-boew/GCWeb/blob/e2ed5bcf9fbfd19df23561676fa2b761fb0dd5b6/src/views/_screen.scss#L11

duboisp (changes requested):

src/plugins/actionmng/actionmng.js
```diff
@@ -163,6 +165,16 @@ var $document = wb.doc,
 		}
 		$elm.addClass( data.class );
 	},
+
+	/* open action start
+	Open overlay action
+	data-wb-doaction='{ "action": "open", "source": "#id" }'
```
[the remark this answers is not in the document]
duboisp: as discussed during the weekly code sprint, let's use "open-dialog"

src/plugins/bubble/bubble.js
```diff
@@ -0,0 +1,136 @@
...
+			} );
+		} else {
+
+			$( "main" ).after( "<footer id=\"wb-info\"></footer>" );
+			$( "#wb-info" ).addClass( componentName + "-mrgn" );
+
+		}
+
+		// Close notification aside bubble
+		$( ".notif-close", $selector ).on( "click", function( event ) {
```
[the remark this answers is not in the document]
duboisp: After discussion, the notification is for the first time the bubble it's show to the user.

Rename "notification" for "welcome" message.
We will need a way to make it reappear after like 30 days or somethings or use session storage.
We will need to find a way to be able to have multiple bubble message, Like the one from the chat wizard is hidden, but the one from the invitation exit survey will be displayed at least once. We can probably make it configurable through the plugin.
To reproduce and display the message, open the chat wizard in a private browser windows.
Use the text of the button as the welcome message to display.

PR #1740, opened 2020-10-14
Add Invitation Manager version 1.0 

djamilab:


duboisp:
There is some linting error, see: https://travis-ci.org/github/wet-boew/GCWeb/builds/735815081#L540

Also, we need to revisit the design pattern to invite the user to the server. We would might need to change the pop-up for an in-page banner. We should avoid to manipulate the focus.

duboisp (changes requested):
Please remove the consoleLog and if those are absolutely required, please write test case instead with the expected result.

The whole management of the dialog/overlay need to be removed. there is too much focus manipulation and a different design pattern to show the survey message need to be use, like a banner.

This is an incomplete review.

src/plugins/invitationmanager/demo/im.json
```diff
@@ -0,0 +1,179 @@
+﻿{
+  "settings":
+  {
+    "duration-delay": 15,
+    "page-probability": 0.7,
+    "site-probability": 0.3,
+    "blacklist": [{"url" : ""}],
+    "whitelist": [{ "url" : "^localhost\/"}]
```
GormFrank, line 8: Can you please add the GCWeb website hosted by GitHub in the whitelist for the purpose of demoing?

**wet-boew.github.io**

GormFrank (changes requested):

src/plugins/invitationmanager/im-doc-en.hbs
```diff
@@ -0,0 +1,151 @@
...
+}
+</code></pre>
+</details>
+<p>The im.json file contains the data and configuration of GC Invitation Manager, by opening the file you will be able to adjust and customize the following settings:</p>
+<ul>
+<li><p><code>duration-delay</code> value contains the minimal number of days a visitor will see their next popup invitation when visiting your site(s).</p>
+<p>For example, put 15 for fifteen days.</p>
+</li>
+<li><p><code>page-probability</code> value contains the rolling probability a visitor will see page level popup invitation.</p>
+<p>For example, put 0.7 for 70%.</p>
```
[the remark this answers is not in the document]
duboisp: And the type of value. ex: String, URL, number, Array of strings

src/plugins/invitationmanager/gcim.js
```diff
@@ -0,0 +1,720 @@
...
+
+					//convert the value for the site into RegExp non-case sensitive
+					var b = new RegExp( sublist[ c ].site, "i" );
+
+					//if the url matches the regEx
+					if ( b.test( url ) ) {
+
+						// Store this survey's id and probability in the new object
+						// (this object is later fed to getWeightedRandom)
+
```
duboisp, line 488: Remove the empty line.

src/plugins/invitationmanager/gcim.js
```diff
@@ -0,0 +1,720 @@
...
+						// Convert site survey probability into rate since it is
+						// between 0 and Visitor Allocation for site wide type
+
+						surveySubList[ survey_i.id ] = survey_i.probability * surveyDB.settings[ "site-probability" ];
+
+
```
duboisp, line 494: Only one 1 empty line not 2

src/plugins/invitationmanager/gcim.js
```diff
@@ -0,0 +1,720 @@
...
+	},
+
+
+	/*
+	* Display the popup given the survey parameters
+	*/
+	invite = function( survey ) {
+
+		var html =
+		"<aside id='gc-im-popup' class='wb-overlay wb-overlay-im modal-content-im overlay-def wb-popup-im ' tabindex='0'>" +
```
duboisp, line 544: @djamilab Add 1 more tab indentation. 

@GormFrank eslint usually grab that kind of error, we might require a config change in our build script

duboisp (changes requested):
We will need to re-think on how the plugin start. Adding the div to open up the survey on every pages is might the best approach. 
The init function will be better if it is based on when WET isReady, like:
```
$( document ).on( "wb-ready.wb", init);
```
but "wb-ready" string is a saved in a variable (initEvent) in the top of the plugin code.

---

We need a way to specify a global configuration, instead of adding the URL on every pages.
- Explore how we can modify WET to support centralized data file location. Like a core JSON configuration that will be located along with wet-boew core file.
- We will need to develop a kind of central data base in wet, like around those line here: https://github.com/wet-boew/wet-boew/blob/master/src/core/wb.js#L85
Like having wet core to download a centralized JS config file, and then this plugin will retreive the path for the invitation manager.

---

For consolelog, try to identify the case where it cause an issue and use "throw" keyword to display the error in the user console.

---

Do the indentation in the *.hbs file

GormFrank (changes requested):

PR #1743, opened 2020-10-16
Feature: responsive table

jmealing:
- Turns tables into cards items when on mobile (xxs and xs) view
- Optional responsive class to create cards on tablet (sm) view for tables that have extensive word content that should break into cards at this viewport too
- Optional class to make font size smaller for lengthy worded table content (just playing with it, may remove)

### DIFF
--- a/site/pages/component/gc-table-en.hbs
+++ b/site/pages/component/gc-table-en.hbs
@@ -1,0 +1,188 @@
+---
+{
+	"altLangPrefix": "gc-table",
+	"breadcrumb": [
+		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/themes-dist/GCWeb/index-en.html" },
+		{ "title": "Provisional functionality", "link": "https://wet-boew.github.io/themes-dist/GCWeb/provisional-en.html" }
+	],
+	"dateModified": "2020-10-16",
+	"description": "Working examples for tables - provisional feature",
+	"language": "en",
+	"parentdir": "component",
+	"tag": "table",
+	"title": "Tables - Provisional feature"
+}
+---
+{{>alert-provisional-demos}}
+<span class="wb-prettify all-pre hide"></span>
+
+<dl class="horizontal">
+	<dt>Status</dt>
+	<dd>Provisional</dd>
+	<dt>Type</dt>
+	<dd>Canada.ca design pattern</dd>
+	<dt>Last review</dt>
+	<dd>2021-02-01</dd>
+	<dt>Conforming to</dt>
+	<dd>Content and IA spec 2.1</dd>
+	<dt>Guidance</dt>
+	<dd>Tables - Canada.ca design pattern [coming soon]</dd>
+	<dt>Technical name</dt>
+	<dd>
+		<ul>
+			<li><code>gc-table</code></li>
+		</ul>
+	</dd>
+</dl>
+
+<h2>Working example</h2>
+
+<p>A simple CSS solution to that turns tables into responsive cards for smaller viewports.</p>
+
+<p>This pattern currently is:</p>
+<ul>
+	<li>only for basic/simple tables</li>
+	<li>not supported by IE 11</li>
+</ul>
+
+<h3>Responsive cards</h3>
+<table class="provisional gc-table table" id="myTable1">
+		<caption>Population growth in Canadian cities</caption>
+		<thead>
+		<tr>
+			<th>City</th>
+			<th>Population in 2007</th>
+			<th>Population in 2017</th>
+			<th>Percentage change</th>
+		</tr>
+		</thead>
+		<tbody>
+		<tr>
+			<td data-label="City">Toronto</td>
+			<td data-label="Population in 2007">5,418,207</td>
+			<td data-label="Population in 2017">6,346,088</td>
+			<td data-label="Percentage change">17.1%</td>
+		</tr>
+		<tr>
+			<td data-label="City">Montréal</td>
+			<td data-label="Population in 2007">3,714,846</td>
+			<td data-label="Population in 2017">4,138,254</td>
+			<td data-label="Percentage change">11.4%</td>
+		</tr>
+		<tr>
+			<td data-label="City">Vancouver</td>
+			<td data-label="Population in 2007">2,218,134</td>
+			<td data-label="Population in 2017">2,571,262</td>
+			<td data-label="Percentage change">15.9%</td>
+		</tr>
+		<tr>
+			<td data-label="City">Ottawa–Gatineau</td>
+			<td data-label="Population in 2007">1,188,073</td>
+			<td data-label="Population in 2017">1,377,016 </td>
+			<td data-label="Percentage change">15.9%</td>
+		</tr>
+		</tbody>
+	</table>
+
+<h3>Responsive cards with text left on one item</h3>
+
+<p><span class="label label-info">Optional</span></p>
+
+<p>The class <code>text-left</code> can be added to any <code>&lt;td&gt;</code> so that it aligns to the left side on the cards.</p>
+<table class="provisional gc-table table" id="myTable2">
+		<caption>Telephone numbers for personal taxes, benefits and trusts</caption>
+		<thead>
+		<tr>
+			<th>Reason for your call</th>
+			<th>Number</th>
+			<th>Hours</th>
+		</tr>
+		</thead>
+		<tbody>
+		<tr>
+			<td data-label="Reason for your call">
+				<ul class="text-left">
+					<li>Item one</li>
+					<li>Item two</li>
+				</ul>
+			</td>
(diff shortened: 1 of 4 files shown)


src/views/_screen-md-min.scss
```diff
@@ -1,4 +1,18 @@
 /*
-  WET-BOEW
-  @title: Medium view and over (screen only)
+	WET-BOEW
+	@title: Medium view and over (screen only)
  */
+ 
+.provisional {
+	&.gc-table {
+		.tbl-small-txt {
```
GormFrank, line 8: What element would be needed for the **.tbl-small-txt** class to be used if it is a child of the `<table>` but a parent of the `<thead>`?

src/views/_screen-xs-max.scss
```diff
@@ -10,3 +10,58 @@
 		width: 100%;
 	}
 }
+
+/* Responsive tables - for Tablet Cards - optional */
+.provisional {
+	&.gc-table {
+	.tblt-crds {
```
GormFrank, line 17: Same question here but for the `.tblt-crds` class (and even `.left-xs`). Is there a Javascript plugin that does all the transformations here and that is not included in this PR or does a publisher/web author would need to put all the classes at the right spots?

A working example would definitely help seeing everything working together. Could you please add an **.hbs** file that contains such example in the following folder? https://github.com/wet-boew/GCWeb/tree/master/site/pages/component

GormFrank (commented):
@jmealing have you heard of Tablesaw? https://www.filamentgroup.com/lab/tablesaw.html

I am guessing that what you are trying to accomplish is kind of the "Stack" option from Tablesaw? That would be a great start for simple tables. Just take a few minutes to look at the other options from Tablesaw too, it would be great to implement some of them later on to also take care of more complex tables. IT is to be expected however that some tables may be too complex for even being responsive.

Thank you for your contribution! 🥇

src/views/_screen-md-min.scss
jmealing: This is a PR draft - I just brought in my original CSS and converted - I'm still working on it

src/views/_screen-xs-max.scss
jmealing: This is a PR draft - I'm still working on it, there's likely not even going to be a need for the tablet cards because my initial breakpoints in my prototype were different than what GCWeb actually uses.

I am not planning a javascript plugin at this time.

I do have the working example .hbs file - but it's incomplete

jmealing:
Yeah - that would be the idea - something like Tablesaw.

This provisional pattern is a quick CSS solution to be used for simple tables as a MVP, with plans to expand and improve for more uses later.

src/views/_screen-md-min.scss
jmealing: @GormFrank I do have basic documentation of how the classes would work - but be warned it is very outdated. I have to finish converting it all over and see what makes sense and what needs to be restructured. 

David E has also helped with the naming convention so some classes have changed name.

https://design.canada.ca/components/tables-alpha-demo.html

GormFrank: I personnally very much like the look and semantics from what I can see in the example on design.canada.ca!

duboisp:
for your reference: here an implementation of tablesaw in WET - http://universallabs.org/wet/labs/tablessaw/demo1/stack.php

src/views/_screen-sm-max.scss
```diff
@@ -61,3 +61,56 @@ h2,
...
+			}
+		}
+	}
+		tr {
+			border: 1px solid #ddd;
+			display: block;
+			margin-bottom: .625em;
+			padding: .35em;
+		}
+		border: 0;
```
GormFrank, line 88: Should this be here, in the middle of children selectors? Declared here will apply the rule on `.gc-table` so it could be at line 68.

site/pages/component/gc-table-en.hbs
```diff
@@ -0,0 +1,180 @@
...
+
+<p>Responsive cards</p>
+<table class="provisional gc-table table" id="myTable">
+		<caption class="wb-inv">Population growth in Canadian cities</caption>
+		<thead>
+		<tr>
+			<th scope="col">City</th>
+			<th scope="col">Population in 2007</th>
+			<th scope="col">Population in 2017</th>
+			<th scope="col">Percentage change</th>
```
duboisp, line 48: Can you remove the scope attribute?

site/pages/component/gc-table-en.hbs
```diff
@@ -0,0 +1,180 @@
...
+		</tr>
+		</thead>
+		<tbody>
+		<tr>
+			<td data-label="Reason for your call">
+				<span class="left-xxs">
+					<ul>
+						<li>Item one</li>
+						<li>Item two</li>
+					</ul>
```
duboisp, line 96: Try to move the ".left-xxs" class to be set on the `<ul>` element and remove the `<span>`

site/pages/component/gc-table-en.hbs
```diff
@@ -0,0 +1,180 @@
...
+		<ul>
+			<li><code>gc-table</code></li>
+		</ul>
+	</dd>
+</dl>
+
+<h2>Working example</h2>
+
+<p>Responsive cards</p>
+<table class="provisional gc-table table" id="myTable">
```
GormFrank, line 41: Give a different `id` value on each table.

src/_provisional.scss
```diff
@@ -286,6 +286,29 @@
 		}
 	}
 
+	/* Responsive table - part of design, not responsive part */
+	&.gc-table  {
+		caption {
+			text-align: left;
+		}
+		&.table-small-txt {
```
duboisp, line 294: Leverage the "small" CSS that is included in bootstrap. And then just add the variation you need for the `<tr>` or `<thead>`

site/pages/component/gc-table-en.hbs
```diff
@@ -0,0 +1,180 @@
...
+				</span>
+			</td>
+			<td data-label="Number">1-XXX-XXX-XXXX</td>
+			<td data-label="Hours">Always open (automated line)</td>
+		</tr>
+		</tbody>
+	</table>
+
+<p>Responsive cards with smaller text option</p>
+<table class="provisional gc-table table table-small-txt" id="myTable">
```
GormFrank, line 142: Investigate the existing `.small` class vs the `.table-small-txt`

src/_provisional.scss
```diff
@@ -286,6 +286,29 @@
...
+			}
+		}
+		td {
+			ul {
+				padding-inline-start: 20px;
+			}
+		}
+		.nobreak {
+			white-space: nowrap;
+		}
```
duboisp, line 309: There is already a CSS class called `nowrap`, please remove

site/pages/component/gc-table-en.hbs
```diff
@@ -0,0 +1,180 @@
...
+			<li><code>gc-table</code></li>
+		</ul>
+	</dd>
+</dl>
+
+<h2>Working example</h2>
+
+<p>Responsive cards</p>
+<table class="provisional gc-table table" id="myTable">
+		<caption class="wb-inv">Population growth in Canadian cities</caption>
```
duboisp, line 42: One or the other
* remove the caption element and ensure the previous paragraph/heading explain the table context
* Display the caption element

GormFrank (changes requested):

duboisp (changes requested):
An eventual upgrade will be to auto-populate the `data-label` by using some JS.

Might revisit the css name "left-xss" to something more meaningful

src/views/_screen-sm-max.scss
jmealing: Doesn't seem to actually do anything, despite my notes of what it's for, I've removed it.

src/views/_screen-sm-max.scss
```diff
@@ -60,4 +60,54 @@ h2,
...
 	}
+
+	/* Responsive tables cards */
+	&.gc-table {
+		.table-align-left {
+			clear: both;
+			display: block;
+			text-align: left;
+		}
+		>tbody {
```
duboisp, line 71:
```suggestion
		>tbody, >tfoot {
```

src/views/_screen-sm-max.scss
```diff
@@ -60,4 +60,54 @@ h2,
...
 		}
 	}
+
+	/* Responsive tables cards */
+	&.gc-table {
+		.table-align-left {
+			clear: both;
+			display: block;
+			text-align: left;
+		}
```
duboisp, line 70:
```suggestion
		.text-left {
			clear: both;
			display: block;
		}
```

duboisp (changes requested):
Other than those 2 changes, it's all good

duboisp (approved):

GormFrank (approved):

PR #1755, opened 2020-11-24
GC subway: Add noscript support

EricDunsworth:
The subway menu previously didn't appear at all when JavaScript was disabled.

This addresses it by only hiding the menu if the "js" class is detected and if the page isn't in basic HTML mode. The menu is subsequently shown once the "no-blink" class is injected.

It should be noted that this method can cause a quick flash of the pre-JS subway since it depends on a JS-injected class.

Couldn't come up with a better solution due to the following:
* Unlike WET, GCWeb lacks its own noscript file. Adding one would necessitate HTML markup changes to all page templates...
* The scripting CSS media query can be used to achieve the same purpose, but no browsers currently support it.

CC @delisma

### DIFF
--- a/src/plugins/gc-subway/_screen-md-min.scss
+++ b/src/plugins/gc-subway/_screen-md-min.scss
@@ -13,13 +13,9 @@
 	&.gc-subway {
 		border-right: 0px;
 		border-top: 0px;
-		display: none;
 		margin-top: 25px;
 		padding-left: 15px;
 
-		&.no-blink {
-			display: block;
-		}
 		.gc-subway-menu-nav {
 			float: right;
 			width: 33.33%;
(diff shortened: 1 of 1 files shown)


EricDunsworth:
Marked this as a draft for the time being in case anyone's open to introducing a noscript file into GCWeb or if there might be any other tricks out there that I hadn't considered.

duboisp:
We would need to re-evaluate later.

As eric said, a better solution would be to add a `<noscript>` tag where it will link a  CSS specialized for noscript situation.
Considering that no-javascript are not in scope to meed WCAG, we will need to reevalute this. 

The issue is, coded like it défait the no-blink feature added to gc-subway. We might need to look at alternative.

EricDunsworth:
Related PR: #2360 (fulfills the same goal as this PR set out to... albeit without leveraging a noscript file)

Might close this PR once that one gets merged. I don't get the vibe that there's any appetite to officially add a noscript file to GCWeb.

Although... the [``scripting`` media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/scripting) is now supported by all major browser engines. So it's now technically possible to achieve the same effect as ``noscript`` _without_ the namesake element/file or special class.

CC @Garneauma

duboisp:
Closing this issue as an alternative solution has been merged via #2360

PR #1756, opened 2020-12-03
Feature/gc checklist

delisma:


duboisp (changes requested):
Consider to use HTML5 native checkbox element instead of using ARIA.
3rd state is also available according to HTML5 spec - https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)

duboisp:
Approved upon conflict is fixed, second quick code review to ensure it require ".provisional" and after quick testing locally. Once merged or just before, we will need to update the provisional main page to include this new plugin.

src/_provisional.scss
```diff
@@ -285,6 +289,138 @@
 			}
 		}
 	}
+
+	&.gc-checklist {
```
delisma, line 293: Need to find a way to optimize the SCSS code from here

delisma:
> Approved upon conflict is fixed, second quick code review to ensure it require ".provisional" and after quick testing locally. Once merged or just before, we will need to update the provisional main page to include this new plugin.

Would it be possible to have a selector like `:not(form) .provisional.gc-checklist` if we don't want people to use this plugin inside forms?

duboisp:
> Would it be possible to have a selector like `:not(form) .provisional.gc-checklist` if we don't want people to use this plugin inside forms?

If it works as expected with our officially supported browser, then yes.

src/plugins/gc-checklist/_base.scss
```diff
@@ -0,0 +1,115 @@
...
+				height: 24px;
+				width: 24px;
+			}
+
+			&[aria-checked="true"]::after {
+				border-width: 0 4px 4px 0;
+				height: 16px;
+				width: 9px;
+			}
+		}
```
duboisp, line 107: Don't style aria code, there is even not an example.

src/plugins/gc-checklist/_base.scss
```diff
@@ -0,0 +1,115 @@
...
+
+			&:focus::before {
+				box-shadow: 0 0 0 2px #000, 0 0 8px 4px rgb(59, 153, 252);
+			}
+		}
+
+		.lst-none ul li {
+			list-style-type: disc;
+
+		}
```
duboisp, line 83: I didn't saw a working example that require this style.

src/plugins/gc-checklist/gc-checklist-en.hbs
```diff
@@ -0,0 +1,173 @@
...
+		</li>
+		<li class="checkbox">
+			<input type="checkbox" id="cond2">
+			<label for="cond2">[Condition 2]</label>
+		</li>
+		<li class="checkbox">
+			<input type="checkbox" id="cond3" required>
+			<label for="cond3">[Condition 3] (required)</label>
+		</li>
+		<li class="feedback"><span class="label label-danger">[You need all of the required criteria to call]</span></li>
```
duboisp, line 55: This list item are not similar to it's sibling. Please move it as paragraph after the unordered list or remove it.

src/plugins/gc-checklist/gc-checklist-en.hbs
```diff
@@ -0,0 +1,173 @@
...
+		</li>
+		<li class="checkbox-sm">
+			<input type="checkbox" id="cond5" required>
+			<label for="cond5">[Condition 5] (required)</label>
+		</li>
+		<li class="checkbox-sm">
+			<input type="checkbox" id="cond6" required>
+			<label for="cond6">[Condition 6] (required)</label>
+		</li>
+		<li class="feedback"><span class="label label-danger">[You need all of the above to call]</span></li>
```
duboisp, line 74: This list item are not the same kind compared to it's sibling. Please move it as paragraph after the unordered list.

src/plugins/gc-checklist/_base.scss
```diff
@@ -0,0 +1,115 @@
...
+				content: "";
+			}
+
+			&:active::before,
+			&[aria-checked="false"]:hover::before {
+				background-image: linear-gradient(to bottom, hsl(300, 3%, 73%), hsl(300, 3%, 93%) 30%);
+			}
+
+			&[aria-checked="true"] {
+
```
duboisp, line 57: Don't style aria code, there is even not an example.

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,64 @@
...
+			for ( var i = 0; i < $elms.length; i++ ) {
+				var domNode = $elms.get( i );
+				domNode.tabIndex = 0;
+			}
+			wb.ready( $elm, componentName );
+		}
+	};
+
+// Add your plugin event handler
+$document.on( "click", selector, function( event ) {
```
duboisp, line 43: This will be trigger when anything inside the selector is clicked. Add a more specific selector to avoid conflict with other inner click event.

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,64 @@
...
+		data = JSON.parse( $elm.attr( "data-" + componentName ) );
+
+	if ( data ) {
+		if ( $elm.find( "[required]" ).length === $elm.find( "[required]" ).filter( ":checked" ).length ) {
+			$elm.find( data.feedbackTarget ).text( data.feedbackSuccess );
+			$elm.find( data.feedbackTarget ).addClass( "label-success" ).removeClass( "label-danger" );
+		} else {
+			$elm.find( data.feedbackTarget ).text( data.feedbackFailure );
+			$elm.find( data.feedbackTarget  ).addClass( "label-danger" ).removeClass( "label-success" );
+		};
```
duboisp, line 55: Do variable caching to avoid to call multiple time the same function. This will improve the code readability and be more optimal.

Like:

```
var $feedbackTarget = $elm.find( data.feedbackTarget );
...
			$feedbackTarget.text( data.feedbackSuccess );
			$feedbackTarget.addClass( "label-success" ).removeClass( "label-danger" );
...
```

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,64 @@
...
+			$elm;
+
+		if ( elm ) {
+			$elm = $( elm );
+
+			var $elms = $elm.find( $( "input[type='checkbox']" ) );
+
+			for ( var i = 0; i < $elms.length; i++ ) {
+				var domNode = $elms.get( i );
+				domNode.tabIndex = 0;
```
duboisp, line 36: Is the checkbox already focus-able element?

duboisp (changes requested):
The CSS duplicate would need to be merged. A single file or into the main provisional file is fine, but not both. Also both style are not the same.

src/plugins/gc-checklist/gc-checklist.js
delisma: @duboisp Back when they were `div` elements they were. But now since they piggy-back on the `input[type=radio]`, do they have the focus state by default?

src/plugins/gc-checklist/_base.scss
delisma: You're right this file should be removed now.

src/plugins/gc-checklist/gc-checklist.js
delisma: I could but I need to access the data attribute of the currently clicked selector.

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,65 @@
...
+			}
+			wb.ready( $elm, componentName );
+		}
```
GormFrank, line 39:
```suggestion
		var elm = wb.init( event, componentName, selector );

		if ( elm ) {
			wb.ready( $( elm ), componentName );
		}
```

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,65 @@
...
+	};
+
+// Add your plugin event handler
+$document.on( "click", selector, function( event ) {
```
GormFrank, line 43:
```suggestion
$document.on( "click", selector + " input[type=checkbox]", function( event ) {
```

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,65 @@
...
+	var elm = $( event.currentTarget ),
+		$elm = $( elm ),
```
GormFrank, line 45: Since the element is not the parent div anymore, you will need to get the parent in a variable with something like `.parentUntil(selector)` from jQuery.

src/plugins/gc-checklist/gc-checklist.js
```diff
@@ -0,0 +1,65 @@
...
+		data = JSON.parse( $elm.attr( "data-" + componentName ) ),
```
GormFrank, line 46:
```suggestion
		data = wb.getData( $parent, componentName ),
```

src/plugins/gc-checklist/gc-checklist-en.hbs
delisma: > This list item are not similar to its sibling. Please move it as paragraph after the unordered list or remove it.

If it's outside the component it will be tricky to attach the event to the feedback

delisma: And that would make the targeting a bit too explicit

src/_provisional.scss
delisma: Done

GormFrank (changes requested):

src/_provisional.scss
```diff
@@ -329,6 +333,133 @@
...
 
+:not(form) .provisional {
+	&.gc-checklist {
+		ul {
+			padding-left: 0;
+		}
+
+		output {
+			font-size: 20px;
+			font-weight: bold;
```
duboisp, line 344: I suggest to style the `output` like we do for `panel` or `well`.

And the use of the "bold" should be assigned based on the content (HTML) that is going to be inserted inside the output element.

duboisp (changes requested):
For the font-size, I think we can re-use some variable

PR #1751, opened 2020-11-15
[WET-137] GC-Subway Index Page Added

namjohn920:
GC-Subway index page added.

PR #1757, opened 2020-12-04
GC Subway - Remove sub-items from the list in working example

GormFrank:
Also:
- got rid of the `.noline` class as requested by DTO in WET-147.
- added proper active menu item styles and semantic.

**Will probably conflict with #1751**

### DIFF
--- a/src/plugins/gc-subway/page2-en.hbs
+++ b/src/plugins/gc-subway/page2-en.hbs
@@ -1,8 +1,8 @@
 ---
 {
-	"title": "[Page 1] - GC Subway map menu (provisional)",
+	"title": "[Page 2] - GC Subway map menu (provisional)",
 	"language": "en",
-	"altLangPrefix": "gc-subway",
+	"altLangPrefix": "page2",
 	"pageclass": "cnt-wdth-lmtd",
 	"noContentTitle": true,
 	"breadcrumb": [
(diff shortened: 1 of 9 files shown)


duboisp (approved):

duboisp:
I tested @GormFrank changes and it works as expected. 

I am going to push additional changes + working example to show how to create mobile documentation navigation anchor.

GormFrank:
Noticed that after adding the two `<a>` tags in the `<li>`s, the white circles have vanished in mobile view.

Also, the `.cnt-wdth-lmtd` class causes the bottom border to be cut short from the right end as you move from the desktop to the mobile view.

GormFrank:
Noticed that after adding the two `<a>` tags in the `<li>`s, the white circles have vanished in mobile view.

Also, the `.cnt-wdth-lmtd` class causes the bottom border to be cut short from the right end as you move from the desktop to the mobile view.

duboisp:
@GormFrank I fixed both issue.

GormFrank:
> @GormFrank I fixed both issue.

I'll look at it tomorrow first thing in the morning!

src/plugins/gc-subway/page2-en.hbs
```diff
@@ -0,0 +1,58 @@
...
+	],
+	"secondlevel": false,
+	"dateModified": "2020-12-07",
+	"share": "true"
+}
+---
+{{>alert-provisional-demos}}
+
+<nav class="provisional gc-subway">
+	<h1 id="gc-document-nav">Service initiation template</h1>
```
delisma, line 19: Can we replace **Service initiation template** by **Subway navigation component**?

src/plugins/gc-subway/page2-fr.hbs
```diff
@@ -0,0 +1,58 @@
...
+	],
+	"secondlevel": false,
+	"dateModified": "2020-12-07",
+	"share": "true"
+}
+---
+{{>alert-provisional-demos}}
+
+<nav class="provisional gc-subway">
+	<h1 id="gc-document-nav">Gabarit d'initiation de service</h1>
```
delisma, line 19: Can we replace **Gabarit d'initiation de service** by **Composante de navigation de style métro**?

delisma (commented):
We wouldn't want to convey that this is a template. The example should stick to the component idea as much as possible.  We may publish a template once it's stable.

PR #1762, opened 2020-12-14
Mobile center - Prevent Close link to appear on table nav next

duboisp:
Related to WET-115

Prevent the "X close" link to appear beside each individual block when a user navigate with the table multi-page navigation.

To reproduce: 
1. Go in the mobile center page,
2. Click on "Next" underneath the table (in the table navigation)
3. The "X Close" link appear beside each block.

Expected
* There is no "X Close" that appear on step 3

### DIFF
--- a/src/mobile-centre/_base.scss
+++ b/src/mobile-centre/_base.scss
@@ -86,7 +86,6 @@
 				&:hover {
 					background-color: darken(#f5f5f5, 5%);
 					border-color: darken(#e3e3e3, 5%);
-					cursor: pointer;
 				}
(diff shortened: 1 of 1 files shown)


GormFrank (commented):
I would really want to get rid of the cursor: pointer at line 89 too but I am not sure if this is out of scope. Other than that, bug is fixed!

GormFrank (approved):

duboisp:
This change is required because of departmental priority to deploying this minor release on Canada.ca by the end of this week.

PR #1758, opened 2020-12-07
Theme/topic - apply new template design

GormFrank:
### Do not merge until PR #1765 is merged

---

Related to WET-126

As prototyped by TBS: https://design.canada.ca/coded-layout/theme_topic_guidance.html

Closes #1720

PR #1765, opened 2020-12-21
Adding SASS variables to ease component development

GormFrank:
... and increase reusability.

As demonstrated in #1758 also.

### DIFF
--- a/src/_defaults.scss
+++ b/src/_defaults.scss
@@ -106,7 +106,7 @@ h1 {
 
 	// Overwride sizing
 	h3, h4, h5, h6 {
-		font-size: 20px; // same as h5
+		font-size: $h5-size;
 		margin-bottom: 5px;
 		margin-top: 23px; // same as h5
 	}
@@ -117,7 +117,7 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 {
 	@extend %defaults-alternative-font-family;
 	font-variant-ligatures: no-common-ligatures;
-	font-weight: 600;
+	font-weight: $bold-weight;
 }
 
 .h3,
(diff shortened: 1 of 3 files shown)


src/_defaults.scss
```diff
@@ -106,7 +106,8 @@ h1 {
 
 	// Overwride sizing
 	h3, h4, h5, h6 {
-		font-size: 20px; // same as h5
+		font-size: $h5-size;
+		font-weight: $bold-weight;
```
duboisp, line 110: Do we need to set the font-weight here ? It's already set at [line 120](https://github.com/wet-boew/GCWeb/pull/1765/files#diff-8e0bb7658c6bab5067e061ee1d8d7d4fbb26fb750ecacee5f8fa133a6dcfc003L120)

duboisp (changes requested):
I tested the font-size override for the gc-srvinfo and the font-size remain 20px as expected.

src/_defaults.scss
GormFrank: The reason why I put it there is because of the h6 only, which is re-overwritten at [line 591](https://github.com/wet-boew/GCWeb/pull/1765/files#diff-8e0bb7658c6bab5067e061ee1d8d7d4fbb26fb750ecacee5f8fa133a6dcfc003R591).

I assumed that in the first place you were expecting h6 to be bold too in the service and info pattern so that was my rationale. If the goal is not to have **h6** in bold in the service and info then I can remove it, but if it should be bold then i suggest we leave it there or I put it in a separate rule block just for the **h6** only like so: `h6 { font-weight: $bold-weight; }`.

Also, do you want me to apply the $font-weight variable on line 120?

duboisp: > Also, do you want me to apply the $font-weight variable on line 120?

That could be a good idea

>  because of the h6 only,

I don't think there is a use case when this pattern is used with h6. Probably that selector was too aggressive for the need. I suggest to leave it as is for now until we have a situation that require an h6 in this pattern.

duboisp (approved):

duboisp:
This change is required to let us progress on a departmental priority, the beta theme/topic page.

PR #1766, opened 2021-01-18
Fieldflow - Inline style and use first item as default

duboisp:
Related to WET-162

### DIFF
--- a/src/plugins/fieldflow/fieldflow-en.hbs
+++ b/src/plugins/fieldflow/fieldflow-en.hbs
@@ -35,14 +35,13 @@
 		<li><a href="#ajax">Ajax</a></li>
 		<li><a href="#grouping">Grouping</a></li>
 		<li><a href="#nesting">Nesting</a></li>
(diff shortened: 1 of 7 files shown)


src/plugins/fieldflow/fieldflow-en.hbs
```diff
@@ -224,3 +223,29 @@
 		&lt;/li&gt;
 	&lt;/ul&gt;
 &lt;/div&gt;</code></pre>
+
+<h2 id="inline">Inline button, select an option to go to...</h2>
+<p>Note: when inline, it's context must be clear enough in regard of how to use the selection of options. It's designed to be use where there is one heading representing the action, one description and one field flow control.</p>
```
GormFrank, line 228:
```suggestion
<p>Note: when inline, its context must be clear enough in terms of how to use the selection of options. It's designed to be use where there is one heading representing the action, one description and one field flow control.</p>
```

GormFrank (changes requested):

GormFrank (approved):

PR #1767, opened 2021-01-18
WET-140 Spacing Problem Fixed

namjohn920:
As discussed from https://github.com/wet-boew/wet-boew/pull/9018#pullrequestreview-568498021,

The last "p" tag has to have the same bottom spacing as the top spacing of the first "p".
Since the first "p" has neither margin/padding-top or margin/padding-bottom, I just got rid of the last child selector.


duboisp:
Cancelled, we are not able to properly reproduce the issue

PR #1768, opened 2021-01-18
gc-feature - Design pattern add - Cohabit

duboisp:
This design pattern update is to support the beta theme/topic template, as proposed by #1758

### DIFF
--- a/site/pages/component/gc-features-en.hbs
+++ b/site/pages/component/gc-features-en.hbs
@@ -7,7 +7,7 @@
 		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/themes-dist/GCWeb/index-en.html" }
 	],
 	"secondlevel": false,
-	"dateModified": "2020-02-18",
+	"dateModified": "2021-01-18",
 	"share": "true"
 }
 ---
@@ -21,7 +21,7 @@
 	<dd>Canada.ca design pattern</dd>
 
 	<dt>Last review</dt>
-	<dd>2020-02-18</dd>
+	<dd>2021-01-18</dd>
 
 	<dt>Conforming to</dt>
 	<dd>Content and IA spec 2.1</dd>
@@ -35,7 +35,10 @@
 			<li><code>gc-features</code></li>
 			<li><code>gc-initiatives</code> (extends)</li>
 		</ul>
-		</dd>
+	</dd>
+
+	<dt>Component version</dt>
+	<dd>Version 3.1</dd>
 </dl>
 
 <h2>Working example</h2>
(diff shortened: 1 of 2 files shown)


site/pages/component/gc-features-en.hbs
```diff
@@ -268,6 +269,67 @@
...
+			&lt;!-- gc-feature here --&gt;
+		&lt;/section&gt;
+	&lt;/div&gt;
+	&lt;div class="col-md-4"&gt;
+		&lt;!-- Cohabit component here. For example the Follow us component --&gt;
+	&lt;/div&gt;
+&lt;/div&gt;</code></pre>
+
+<h6>One feature tile - Cohabit design pattern</h6>
+<pre><code>&lt;div class="row row-eqht mrgn-tp-xl"&gt;
```
GormFrank, line 286: We should try to use the pure CSS equal height instead of the JS plugin at this point.

GormFrank (changes requested):

site/pages/component/gc-features-fr.hbs
```diff
@@ -270,6 +270,67 @@
...
+			&lt;!-- gc-feature here --&gt;
+		&lt;/section&gt;
+	&lt;/div&gt;
+	&lt;div class="col-md-4"&gt;
+		&lt;!-- Cohabit component here. For example the Follow us component --&gt;
+	&lt;/div&gt;
+&lt;/div&gt;</code></pre>
+
+<h6>One feature tile - Cohabit design pattern</h6>
+<pre><code>&lt;div class="row wb-eqht-grd mrgn-tp-xl"&gt;
```
GormFrank, line 287: On the one feature tile pattern, I don't think that we should be using equal height at all since when the cohabit section has a significant height, it makes the feature tile link's clickable area huge because of the `stretched-link` class.
```suggestion
<pre><code>&lt;div class="row mrgn-tp-xl"&gt;
```

GormFrank (changes requested):

site/pages/component/gc-features-fr.hbs
duboisp: It is not for the gc-feature it-self, but for the grid container that have gc-feature and another feature like gc-followus

duboisp: Ok, after our discussion I see, I am removing it.

GormFrank (approved):
Approved assuming Travis-ci builds. @duboisp you can merge it if you need to for the release.

duboisp:
Considering that we need to release GCWeb prior the technical review and this PR is ready, we are including it in the same release.

PR #1770, opened 2021-01-19
Added p-0 class to implimented pattern

Christopher-O:
To align header and pattern to rest of content on left side for main page pattern.
@delisma @GormFrank


duboisp (approved):

duboisp:
Considering that we need to release GCWeb prior the technical review and this PR is ready, we are including it in the same release.

PR #1772, opened 2021-01-22
Topic template - Test case 1 - Alternative template when right column is higher

duboisp:
This alternative template avoid to have the "contributor" section to float in the middle of the page when there is not a lot of doormat link in the service and information section.

The fix is to remove the floating CSS class (pull-right, pull-left) and to move the "Most requested" section after the "Service and information" section. 

PP reference number: WET-155

@delisma can you ask if DTO are ok with that change for this special edge case?

### DIFF
--- a/site/pages/index-en.hbs
+++ b/site/pages/index-en.hbs
@@ -130,6 +130,15 @@
 			<td>v1.4</td>
 		</tr>
 		<tr>
+			<td><a href="topic-testcase-1-en.html">Topic page, test case 1</a></td>
+			<td><time>2021-01-22</time></td>
+			<td>Templates</td>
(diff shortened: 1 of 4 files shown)


delisma:
> This alternative template avoid to have the "contributor" section to float in the middle of the page when there is not a lot of doormat link in the service and information section.
> 
> The fix is to remove the floating CSS class (pull-right, pull-left) and to move the "Most requested" section after the "Service and information" section.
> 
> PP reference number: WET-155
> 
> @delisma can you ask if DTO are ok with that change for this special edge case?

I'm gonna check the [guidance](https://design.canada.ca/coded-layout/theme_topic_guidance.html) about it and get back to you.

GormFrank (approved):

PR #1774, opened 2021-01-27
Banner: Affect style on img instead of a + remove hyperlink from FIP on home page

GormFrank:
Related WET-161

### DIFF
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,11 @@
 {{#unless provisional}}
 	<div class="brand col-xs-9 col-sm-5 col-md-4" property="publisher" resource="#wb-publisher" typeof="GovernmentOrganization">
+		{{#contains page.pageclass "home"}}
+		<img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /><span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span>
+		<link href="https://www.canada.ca/{{language}}.html" property="url" />
+		{{else}}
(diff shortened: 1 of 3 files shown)


site/includes/brand.hbs
```diff
@@ -1,6 +1,11 @@
 {{#unless provisional}}
 	<div class="brand col-xs-9 col-sm-5 col-md-4" property="publisher" resource="#wb-publisher" typeof="GovernmentOrganization">
+		{{#contains page.pageclass "home"}}
+		<img src="{{assets}}/../{{site.theme}}/assets/sig-blk-{{language}}.svg" alt="{{{i18n "tmpl-gc-sig"}}}" property="logo" /><span class="wb-inv"> / <span lang="{{#is page.language "en"}}fr">{{{i18n "tmpl-gc-sig" language="fr"}}}{{else}}en">{{{i18n "tmpl-gc-sig" language="en"}}}{{/is}}</span></span>
```
EricDunsworth, line 4: FWIW this could be simplified and made less redundant by always using an ``<a>`` element and only using handlebars conditions to omit ``href="https://www.canada.ca/{{language}}.html" property="url"`` from the home page.

An ``<a>`` element with no ``href`` attribute represents a placeholder link in HTML (perfect fit for this kind of scenario - it's a placeholder for where a real link would've been if the user wasn't already on the page the link points to).

duboisp: Yeah, but in this scenario we are also adding a custom link to keep the URL for the publisher organisation

src/banner/_screen-sm-max.scss
```diff
@@ -3,7 +3,7 @@
  */
 header {
 	.brand {
-		a {
```
duboisp, line 6: I think we would keep the margin style applied on the anchor.

src/banner/_screen-xs-max.scss
```diff
@@ -3,7 +3,7 @@
  */
 header {
 	.brand {
-		a {
```
duboisp, line 6: I think we would keep the margin style applied on the anchor.

duboisp (dismissed):

duboisp (approved):

site/includes/brand.hbs
duboisp: We would prefer a pattern that would produce less ambiguity among accessibility expert, even if managing it would require us more effort because of complexing that handlebar includes.

PR #1779, opened 2021-02-09
Theme: Improve h1-h6 font size declarations

EricDunsworth:
* Replaces GCWeb's heading font size variables (``$h#-size``) with overrides for Bootstrap's ones (``$font-size-h#``)
* Removes redundant SCSS selectors (Bootstrap already has them)
* Corrects estimated pixel size comment for the H1 heading's top margin in small view and under (from ~15px to ~26px)

### DIFF
--- a/components/baseline/_base.scss
+++ b/components/baseline/_base.scss
@@ -78,7 +78,7 @@ h1.gc-thickline {
 * Panel title
 */
 .panel-title {
-	font-size: $h3-size;
+	font-size: $font-size-h3;
 }
 
 /*
--- a/components/gc-follow-us/_base.scss
+++ b/components/gc-follow-us/_base.scss
@@ -81,7 +81,7 @@ $provisional-social-media-icons-size: 38px;
 
 	&.gc-followus {
 		h2 {
-			font-size: $h4-size;
+			font-size: $font-size-h4;
 		}
 
 		ul {
--- a/misc/views/_screen-sm-max.scss
+++ b/misc/views/_screen-sm-max.scss
@@ -8,7 +8,7 @@
 h1,
 .h1 {
 	font-size: 1.7em; // expected 34px;
-	margin-top: .75em; // expected ~15px;
+	margin-top: .75em; // expected ~26px;
 }
 
 h2,
--- a/sites/_defaults.scss
+++ b/sites/_defaults.scss
@@ -106,7 +106,7 @@ h1 {
 
 	// Overwride sizing
 	h3, h4, h5, h6 {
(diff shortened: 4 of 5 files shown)


src/views/_screen-sm-max.scss
```diff
@@ -7,13 +7,13 @@
 // Heading font-sizes
 h1,
 .h1 {
-	font-size: 1.7em; // expected 34px;
-	margin-top: .75em; // expected ~15px;
+	font-size: 34px;
+	margin-top: 26px;
```
EricDunsworth, line 11: The expectations comment for the old ``margin-top`` property was way off... in practice Gecko and Blink were computing ``.75em`` to ``25.5px``. So I rounded it up to ``26px``.

duboisp:
Need to be tested and we need to ensure that the font-height exception for the header styling are kept. For example the gc-srvinfo

duboisp:
In what use case you need to use pixel?

EricDunsworth:
Just force-pushed an update to revert the pixel units back to ems.

@duboisp We've already spoken about my use case for using pixels, but just to put my stance in writing:

* Both Bootstrap 3.4.x and WET 4.0 use pixel units for their heading font sizes. Since GCWeb is a WET theme, I don't see any reason as to why it shouldn't also use pixels (aside from backwards compatibility with any code that intentionally misused the heading classes for font scaling).
* Font sizes that are based on pixel units are predictable. They'll always look the same size no matter what. By contrast, em-based font classes are open to being misused. It's possible to nest them together to scale the font size in ways that differ from what GCWeb intended. Take this super giant H5 for instance - that wouldn't be possible with pixel units:
  * ```<div class="h2"><div class="h2"><div class="h2"><div class="h2"><div class="h2"><div class="h2"><h5>H5</h5></div></div></div></div></div></div>```
* My down-to-earth use case was that I ran into a situation where I needed to place a heading within a ``<summary>`` element. That's not as simple as just throwing in a ``<h#>`` element. In order to make the summary's disclosure widget (triangle arrow) adopt the same font size as the heading, it's necessary to add a ``class="h#"`` attribute to the summary. But as a result of GCWeb's use of em units, ``<h#>`` grows way out of proportion since its font scales twice.

GormFrank:
Looks good, needs to be tested.

GormFrank (changes requested):
Variables change look as expected.

Need to:

- [x] Rebase code

EricDunsworth:
Rebased :)!

duboisp (approved):

PR #1780, opened 2021-02-10
On social media (.gc-followus): Reduce line height

EricDunsworth:
Previously, when a long social media channel name was placed into one of this component's list items, it'd split across multiple lines. But its line height would create an excessive amount of space between each line.

This addresses it by reducing the line height and by using another means of making single-line text look vertically-centered compared to the social media icons.

**Screenshots...**

**Before:**
gc-followus-before

**After v1 (outdated - YouTube link's lines are noticeably closer together):**
gc-followus-after

**After v2 (YouTube link's lines are even closer together):**
gc-followus-after-v2

CC @delisma @luc-bertrand-hrsdc @GormFrank


delisma:
@EricDunsworth I don't get it do we notice the difference or we don't?
> But the difference is barely noticeable
> lines are noticeably closer together

GormFrank:
> Previously, when a long social media channel name was placed into one of this component's list items, it'd split across 2 lines. But its custom line height would create an excessive amount of space between both lines.
> 
> This addresses it by removing the line height override. As a side effect, all links have now shifted a few pixels upwards in relation to their icons. But the difference is barely noticeable and arguably looks better than before.
> 
> **Screenshots...**
> 
> **Before:**
> gc-followus-before
> 
> **After (YouTube link's lines are noticeably closer together):**
> gc-followus-after
> 
> CC @delisma @luc-bertrand-hrsdc @GormFrank

@EricDunsworth My only concern when simply removing this line height property is that the text becomes slightly uncentered vertically (1-2px too high). You know how picky I am when it comes to those 1-2px off things lol.

GormFrank:
@EricDunsworth For me personally it is noticeably worse than it was since it is not vertically centered by doing this change. But you know how picky I am when it comes to those 1-2px off things lol.

I haven't seen any real Theme-Topic page in action yet but I don't think social links will fall on multiple lines very often, so the upward shift bothers me more (again personally).

EricDunsworth:
@delisma

> @EricDunsworth I don't get it do we notice the difference or we don't?

I feel like the parts you quoted omitted some nearby context lol, but just to clarify:

>> But the difference is barely noticeable

IMO the vertical positioning difference between the single-line links and their respective icons is barely noticeable between this PR and how things were like before.

>> lines are noticeably closer together

IMO multi-line social media links are noticeably closer together with this PR. I wouldn't mind reducing the line height even more but didn't want go too far for what amounts to very minor visual effects. But I didn't expect anyone to care so umm I guess I'll need to go back to the drawing board.



@GormFrank

> @EricDunsworth For me personally it is noticeably worse than it was since it is not vertically centered by doing this change. But you know how picky I am when it comes to those 1-2px off things lol.

Oh didn't realize you were that picky about those kinds of things lol. I'm usually super picky about them too but in my testing I found the difference to only be noticeable when doing a direct head-to-head comparison. Nothing felt off to me when looking at this PR's version of the new theme/topic page template on its own.

> I haven't seen any real Theme-Topic page in action yet but I don't think social links will fall on multiple lines very often, so the upward shift bothers me more (again personally).

I'm working on a page that'll use the new theme/topic page (albeit with a bit of custom CSS). Its Facebook link is slated to be called "Employment and Social Development Canada"... which takes up 2 lines. When my clients saw how it looked, they requested that I reduce the spacing between that link's lines. I agreed and felt like it was a flaw in the template - which lead to this PR.

@delisma @GormFrank 
So the vibe I'm getting is that you'd prefer that I revise this PR to:
* Use an even smaller height (lower than the ``<main>`` element's 1.65em) to further reduce the spacing of multi-line links so they can stick together a bit more
* Add some extra properties to keep the spacing/positioning of single-line items and their icons exactly the same as before

Thoughts?

duboisp:
If we are talking line-heigh, we should look at assigning a value that would meet the [WCAG 2.2 1.4.12 Text Spacing](https://www.w3.org/TR/WCAG22/#text-spacing) Just saying. 
May be it will worth to have that discussion at our next WET code sprint? What do you think?

EricDunsworth:
Just force-pushed an update and edited the OP :). The revised commit further reduces the line height and retains the exact same pixel measurements as before for single-line items.

@duboisp I was initially planning to set the line height to 1.3 in my update, but ended up going with 1.54 after seeing your comment.

Btw it looks like SC 1.4.12 just cares about ensuring no loss of content or functionality occurs if a user overrides the line height to at least 1.5. So I would've still been able to pass it with ``line-height`` set to1.3, but felt like the difference between 1.3 and 1.54 wasn't big enough to make it worthwhile. Plus 1.54 matches what I'm planning to use for #1684's ``$gcweb-line-height-large`` variable.

duboisp (approved):

PR #1786, opened 2021-02-26
Details/Summary: Improve support for heading classes

EricDunsworth:
Fixes layout issues when styling summaries as headings and/or placing headings within them.

Specific changes:
* Add a second set of heading font size variables that represent pixel units (uses a similar naming scheme as #1779)
* Don't disable borders on summaries that use the "h3, "h4", "h5" and "h6" classes (they were disabling the summary's border on collapsed details elements)
* Rename the $details-identation variable to $details-indentation (typo fix)

Note:
* Complements wet-boew/wet-boew#9045


EricDunsworth:
@GormFrank This will fail to build until wet-boew/wet-boew#9045 has landed and GCWeb is correctly referencing it (since both WET and GCWeb are currently using the typo-ed ``$details-identation`` variable).

Btw I'd also recommend merging #1779 prior to this. This PR **might** need to be rebased afterwards... Neither PR changes the exact same lines, but they change blocks of variables that are situated near one-another. So if git tries to use one of the lines #1779 changed to infer the position of this PR's new variable declarations, it might get treated like a conflict.

src/_defaults.scss
```diff
@@ -120,13 +120,40 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
...
-.h3,
-.h4,
-.h5,
-.h6 {
+.h3:not(summary),
+.h4:not(summary),
+.h5:not(summary),
+.h6:not(summary) {
 	border: 0;
 }
```
EricDunsworth, line 128: IMO these selectors are really strange :S. They might even be pointless...

I only kept them in place out of an abundance of caution since I'm trying to sneak in this PR at the last minute.

Those selectors only target a subset of heading classes and leave their equivalent heading elements alone. That causes deviations between the heading classes and elements. Plus neither Bootstrap 3.4.x, WET 4.x nor GCWeb contain any selectors that attempt to setup borders on H2-H6 headings.

FWIW these selectors were first introduced in #33 and #868. The former seems to indicate it was related to IA spec changes (possibly something to do with GCWeb's old mega menu or mobile menu). No idea if they still serve any useful purpose these days.

duboisp:
ci build fail:

> Running "sass:all" (sass) task
> Fatal error: Undefined variable: "$details-indentation".

EricDunsworth:
@duboisp wet-boew/wet-boew#9045 and this PR are like two sides of the same coin. You can't successfully build one without the other.

To make this PR build successfully:
1. Merge wet-boew/wet-boew#9045
2. Release a new version of WET
3. Update GCWeb's dependencies to use the new version of WET
4. Trigger a rebuild of this PR

The WET PR corrected a typo in the ``$details-identation`` variable's name (it's now called ``$details-indentation``). But GCWeb references that variable. So I made the same typo fix in this PR too. If I undo it, GCWeb's CI build will start working... but will break again once the WET PR lands. So my hands are tied.

EricDunsworth:
I rebased this yesterday and it's currently fully in line with wet-boew/wet-boew#9045... but that PR is still in draft and is probably going to require significant changes that'll affect this one. So I'll mark this as a draft for now.

PR #1802, opened 2021-04-13
Multimedia player: Increase CC container's height

EricDunsworth:
Prevents closed captions that contain 2 lines of text from causing the CC container's height to shift during playback (due to GCWeb's taller line height).

The shifting can be distracting when the captions update (i.e. switching between no captions, 1 line captions and 2 line captions).

Depends on wet-boew/wet-boew#9082.

**Screenshots...**

**Before** (CC container height: hardcoded 3em + 1em for padding):
gcweb-multimedia-cc-container-before

**After** (CC container height: calculated 3.3em + 1em for padding):
gcweb-multimedia-cc-container-after

### DIFF
--- a/components/baseline/wet-boew-overrides.scss
+++ b/components/baseline/wet-boew-overrides.scss
@@ -9,11 +9,24 @@
 .wb-mltmd {
 	margin-bottom: 10px;
 	margin-top: 10px;
+
+	&.cc_on {
+		.wb-mm-cc {
+			height: calc(#{1.65 * $mm-cc-lines}em + #{$mm-cc-padding * 2});
+		}
+	}
 }
 
 .wb-mltmd.video:not(.playing):not(.waiting)
 .display::after {
 	line-height: 1.5em;
+}
+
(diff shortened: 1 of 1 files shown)


EricDunsworth:
Keeping this as a draft for the time being since I haven't tried compiling my SCSS yet and since the media player demos are currently broken due to wet-boew/wet-boew#9067.

Btw this PR wouldn't of been necessary if WET used SCSS variables/calculations to determine the CC container's height and if GCWeb re-used said variables.

I might send a PR to improve how WET goes about it in the future, but I don't think GCWeb could inherit it "for free" unless it directly overrides Bootstrap's ``$line-height-base`` variable (unlikely to happen anytime soon - if ever).

duboisp (approved):
pre-approved, We will just need to wait for the next release wet-boew to ensure CI pass

duboisp:
@EricDunsworth please rebase

EricDunsworth:
Just rebased it :)!

duboisp (approved):

PR #1806, opened 2021-04-24
Templates move - Legislation

duboisp:


templates/index.json-ld
```diff
@@ -110,28 +110,6 @@
 				"path": "theme-fr.html"
 			},
 
-			{
-				"title": "[Regulation name] <small>([NN-XX])</small>",
```
GormFrank, line 114: @duboisp Shouldn't it be in the master list also?

duboisp: No, only in the template. This main json-ld is only a temporary measure to take the time of creating every single one.

GormFrank (approved):

PR #1808, opened 2021-04-28
Add wb-suggest feature

duboisp:
related to WET-175

This plugin allow to load dynamically datalist suggestion from a JSON file. It is currently used by the Canada.ca search page.


components/wb-suggest/suggest.js
```diff
@@ -0,0 +1,279 @@
...
+	};
+
+// Bind the init event of the plugin
+$document.on( "timerpoke.wb " + initEvent + " " + jsonFetched, selector, function( event ) {
+
+	var eventType = event.type,
+		eventTarget = event.target,
+		eventCurrentTarget = event.currentTarget;
+
+	if ( eventCurrentTarget === eventTarget ) {
```
GormFrank, line 261: You could have used `event.currentTarget` directly and not create the `eventCurrentTarget` variable since it is only used once.

components/wb-suggest/suggest.js
```diff
@@ -0,0 +1,279 @@
...
+	addDataListOptions = function( filter, limit, attrSuggestions ) {
+		var suggestions = attrSuggestions || JSON.parse( this.dataset.wbSuggestions || [] ),
+			filterType = this.dataset.wbFilterType || "any",
+			filterRegExp,
+			lenSuggestions = suggestions.length,
+			idx, suggestion, suggestionNorm,
+			clone, option,
+			optionAddedUnAccent = [],
+			currentOptions = this.childNodes,
+			j, j_cache, j_len = currentOptions.length - 1,
```
GormFrank, line 64: I know it would probably be in extreme cases, but could the currentOptions length be 0 and give a result of -1, potentially causing an infinite loop at line 106? If it is a concern, maybe just add a simple check for length greater than 0 or else leave as is.

GormFrank (changes requested):

GormFrank (changes requested):
Also noticed that the text does not scale as you zoom in. 

Also, do we have a fallback or recommendations for Basic HTML?

components/wb-suggest/suggest.js
duboisp: This didn't happen during testing

components/wb-suggest/suggest.js
duboisp: Yes

duboisp: Fixed

GormFrank (approved):

PR #1809, opened 2021-04-28
Field Flow: Increased font size alternative

GormFrank:
Increases the font size of Field flow elements following visual update in 2019, by adding the `.gc-font-2019`.

(Relates to WET-180)

### DIFF
--- a/sites/_defaults.scss
+++ b/sites/_defaults.scss
@@ -130,6 +130,7 @@ h1, .h1, h2, .h2, h3, .h3, h4, .h4, h5, .h5, h6, .h6
 
 /* Exclude temporary the form element from the font update, keep same style as before
  * - Rational: More testing and various adjustment need to be completed before to apply it
+ * - Fieldflow is brought to speed with the visual update with a clearly defined temporary class .gc-font-2019, but the class can be tried out on other forms
  */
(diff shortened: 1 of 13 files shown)


components/wb-fieldflow/_base.scss
```diff
@@ -5,6 +5,29 @@
...
+				margin-top: 11px;
+			}
+
+			[type="submit"], select {
+				padding: .54em 1em;
+			}
+		}
+	}
+}
+
```
duboisp, line 30: Move this code in the GCWeb and here add a comment about it with a link to site/_defaults.scss

duboisp (changes requested):

sites/_defaults.scss
```diff
@@ -141,6 +142,26 @@ form .btn:not(.btn-call-to-action), .form-control {
 legend {
 	line-height: 1.65em;
 }
+.gc-font-2019 {
+	&, & ~ .wb-fieldflow-form, & + div {
```
duboisp, line 146: This selector should be ideally component specific. Currently when I apply `gc-font-2019` on the first fieldflow component, it's all components in the page that are transformed. It is acting like a global CSS that we would apply to the body element.

duboisp (changes requested):

duboisp (approved):
Look good, I reviewed and tested locally with no issue and I was getting the expected behaviour

PR #1817, opened 2021-05-12
GC Checkboxes and radio - Plugin stabilization

GormFrank:
Related to WET-182

PR to apply this "GC Checkboxes and radio" pattern to Field flow coming soon...


components/gc-chckbxrdio/index.json-ld
```diff
@@ -14,13 +14,22 @@
...
-	"modified": "2018-07-12",
+	"modified": "2021-05-07",
 	"componentName": "gc-chckbxrdio",
-	"status": "provisional",
-	"todos": [
-		"Move CSS from generic provisional to locally here"
-	],
+	"status": "stable",
 	"pages": {
+		"examples": [
```
duboisp, line 21: This shoudl be "docs" and the other "examples"

duboisp (changes requested):
TODO: test it locally

delisma:
> Related to WET-182

Can we be a bit more detailed about WET-182?

GormFrank:
@Delisma WET-182 is the ticket number of our internal request, which about stabilizing the Checkboxes and radio buttons pattern and then applying that pattern to Field flow. I didn't think there was any description needed for this PR since the title describes it enough. PR for implementation with Field flow will come soon. There you go I'll add that.

duboisp (commented):
Approved after local test and wait for the implementation of the GC Checkboxes in field flow

duboisp:
@GormFrank please resolve conflict

components/gc-chckbxrdio/gc-chckbxrdio-doc-en.html
```diff
@@ -0,0 +1,176 @@
...
+
+<dl class="horizontal">
+	<dt>Status</dt>
+	<dd>Stable</dd>
+	<dt>Type</dt>
+	<dd>Canada.ca design pattern</dd>
+	<dt>Last review</dt>
+	<dd>2021-05-07</dd>
+	<dt>Conforming to</dt>
+	<dd>Content and IA spec 2.1</dd>
```
duboisp, line 22: This is not defined by the content and IA spec, but just by the "Template and design pattern library" which is not versioned currently.

components/gc-chckbxrdio/index.json-ld
```diff
@@ -14,14 +14,23 @@
...
+				"language": "en",
+				"path": "gc-chckbxrdio-doc-en.html"
+			},
+			{
+				"title": "Cases à cocher et boutons radio",
+				"language": "fr",
+				"path": "gc-chckbxrdio-doc-fr.html"
+			}
+		],
+		"examples": [
```
duboisp, line 33: Can you add the reference to the design system, like how it is added in the legislation pages

components/provisional-fr.html
```diff
@@ -125,11 +125,11 @@ <h3><code>.experimental</code> depricated</h3>
 			<td>
 				<p>Adoptez les nouvelles grandes cibles tactiles d'entrée pour les cases à cocher et les boutons radio</p>
 				<ul>
-					<li><a href="component/gc-chckbxrdio-fr.html">Cases à cocher et bouttons radio - Configuration de conception de Canada.ca</a></li>
+					<li><a href="components/gc-chckbxrdio-fr.html">Cases à cocher et bouttons radio - Configuration de conception de Canada.ca</a></li>
 				</ul>
 			</td>
 			<td>v8.0.1</td>
-			<td></td>
+			<td>Stable</td>
```
duboisp, line 132: Add the date or the next expected version number of GCWeb (Minor)

components/gc-chckbxrdio/gc-chckbxrdio-doc-en.html
```diff
@@ -0,0 +1,174 @@
+---
+title: Checkboxes and radio buttons
+language: en
+tag: chckbxrdio
+category: Pattern
+parentdir: components
```
duboisp, line 6: This is not needed

duboisp (changes requested):
There is a small glitch with the inline checkbox and inline radio button

1px checkbox
1px checkbox-checked
1px radio

I will check what I can do to fix that issue.

There is also unnecessary `class="disable"` code. I going to include that change too.

duboisp (changes requested):
We would need to replace the class `chckbxrdio-grp` because it is associated to no style. However, I think the intend was to use this style : `chkbxrdio-grp` (without "c" as the third character)

duboisp (changes requested):
The HTML markup in the working example are not conform to WCAG. More particularly the information and relationship because some content don't belong to their parent heading and some section are identified with multiple top level heading.

Note: I am going to submit a change to fix that.

duboisp (approved):
Approved,
I applied the required changes and I fixes the checkbox/radio display issue when it was displayed inline
I successfully tested locally and did a partial WCAG assessment only on how the checkboxes/radio are implemented and checked the HTML pages structure.

GormFrank:
> The HTML markup in the working example are not conform to WCAG. More particularly the information and relationship because some content don't belong to their parent heading and some section are identified with multiple top level heading.
> 
> Note: I am going to submit a change to fix that.

@duboisp Can you be more specific in this review? I did not change the semantic of the page from the provisional one so I don't have a reference on what you are pointing out now that you have pushed the changes. Just want to see how obvious it was. Thanks!

GormFrank:
> We would need to replace the class `chckbxrdio-grp` because it is associated to no style. However, I think the intend was to use this style : `chkbxrdio-grp` (without "c" as the third character)

@duboisp Yes, the intent is to use the one without the "c", which I thought I did a Find and replace everywhere. The typo dates back to when it was in provisional state.

GormFrank:
> There is a small glitch with the inline checkbox and inline radio button

@duboisp I am not sure I understand what the glitch is/was from those screenshots. I'm testing your change now.

components/gc-chckbxrdio/_base.scss
```diff
@@ -15,6 +15,11 @@
 	box-shadow: 0 0 0 2px #000, 0 0 8px 4px rgb(59, 153, 252);
 }
 
+fieldset.gc-chckbxrdio {
+	border-top: 0;
+	padding-top: 0;
+}
+
```
GormFrank, line 22: @duboisp You decided to include the `chkbxrio-grp` styles so that we don't need to use the helper? Good idea.

GormFrank (commented):
> There is a small glitch with the inline checkbox and inline radio button

What browser were you using? I never encountered this glitch. Your quick change fixes it though 🥇

GormFrank:
@duboisp I'm just asking myself if it is necessary to repeat "with explicit label" on every single `<h2>` on the page. Everything looks good now though.

duboisp (approved):

duboisp:
Just find out that the provisional code for the checkbox was not removed from the provisional CSS, like here: https://github.com/wet-boew/GCWeb/blob/master/components/_provisional.scss#L158

PR #1827, opened 2021-05-27
Méli-mélo - 2021-5-conjunction

Christopher-O:
Added index page with all class types and multiple examples. Also included evergreen implementation plan and considerations.

### DIFF
--- a/méli-mélo/2021-05-conjunction/and-or.css
+++ b/méli-mélo/2021-05-conjunction/and-or.css
@@ -1,0 +1,242 @@
+[class*=cnjnctn-type-] {
+	display: flex;
+	flex-wrap: nowrap;
+	flex-direction: column;
+	margin-bottom: 15px;
+	margin-left: 5px;
+	margin-right: 0px;
+	margin-top: 15px;
+	position: relative;
+	min-height: 60px;
+}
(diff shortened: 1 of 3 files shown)


duboisp (changes requested):
Everything look good, but we would need you to add working example when the feature is used with the following CSS class:

* cnjnctn-col-90
* cnjnctn-col-80
* cnjnctn-col-75
* cnjnctn-col-70
* cnjnctn-col-60
* cnjnctn-col-40
* cnjnctn-col-30
* cnjnctn-col-25
* cnjnctn-col-20

Christopher-O:
@duboisp  the additional examples and code snippets have been added.

duboisp (changes requested):
We would need a small French working example page to illustrate at minimum the use of the CSS selector `html:lang(fr)`.

Christopher-O:
@duboisp  should we keep the CSS selector at the html element (html:lang(fr)), or should we decouple it (*:lang(fr))?  I don't really see a need for someone to put an entire and/or statement in another language on a page...but then again it would make it more flexible.

duboisp (approved):

duboisp:
I tested and reviewed this contribution to ensure all CSS is demoed and I confirm it don't conflict with wet-boew neither GCWeb.

duboisp (approved):

PR #1830, opened 2021-06-01
Canada Day 2021 red thematic

nfrenette:
Ready for peer review and approval

Here a working example: https://nfrenette.github.io/GCWeb/m%C3%A9li-m%C3%A9lo/th-canadaday/canada-day.html

### DIFF
--- a/méli-mélo/th-canadaday/style.css
+++ b/méli-mélo/th-canadaday/style.css
@@ -1,45 +1,30 @@
-a .bg-canadaday {
-text-decoration:underline;
-text-decoration-color:#fff;
+/* Thematic red */
+.bg-canadaday {
+    background-color: #990000;
 }
 
-a div.bg-canadaday h2:focus {
-outline:3px solid #fff;
+.panel-canadaday,
+hr.brdr-canadaday {
+    border-color:#990000;
 }
 
-a p {
-text-decoration:none;
-}
-
-/* Main red */
-.bg-canadaday,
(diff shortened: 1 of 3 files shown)


méli-mélo/th-canadaday/style.css
```diff
@@ -0,0 +1,35 @@
...
+button.bg-canadaday:focus {
+    background-color: #6b0303;
+}
+
+/* Active state */
+a.bg-canadaday:active,
+a.bg-canadaday:active:focus,
+a.bg-canadaday.active:focus,
+a.bg-canadaday.active.focus,
+a.bg-canadaday:active.focus,
```
duboisp, line 25: Do you have a demo where it use the CSS class `active` and `focus` on anchor along with the `bg-canadaday` CSS class?

As you mentioned, you wanted to avoid the primary blue, my local test showed me it is when the anchor state is at the same time `:active`  and `:hover`. 
I propose you the following update instead

```suggestion
a.bg-canadaday:active:hover,
```

méli-mélo/th-canadaday/style.css
```diff
@@ -0,0 +1,35 @@
...
+button.bg-canadaday:active {
+    background-color: #000000 !important;
```
duboisp, line 27: I don't think you need to use `!important` here. In what situation it is needed?

duboisp (changes requested):

duboisp (approved):

PR #1831, opened 2021-06-01
GC Subway: Fix index page display issue on desktop

GormFrank:
Also fixed sub-pages working examples layouts.

### DIFF
--- a/components/gc-subway/gc-subway.js
+++ b/components/gc-subway/gc-subway.js
@@ -22,7 +22,6 @@ var $document = wb.doc,
 	mainClass = "gc-subway-section",
 	toggleClass = "wb-inv",
 	desktopInited = false,
-	skipLink = false,
 	$html = wb.html,
 	$h1, $h2, $h1Copy, $menu, $main,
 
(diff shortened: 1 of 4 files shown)


components/gc-subway/gc-subway.js
```diff
@@ -44,14 +44,17 @@ var $document = wb.doc,
 			$h1 = $( "h1", $elm );
 			h1 = $h1.get( 0 );
 
-			// Ensure the element have an ID
-			if ( !h1.id ) {
-				h1.id = wb.getId();
-			}
+			if( h1 ) {
```
duboisp, line 47: Improve this condition, the 2 sub-instruction are only "if" conditional. For example `if ( h1 && !skipLink )`. But wait, is the variable "skipLink" is required? Is the init are called twice?

duboisp (changes requested):

components/gc-subway/gc-subway.js
GormFrank: @duboisp Yes, init gets called multiple times and this was causing multiple links to be added to the Skip links.

components/gc-subway/gc-subway.js
```diff
@@ -44,14 +43,13 @@ var $document = wb.doc,
 			$h1 = $( "h1", $elm );
 			h1 = $h1.get( 0 );
 
-			// Ensure the element have an ID
-			if ( !h1.id ) {
-				h1.id = wb.getId();
-			}
+			// Add Subway H1 to skip links only once and if it is a sub-page
+			if ( h1 && event.currentTarget === event.target ) {
```
duboisp, line 47: Move up at line 41 the currentTarget check

duboisp (changes requested):
After change completed and upon testing. It's pre-approved.

duboisp (approved):

duboisp:
I tested and work as expected. I did a second code review and it looks good.

PR #1838, opened 2021-06-04
GC tables: Left-align column labels

EricDunsworth:
This component normally right-aligns data cells and floats their column labels to the left. That's fine for column labels that can fit on a single line. But ones that span across multiple lines end up becoming right-aligned - which looks really strange.

This resolves it by explicitly aligning column label text to the left.

**Screenshots...**

**Before:**
gc-tables-label-alignment-before

**After:**
gc-tables-label-alignment-after

### DIFF
--- a/components/gc-table/_screen-sm-max.scss
+++ b/components/gc-table/_screen-sm-max.scss
@@ -45,6 +45,7 @@
 				content: attr(data-label);
 				float: left;
 				font-weight: bold;
+				text-align: left;
 			}
 			&:last-child {
 				border-bottom: 0;


duboisp (changes requested):
I am ok with this changes, but we would need a working example.

@delisma @jmealing what do you think?

EricDunsworth:
@duboisp Technically all the pre-existing examples on the [GC tables demo page](https://wet-boew.github.io/GCWeb/components/gc-table/gc-table-en.html) would demonstrate this change if it were to be merged in. Especially the ``small`` example (since the second column's name is long).

jmealing:
Thanks Eric - this looks great, thanks for catching that issue. I agree the working example has this example already which should update once merged.

PR #1841, opened 2021-06-04
GC tables: Support the "table-bordered" class

EricDunsworth:
A double-border effect used to appear in small view and under when pairing GC tables alongside Bootstrap's "table-bordered" class.

This resolves it by disabling that class' extra borders when GC tables' styles are in effect. Also adds an example of that class in use.

**Screenshots...**

**Before:**
gc-tables-table-bordered-before

**After:**
gc-tables-table-bordered-after

### DIFF
--- a/components/gc-table/gc-table-en.html
+++ b/components/gc-table/gc-table-en.html
@@ -5,7 +5,7 @@
 		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/GCWeb/index-en.html" },
 		{ "title": "Provisional functionality", "link": "https://wet-boew.github.io/themes-dist/GCWeb/provisional-en.html" }
 	],
-	"dateModified": "2020-10-16",
+	"dateModified": "2021-06-16",
 	"description": "Simple responsive gc tables",
 	"language": "en",
 	"title": "GC tables"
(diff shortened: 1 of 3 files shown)


duboisp (changes requested):
@EricDunsworth We would need a working example showing it.

@delisma @jmealing is the intend of `gc-table` is to work along with the `table-bordered` CSS class

duboisp:
DTO is looking at this, if it is not accept yet, we will add a reference to this PR into the meta file (+ add in a détails page) of the gc-table provisional feature.

EricDunsworth:
@duboisp Just added an example of a table that uses the ``table-bordered`` class into GC tables' demo page.

components/gc-table/gc-table-en.html
```diff
@@ -5,7 +5,7 @@
 		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/GCWeb/index-en.html" },
 		{ "title": "Provisional functionality", "link": "https://wet-boew.github.io/themes-dist/GCWeb/provisional-en.html" }
 	],
-	"dateModified": "2020-10-16",
+	"dateModified": "2021-06-09",
```
EricDunsworth, line 8: Should the last review date on line 22 also be updated? Not sure what exactly it entails compared to date modified.

Technically any incoming PRs already need to be reviewed before merging. But if both dates are supposed to be one in the same, there wouldn't be any point in hardcoding last review's date if it could just refer to ``dateModified``.

duboisp: Yes, the date at line 22 should be updated too. You can use the jekyll variable
```
{{ page.dateModified }}
```
at line 22 to use the same date. You are free to apply it now or doing it in a other PR.

duboisp (approved):
Tested locally and work as expected

duboisp (approved):

PR #1842, opened 2021-06-04
GC tables: Add partial DataTables support

EricDunsworth:
A bottom table border used to appear in small view and under when pairing GC tables alongside WET's tables (DataTables) plugin.

This removes that border when GC tables' styles are in effect. Also adds an DataTables example.

**Note:**
I also wanted to make GC tables disable datatables' optional scrolling features, but was unable to since its scrolling container is situated outside the table.

**Screenshots...**

**Before:**
gc-tables-datatables-before

**After:**
gc-tables-datatables-after

**PS:**
This will need to be rebased after #1841 gets merged.

### DIFF
--- a/components/gc-table/_screen-sm-max.scss
+++ b/components/gc-table/_screen-sm-max.scss
@@ -27,6 +27,11 @@
 				}
 			}
 		}
+		&.dataTable {
+			&.no-footer {
+				border-bottom: 0;
+			}
+		}
 		.text-left {
 			clear: both;
 			display: block;
--- a/components/gc-table/gc-table-en.html
+++ b/components/gc-table/gc-table-en.html
@@ -217,3 +217,36 @@
 		</tr>
 	</tbody>
 </table>
+
+<h3>Responsive cards with ``wb-tables`` class for filterable tables</h3>
+<p><span class="label label-info">Optional</span></p>
+
+<p>The class <code>wb-tables</code> can be added to the <code>&lt;table&gt;</code> to make it use <a href="https://wet-boew.github.io/v4.0-ci/demos/tables/tables-en.html">WET's tables plugin</a>. Please note that scrollable and responsive WET tables are unsupported.</p>
+
+<table class="provisional gc-table table wb-tables" id="myTable5">
+	<caption>Immunization clinics</caption>
+	<thead>
+		<tr>
+			<th>Name</th>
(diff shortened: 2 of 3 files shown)


duboisp (changes requested):
@EricDunsworth We would need a working example. What is your context of use of the dataTable plugin along with the gc-table. If it is only for the filtering feature, you can use the filter plugin which don't impact the table layout.  

@delisma @jmealing any concern?

EricDunsworth:
The goal of this PR is to allow an implementer to use the best of both worlds so to speak:
* Datatables-specific features (like column sorting) in tables that are presented using tabular layouts in medium view and over.
* Linearized tables in small view and under (i.e. making them more mobile-friendly than they'd otherwise be).
* Taking advantage of datatables-specific features that can lend themselves to all views (e.g. filtering, pagination, etc...).

Datatables already has optional [configuration options for responsive tables](https://datatables.net/extensions/responsive/) (see  the example on [datatables' landing page](https://datatables.net/)). But that works very differently from GC tables:
* Responsive datatables work by hiding columns that can't fit on-screen. Visible data cells turn into interactive controls that can show/hide the hidden column data.
* GC tables works by hiding table headers (``<thead>``), linearizing all table rows and presenting them like "mini tables". It also uses pure CSS.

So given all that... some implementers might prefer the way GC tables presents table data in smaller views over responsive datatables. So I don't see any harm in giving them the option to use it. Especially since it barely needs any additional CSS.

Btw this PR won't make all datatables implementations perfectly compatible with GC tables. But it should cover most use cases. The only scenarios I can think of off hand where it wouldn't work very elegantly would be if the [table is scrollable](https://datatables.net/examples/basic_init/scroll_y.html) or if the [table is already responsive](https://datatables.net/extensions/responsive/).

@duboisp I hadn't considered using the filter plugin alongside a normal GC tables implementation. But IMO datatables still has enough going for it that it'd make sense to support.

jmealing:
All good - I haven't noticed that dark line border appearing in my testing. Thank you for removing it Eric

duboisp:
TODO:
* [ ] Create a working example with the data table plugin

duboisp (approved):

duboisp:
pre-approved upon conflict resolved

duboisp (approved):

PR #1843, opened 2021-06-04
GC tables: Remove last row's bottom margin

EricDunsworth:
This component's table rows use bottom margins to space-them out when they're presented like mini-tables. But that margin doesn't serve any purpose on the last row. Especially since the table itself already has its own 23px bottom margin that doesn't collapse with the row margins.

**Screenshots...**

Red is the last row's bottom margin, whereas yellow is the table's bottom margin.

**Before:**
gc-tables-tr-last-of-type-bottom-margin-before

**After:**
gc-tables-tr-last-of-type-bottom-margin-after

### DIFF
--- a/components/gc-table/_screen-sm-max.scss
+++ b/components/gc-table/_screen-sm-max.scss
@@ -24,6 +24,17 @@
 			margin-bottom: .625em;
 			padding: .35em;
 		}
+		> {
+			:last-child {
+				> {
+					tr {
+						&:last-of-type {
+							margin-bottom: 0;
+						}
+					}
+				}
+			}
+		}
 		caption {
 			font-size: 1.1em;
 		}


duboisp (approved):
I might be interesting to check how this affect table containing multiple tbodies. But I think that is not in scope yet of gc-tables.

I am ok with your proposed change.

@jmealing @delisma any concern?

EricDunsworth:
@duboisp Good catch regarding multiple ``<tbody>`` elements :). Didn't cross my mind on Friday.

I tried it out earlier and it removed the bottom margin from the end of every ``<tbody>``... which wasn't desirable :(. So I've force-pushed an update that seems capable of reliably targeting the final ``tr`` in any table (based on my testing).

**PS:**
IMO GC tables doesn't seem to have been designed with multiple ``<tbody>`` elements in mind. Bootstrap's ``table`` class causes a 2px border to appear at the top of adjacent``<tbody>`` elements. GC tables doesn't override it (which looks kind of strange). I think it'd look better if a space appeared below the border. Maybe I'll send out a PR for that later.

jmealing:
Looks great to me, this proposed change works - thank you Eric

PR #1844, opened 2021-06-06
Sites - Add layout for displaying méli-mélo and thématique détails

duboisp:


GormFrank (approved):
Let's go with it. We will reassess for further changes after another review later on.

méli-mélo/th-canadaday/meta.md
```diff
@@ -10,10 +12,15 @@ cssClass:
...
+  Tested by Nick Frenette, nick.frenette at canada.ca. 2021-06-01.
 peNote:
 - The <code>bg-canadaday</code> class must be accompagnied with another dark contrast background colour such as <code>bg-dark</code>
 - The <code>panel-canadaday</code> class must be accompagnied with a fall back color such as <code>panel-default</code>
-lead: PCH - Nick Frenette (@nfrenette)
+pages:
+  examples:
+    - title: Canada day 2021
+      language: en
+      path: index.html
```
GormFrank, line 23: This should be `canada-day.html` instead, because `index.html` doesn't exist for this one.

thématique/gc-thématique-en.html
```diff
@@ -93,17 +94,35 @@ <h2>List of GC promotional feature projects</h2>
...
+a11yStatement: >
+  Accessibility statement, describe what was tested, include any notes/instructions for implementers to follow in order to maintain the accessibility conformance, during and after the thematic expiration.
+  Tested by (Name of the person Nick Frenette, nick.frenette at canada.ca. 2021-06-01.
+peNote:
+- Progressive enhancement notes and instructions
+pages:
+  examples:
+    - title: Title of the page
+      language: en
+      path: index.html
```
GormFrank, line 124: Same for the example here, even though it has 0 impact: `canada-day.html`.

méli-mélo/méli-mélo-en.html
```diff
@@ -18,14 +18,12 @@ <h3>Méli-mélo sub-projects</h3>
...
--->
-
 <ul>
 {% for feature in melimelo_pages %}
-	{% assign first_example = feature.pages.examples | first %}
-	<li><strong>{{ feature.componentName }}</strong>: <a href="{{ feature.componentName }}/{{ first_example.path }}" lang="{{ first_example.language }}">{{ first_example.title }}</a><!-- {{ feature.url }} --></li>
+	{% assign featBasePath = feature.path | slice: 0, 10 %}
+	{% if featBasePath == "méli-mélo/" %}
+	<li><a href="{{ feature.componentName }}/détails-en.html">{{ feature.componentName }}</a></li>
+	{% endif %}
```
GormFrank, line 26: Do you think you could also fix the link on this page at line 86, by removing the following part from it: `/gcweb-compiled-demos`.

GormFrank (changes requested):
Tested locally and found a few hiccups.

méli-mélo/méli-mélo-en.html
duboisp: Yes its probably feasible to do a similar fix, but gcweb-compiled-demos is a different github reposirty

thématique/gc-thématique-en.html
duboisp: This is just an example of what a configuration file might look like. We would need to add more documentation later on and do some change at our current build script

méli-mélo/th-canadaday/meta.md
duboisp: fixed

méli-mélo/méli-mélo-en.html
duboisp: At line 86, it refer to a different github reposority where compilated demos are published. Those compiled méli-mélo demos can only obtained when the build script is run. However, we can probably apply a fix when we are running it into a jekyll local instance

thématique/gc-thématique-en.html
GormFrank: I know, that's why I said it has 0 impact. However, to be consistent with the actual meta file I strongly suggested to change it here too.

duboisp: I removed the Canada day reference

GormFrank (approved):

PR #1846, opened 2021-06-08
GC tables: Support adjacent tbodies

EricDunsworth:
Bootstrap's "tables" class adds a 2px border between adjacent ``tbody`` elements.

But it looked out of place in GC tables. That component adds a bottom margin to every row... which caused the adjacent ``tbody``'s top border to look like it was "sticking" to its first row.

This resolves it by adding a top margin to the adjacent ``tbody``'s first row to prevent the sticking. That allows the top border to look like a real separator. Also adds a second ``tbody`` to the first GC tables example.

Spinoff of #1843.

**Screenshots...**

**Before:**
gc-tables-tbody-adjacent-before

**After:**
gc-tables-tbody-adjacent-after

CC @jmealing @delisma @duboisp

### DIFF
--- a/components/gc-table/_screen-sm-max.scss
+++ b/components/gc-table/_screen-sm-max.scss
@@ -2,6 +2,8 @@
  PROVISIONAL features: Responsive tables cards
 */
 
+$gc-tables-tr-margin-vertical: .625em;
+
 .provisional {
 	&.gc-table {
 		&.table-bordered {
@@ -49,7 +51,7 @@
 		tr {
 			border: 1px solid #ddd;
 			display: block;
-			margin-bottom: .625em;
+			margin-bottom: $gc-tables-tr-margin-vertical;
 			padding: .35em;
 		}
 		> {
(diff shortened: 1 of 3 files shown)


duboisp:
We will need a working example.

@delisma @jmealing what do you think of this style?

duboisp (changes requested):
Need a working example

jmealing:
@duboisp What do we need a working example of? Wouldn't this correction be the default because the examples use the class "table"? 

The correction to fix the overlapped border looks good to me, thank you @EricDunsworth

EricDunsworth:
I've just added a second ``tbody`` into the first example to demonstrate adjacent ones. IMO it didn't warrant its own distinct example given that there's nothing very special about using multiple tbody elements (it's standard HTML markup).

duboisp (approved):

duboisp:
pre-approved upon conflict resolved

duboisp (approved):

PR #1860, opened 2021-06-25
Home page template - hide RAP button

Ricokola:
Related to: wet-190

Add option in order to hide the RAP button from the Home page template.

### DIFF
--- a/sites/main-footer/inc-footer.html
+++ b/sites/main-footer/inc-footer.html
@@ -1,16 +1,18 @@
 <div class="pagedetails{% if include.addContainer %} container{% endif %}">
 	<div class="row">
 		<div class="col-sm-6 col-md-5 col-lg-4">
-			<details class="brdr-0">
-				<summary class="btn btn-default text-center">{{ i18nText-feedback }}</summary>
-				<div class="well row">
-				{%- if i18nText-lang == "fr" -%}
-					{% include main-footer/inc-report-problem-fr.html %}
-				{%- elsif i18nText-lang == "en" -%}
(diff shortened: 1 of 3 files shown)


sites/main-footer/inc-footer.html
```diff
@@ -1,16 +1,18 @@
...
-			<details class="brdr-0">
-				<summary class="btn btn-default text-center">{{ i18nText-feedback }}</summary>
-				<div class="well row">
-				{%- if i18nText-lang == "fr" -%}
-					{% include main-footer/inc-report-problem-fr.html %}
-				{%- elsif i18nText-lang == "en" -%}
-					{% include main-footer/inc-report-problem-en.html %}
-				{%- endif -%}
-				</div>
+			{%- if page.noReportProblem == empty -%}
```
duboisp, line 4: Check if the `unless` would be better

duboisp (changes requested):

templates/home/home-en.html
```diff
@@ -1,12 +1,11 @@
...
-	"share": "true",
-	"title": "Home - Canada.ca"
-}
+title: Home - Canada.ca
+language: en
+altLangPage: home-fr.html
+dateModified: 2021-06-29
+layout: layout-home
+noReportProblem: true
+share: true
```
duboisp, line 8: Just a comment: Should not the "Share this page" be there by default? @Ricokola how it is currently setup in MWS?

duboisp (approved):

duboisp:
gcweb-jekyll MINOR

Issue #1861, opened 2021-06-29
Bug - includes/metadata.html is missing 5 mandatory meta tags


delisma:
## Describe the bug
According to Canada.ca Standard on Metadata; the properties Title, Creator, Language, Issued, Modified and Subject are mandatory. Additional properties are optional. At the moment these metadata properties are missing from the template.

## To Reproduce
Steps to reproduce the behaviour:
1. Build a website with the remote theme GCWeb-Jekyll
2. Create and publish a page on `gh-page`
3. Inspect the `head` of the page
4. See metadata missing

## Current behaviour
The current metadata.html file only have the `description` meta tag

## Additional context
https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=18909

PR #1862, opened 2021-06-29
Adding missing mandatory meta tags in includes/metadata.html

delisma:
---
name: Metadata
about: This pull request add mandatory meta tags in metadata.html
title: 'Adding missing mandatory meta tags in includes/metadata.html'
labels: 'enhancement'
assignees: 'delisma'
---

## Que fait cette demande d'extraction?

Ajoute les métadonnées obligatoires:
* Title
* Creator
* Language
* Issued
* Modified
* Subject

## Type de contribution
- [ ] Apatride ou composante de base
- [ ] Composant interactif
- [ ] Composant utilitaires
- [x] Composant du site
- [ ] Section d'une mise en page
- [ ] Mise en page

## Liste de contrôle générale
- [x] [Documentation technique](FONCTIONNALITE.md) créé / mis à jour
- [x] Conforme aux [guides de style] (https://www.canada.ca/en/gouvernment/apropos/design-system.html)
- [] Entrée du journal des modifications ajoutée, si nécessaire
- [] Tests ajoutés pour cette fonctionnalité / bug

## Problèmes liés

#1861


sites/includes/canonical.html
```diff
@@ -0,0 +1,10 @@
+{% comment %}
+Should only be called if page.altLangPage is set and a BasePath URL exist
+{% endcomment %}
+{%- if site.baseurl -%}
+<link rel="canonical" href="{{ page.url | prepend: site.baseurl }}" />
+<link rel="alternate" href="{{ page.altLangPage | prepend: site.baseurl | prepend: i18nText-altLang }}" hreflang="{{ i18nText-altLang }}" />
+{% elsif site.urlalt %}
+<link rel="canonical" href="{{ page.url | prepend: site.urlalt[ page.lang ] }}" />
+<link rel="alternate" href="{{ page.altLangPage | prepend: site.urlalt[ i18nText-altLang ] }}" hreflang="{{ i18nText-altLang }}" />
+{%- endif -%}
```
duboisp, line 10: Is this file valuable enough to not be in the "metadata.html" include file? Is it reusable outside the file "metadata.html" context of use?

Both file are currently pretty small, I am not sure currently it is valuable enough to be into 2 separate file. I am suggesting to only have it in one single file for now.

sites/includes/canonical.html
```diff
@@ -0,0 +1,10 @@
+{% comment %}
+Should only be called if page.altLangPage is set and a BasePath URL exist
+{% endcomment %}
+{%- if site.baseurl -%}
+<link rel="canonical" href="{{ page.url | prepend: site.baseurl }}" />
+<link rel="alternate" href="{{ page.altLangPage | prepend: site.baseurl | prepend: i18nText-altLang }}" hreflang="{{ i18nText-altLang }}" />
+{% elsif site.urlalt %}
+<link rel="canonical" href="{{ page.url | prepend: site.urlalt[ page.lang ] }}" />
+<link rel="alternate" href="{{ page.altLangPage | prepend: site.urlalt[ i18nText-altLang ] }}" hreflang="{{ i18nText-altLang }}" />
```
duboisp, line 9: `canonical` should represent the final URL of a given page, not the base url or the site neither the alternative URL of the site with `alternate` + `hreflang`.

Those 2 meta are page specific, not site specific.

Note: A permanent URL, if exist, can be seen as the canonical URL.

templates/splash/layout-splashpage.html
```diff
@@ -6,26 +6,26 @@
 		<title>{{ page.title }}</title>
 		<link rel="schema.dcterms" href="http://purl.org/dc/terms/" />
 		<meta content="width=device-width, initial-scale=1" name="viewport" />
-		<meta name="description" content="{{ page.description-en }}" />
-		<meta name="dcterms.description" content="{{ page.description-en }}" />
-		<meta name="dcterms.creator" content="{{ page.creator-en }}" />
+		<meta name="description" content="{{ page.description[ page.lang ] }}" />
+		<meta name="dcterms.description" content="{{ page.description[ page.lang ] }}" />
+		<meta name="dcterms.creator" content="{{ page.creator[ page.lang ] }}" />
```
duboisp, line 11: Use the variable `i18nText-lang` instead of `page.lang` because it support multiple format to define the current language of a given page.

duboisp (changes requested):

sites/layouts/core.html
```diff
@@ -13,6 +13,9 @@
 {% include license.html %}
 <title>{{ page-title }} - {{ i18nText-siteTitle }}</title>
 <meta content="width=device-width, initial-scale=1" name="viewport">
+{%- if page.altLangPage -%}
+{% include canonical.html %}
+{%- endif -%}
```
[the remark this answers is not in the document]
delisma: That makes sense.

sites/includes/canonical.html
delisma: I didn't quite know how to treat the canonical stuff in a Jekyll template. It assumes that sites would always be bilingual. If people are using the tools to mostly set up experiments, should they have the option to turn the canonical stuff off or customize it? If that's the case shouldn't it be separate from the mandatory metadata requirements?

sites/includes/canonical.html
delisma: That's why I wanted to concatenate the base url with the page file name, but the logic still need works. I wanted to take into account that the base path could be the same url for both language or different for both. But that requires some custom properties in the `_config.yml` file and sounds like overhead to me.

templates/splash/layout-splashpage.html
delisma: You would probably need to explain a bit more why we need `i18nText-lang` when `page.lang` is the more usual Jekyll way to do it and kind of leverage the API.

sites/includes/canonical.html
duboisp: Sorry, I missed the "prepend" instruction for site.baseurl.

Don't assume this language toggle: prepend: i18nText-altLang

I don't think the site.urlalt is needed for now. Let build gcweb-jekyll for experimentation for now and let deal with that when we know that some production site use it.

delisma:
> You would probably need to explain a bit more why we need `i18nText-lang` when `page.lang` is the more usual Jekyll way to do it and kind of leverage the API. 

Still not sure why you want publishers/developers to rely on a value that is hard-coded in the template instead of relying on value that they control on page by page basis.

sites/includes/metadata.html
```diff
@@ -1 +1,9 @@
...
\ No newline at end of file
+<meta name="description" content="{% if page.description %}{{ page.description }}{% else %}{{ site.description[page.lang] }} {% endif %}" />
+<meta name="author" content="{% if page.creator %}{{ page.creator }}{% else %}{{ site.creator[page.lang] }} {% endif %}"/>
+<meta name="dcterms.title" content="{{ page.title }}"/>
+<meta name="dcterms.description" content="{% if page.description %}{{ page.description }}{% else %}{{ site.description[page.lang] }} {% endif %}"/>
+<meta name="dcterms.creator" content="{% if page.creator %}{{ page.creator }}{% else %}{{ site.creator[page.lang] }} {% endif %}"/>
+<meta name="dcterms.language" title="ISO639-2/T" content="{{ i18nText-language | downcase | truncate: 3, '' }}"/>
+<meta name="dcterms.subject" title="gccore" content="{{ page.subject }}"/>
+<meta name="dcterms.issued" title="W3CDTF" content="{{ page.dateIssued }}"/>
+<meta name="dcterms.modified" title="W3CDTF" content="{% if page.dateModified %}{{ page.dateModified }}{% else %}{{ page.dateIssued }}{% endif %}"/>
```
duboisp, line 9: Check if we can leverage the jekyll page variable: `page.date`

ref. https://jekyllrb.com/docs/variables/

delisma: > Check if we can leverage the Jekyll page variable: `page.date`

After a few test the `page.date` has to be explicitly declare on a page for Jekyll to use this value

> To Check: I do think we named the front-matter modified and issued.

Right now it seems that `dateIssued` and `dateModified` are used in the splash page. Should make a separate PR to rename these variables.

I think for now this should be ready to merge.

duboisp: That is fine. Thanks @delisma for checking how those page properties was working.

sites/includes/canonical.html
duboisp: As discussed, the canonical link will be in another PR

sites/includes/canonical.html
duboisp: As discussed, canonical link would be in a different PR

sites/includes/metadata.html
```diff
@@ -1 +1,9 @@
-<meta name="description" content="{% if page.description[page.lang] %}{{ page.description[page.lang] }}{% else %}{{ site.site-description[page.lang] }} {% endif %}" />
\ No newline at end of file
+<meta name="description" content="{% if page.description %}{{ page.description }}{% else %}{{ site.description[page.lang] }} {% endif %}" />
+<meta name="author" content="{% if page.creator %}{{ page.creator }}{% else %}{{ site.creator[page.lang] }} {% endif %}"/>
+<meta name="dcterms.title" content="{{ page.title }}"/>
+<meta name="dcterms.description" content="{% if page.description %}{{ page.description }}{% else %}{{ site.description[page.lang] }} {% endif %}"/>
+<meta name="dcterms.creator" content="{% if page.creator %}{{ page.creator }}{% else %}{{ site.creator[page.lang] }} {% endif %}"/>
+<meta name="dcterms.language" title="ISO639-2/T" content="{{ i18nText-language | downcase | truncate: 3, '' }}"/>
```
duboisp, line 6: Add a new jekyll property with the 3 letter alternative instead of truncating.

https://github.com/wet-boew/GCWeb/blob/master/sites/includes/i18n-core.liquid#L12

duboisp (changes requested):
Pre-approved upon successful local testing, code review and when the following change are completed:

* [x] Remove the title attribute
* [x] Add a new jekyll variable for 3 letter language code

duboisp:
I did the requested change in my cloned PR #2059

PR #1865, opened 2021-06-30
Fix liquid template issue breaking gcweb-jekyll

duboisp:


Ricokola (approved):
As reviewed and discussed with Pierre

PR #1874, opened 2021-07-22
Add an SVG logo to be used for email Brand Indicator Message Identification (BIMI)

maxboyer:
- Added an SVG icon to the main canada.ca template (https://www.canada.ca/bimi-icon.svg)
- Updated the mobile favicon to a higher quality that matches the official Canada flag maple leaf design

**Intended use**

Email clients implementing BIMI (Gmail added support this month) can fetch a logo to display next an email that passed DMARC validation. GoC departments or agencies can add a DNS entry to their email domains.

**DNS Record example**

default._bimi.canada.ca  IN TXT v=BIMI1; l=https://www.canada.ca/bimi-icon.svg

**Additional context**

"Brand Indicators for Message Identification or BIMI (pronounced: Bih-mee) is an emerging email specification that enables the use of brand-controlled logos within supporting email clients. BIMI leverages the work an organization has put into deploying DMARC protection, by bringing brand logos to the customer’s inbox. For the brand’s logo to be displayed, the email must pass DMARC authentication checks, ensuring that the organization’s domain has not been impersonated."

**Reference**

https://bimigroup.org/
https://en.wikipedia.org/wiki/Brand_Indicators_for_Message_Identification


delisma:
I think a working example and documentation are in order. Otherwise we would need to put this enhancement behind some kind of flag until it's sorted with the Design System product team.

maxboyer:
Thanks for the feedback. I'll setup a working example on our test sub-domain. It will take a few weeks, while we get SSC to apply the DNS changes.

GormFrank:
Hi @maxboyer, just as a recommendation, we could also simply upload your image to AEM. It would be available at https://canada.ca/content/dam/canada/bimi-icon.svg. If you are not an AEM user, then you could open a ticket at Principal Publisher's service desk. If you are interested I will share the link to the Service desk with you.

Otherwise, regarding your Pull request I am just unsure as to wether or not this belongs to GCWeb as it is not technically related to the Front-end Framework per se, agreed @delisma? If we would end up merging it, it would then be available at https://www.canada.ca/etc/designs/canada/wet-boew/assets/bimi-icon.svg and never at the root (https://www.canada.ca/bimi-icon.svg). @maxboyer What do you think?

delisma:
@maxboyer I was talking with the CITOS (responsable of email configurations like DMARC) at TBS. They were wondering, where do you think the image would be displayed exactly?

If you have mockups or screenshots, that could save you from setting up a testing environment or a sub-domain. Depending on the placement of the brand logo we could make it a policy or a guideline by next fiscal year, so that it could be implemented GC-wide. There's a big transition coming for digital products outside of front-facing websites and we could make a push for broader gc-brand harmonization and adoption.

@GormFrank for what I've been reading on the BIMI specification, the image has to be hosted at the root of the domain for authentification. So we need to figure out a way to enable this eventually.

PR #1887, opened 2021-10-04
Ministerial-profile GCWeb template - Added reduced template and update documentation

Ricokola:
Related to: WET-195 and WET-200
Patch - wet-195 Ministerial-profile GCWeb template updates:
- Create documentation based on Semantic versioning
- Ref: https://github.com/cds-snc/design-gc-conception/issues/42#issuecomment-920005362


templates/ministerial/ministerial-reduced-en.html
```diff
@@ -0,0 +1,18 @@
...
+dateModified: 2021-10-12
+share: true
+---
+
+<div class="row">
+	<div class="col-md-3">
+		<p class="mrgn-tp-lg"><img src="img/265x352.png" alt="" class="img-responsive"></p>
+	</div>
+	<div class="col-md-9">
+		<p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Tempore amet ducimus nihil, voluptate quibusdam? Excepturi in aspernatur rem ipsam aperiam voluptates fugit officiis culpa, ratione, et maxime impedit.</p>
```
duboisp, line 15: Please add an initial spacing "class="mrgn-tp-lg",

Ideally, that spacing would be only added once for the whole row, but considering the other template have it twice and this one is a temporary measure, please add it to the first paragraph with a note in the documentation.

PR #1892, opened 2021-10-22
Adjust the structured data for breadcrumb

polmih:
Previous breadcrumb used typeof=”Breadcrumb”, which doesn’t offer a data structure as rich as the typeof=”BreadcrumbList”.
This PR will replace it with the typeof=”BreadcrumbList” approach, which includes things such as a typeof ListItem on each list items.

Here is a valid example: https://search.google.com/test/rich-results/result?id=zGU8HNwVFWXt4fBPF2cl8w

As per WET-148

### DIFF
--- a/sites/breadcrumbs/breadcrumbs-fr.html
+++ b/sites/breadcrumbs/breadcrumbs-fr.html
@@ -2,11 +2,13 @@
 {
 	"altLangPage": "breadcrumbs-en.html",
 	"breadcrumbs": [
-		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/GCWeb/index.html" }
+		{ "title": "GCWeb accueil", "link": "https://wet-boew.github.io/GCWeb/index-fr.html" },
+		{ "title": "Méli-mélo", "link": "https://wet-boew.github.io/GCWeb/méli-mélo/méli-mélo-fr.html" }
+
 	],
-	"dateModified": "2021-03-30",
+	"dateModified": "2021-12-02",
 	"description": "Documentation sur l'utilisation du fil d'arianne.",
-	"lang": "fr",
+	"language": "fr",
 	"title": "Fil d'arianne"
 }
 ---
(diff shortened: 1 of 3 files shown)


duboisp (changes requested):
Can you add a working example in "site/breadcrumbs"/?

polmih:
working example added. Thanks

GormFrank (approved):

sites/breadcrumbs/breadcrumbs-fr.html
```diff
@@ -17,3 +17,51 @@
...
+			<li property="itemListElement" typeof="ListItem">
+				<a property="item" typeof="WebPage" href="https://wet-boew.github.io/GCWeb/index-fr.html">
+					<span property="name">GCWeb accueil</span>
+				</a>
+				<meta property="position" content="2">
+			</li>
+		</ol>
+	</div>
+</nav>
+<details class="mrgn-tp-lg">
```
duboisp, line 41: Remove the details, but keep the Markup.

But the title that will said, this page breadcrumb should generate the following markup

sites/breadcrumbs/breadcrumbs-fr.html
```diff
@@ -17,3 +17,51 @@
...
+			</li>
+			<li property="itemListElement" typeof="ListItem">
+				<a property="item" typeof="WebPage" href="https://wet-boew.github.io/GCWeb/index-fr.html">
+					<span property="name">GCWeb accueil</span>
+				</a>
+				<meta property="position" content="2">
+			</li>
+		</ol>
+	</div>
+</nav>
```
duboisp, line 40: Can you use the front-matter instead

duboisp (changes requested):

sites/breadcrumbs/breadcrumbs-fr.html
polmih: Done. Thanks

sites/breadcrumbs/breadcrumbs-fr.html
polmih: Done. Thanks

duboisp (approved):

PR #1894, opened 2021-10-26
Features: Fix equal height in template, adjust version number in documentation

GormFrank:
This PR does quite a few things related to Features:

- Fixes the equal height classes in templates
- Clearly indicate .gc-initiatives as deprecated.
- Adjusted version numbers since there was never a version 9.4.2, we jumped straight to 9.5.0 because it was a minor;
- Plus, I adjusted some discrepancies.

Once Equal height CSS is investigated and fixed to work in the features context, we will update the template to accept both equal heights solution with a preference on the pure CSS version.

Also related to MWS-2096

@duboisp

### DIFF
--- a/components/gc-features/gc-features-en.html
+++ b/components/gc-features/gc-features-en.html
@@ -4,7 +4,7 @@ description: "Documentation and working examples for the Context-specific featur
 language: "en"
 altLangPage: "gc-features-fr.html"
 secondlevel: false
-dateModified: "2021-07-21"
+dateModified: "2021-11-29"
 share: "true"
 ---
 <div class="wb-prettify all-pre hide"></div>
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
Change in en vs fr don't match and there is no mention regarding the "gc-initiatives" to be deprecated

components/gc-features/gc-features-en.html
```diff
@@ -273,12 +273,13 @@ <h3>CSS Class (v3.0)</h3>
 				<li>Removed: <code>.mrgn-bttm-md</code></li>
 				<li>Removed: <code>.mrgn-bttm-sm</code></li>
 				<li>Removed: <code>.brdr-rds-0</code></li>
+				<li>Deprecated: <code>.gc-initiatives</code> (had no style nor behaviour attached)</li>
```
duboisp, line 276: I confirmed there is no CSS attached to it.

duboisp (approved):

PR #1897, opened 2021-11-02
Reorg of SASS files to ease implementation of design token and to ease optimization

duboisp:
Reorg of SASS files to ease implementation of design token and to ease optimization

Static working example (as 2021-11-03): https://test-reorgsass.universallabs.org/index-en.html

Todo:
* [x] Test PR's code /// Tester le code de la PR
* [x] Validate changes to ensure all component render the same as before /// Valider que les components sont visuellement pareille avec les changements appliqué
* [x] Rebase this PR to use the latest version of wet-boew (after https://github.com/wet-boew/wet-boew/pull/9195 is merged)


GormFrank (changes requested):
The only glitch I found after looking at all pages (mostly in mobile mode) is a small padding bottom on the FIP. Basically one instruction on the `header .brand` has switched order and therefore lost priority. See images below:

**Before:**
Screen Shot 2021-11-04 at 9 10 16 AM

**After:**
Screen Shot 2021-11-04 at 9 09 57 AM

GormFrank (approved):
Tested all WET-BOEW demos with GCWeb look on top and everything looked good. Plus, I pushed a commit that fixes the banner issue I previously raised. For you to test my commit @duboisp and we'll wait to point to the latest version of WET.

duboisp:
I updated the wet-boew dependencies to ensure the code pass the build.
I tested the header and it work as expected. Thank @GormFrank for testing and making the adjustment for the site header.

PR #1901, opened 2021-11-09
URL mapping: Add action to pre-set value on input

GormFrank:
Execute pre-configured action based on url query string to set value on an input.

### DIFF
--- a/_data/components.json
+++ b/_data/components.json
@@ -902,6 +902,16 @@
 				"title": "Filtrage d'une géomap",
 				"language": "fr",
 				"path": "geomap-fr.html"
+			},
+			{
+				"title": "Set value",
+				"language": "en",
+				"path": "value-en.html"
+			},
+			{
+				"title": "Établir une valeur",
+				"language": "fr",
+				"path": "value-fr.html"
 			}
 		],
 		"docs": [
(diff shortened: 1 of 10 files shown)


duboisp (commented):
Check what we did for the data-json regarding to update form input

duboisp:
I added the support for setting value on attribute and properties. 

* [ ] Documentation and the working example need to be completed.

Ricokola:
Here are the steps to pre-filter results using a value via URL mapping on page load:

1. Set a JSON Manager component has dataset which include a unique ID and a "name" parameter. (Ex: `<div id="jmexample1" data-wb-jsonmanager='{"name": "example1"}'></div>`)
2. Set the URL Mapping component to add value to the dataset
```
"action": "patch",
"source": "#jmexample1",
"match": "ab|bc|mb|nb|nl|nt|ns|nu|on|pe|qc|sk|yt|all",
"base": "/province",
"default": "all",
"patches":
{
	"op": "add",
	"path": "{base}",
	"value": "{qval}"
}
```
3.

GormFrank:
@Ricokola Thank you for this alternative, it would be nice to eventually create  a "Tips & tricks - how to preset value on load" page in the other components section in GCWeb to share with everyone more formally how to accomplish such thing :)

I will now close my PR.

PR #1928, opened 2022-02-08
Font - Improving loading strategy

duboisp:


Ricokola (approved):
From my understanding, I agree with the approach and I don't have any change to suggest.

PR #1944, opened 2022-03-03
Release v10.5.2

Ricokola:
Release v10.5.2


duboisp (changes requested):
Missing the reference to wet-boew in the package lock

duboisp (approved):

PR #1903, opened 2021-11-12
Contextual optional Sign in button and Signed-On / Signed-Off

Ricokola:
Related to: WET-149

New Jekyll page variables:

- **auth**:  This variable must be set to initiate the authentication pattern.
- **type**: This parameter must be set to either "contextual", "signedoff" or "signedon" to determine the authentication pattern that will be initiated.
- **link**: this parameter must be set when using the contextual authentication pattern to set a link to the sign-in button.
- **label**: this parameter must be set when using the contextual authentication pattern since it will be the default label text for the sign-in button on any viewports.
- **labelExtended**: this parameter is optional when using the contextual authentication pattern and must be used along with the label pattern. When set, it will generate the label text for the sign-in button on xs viewports and under.

PR #1947, opened 2022-03-08
Theme: Use 2 columns for first footer list in xs view

EricDunsworth:
GCWeb's footer contains two lists of links:
* The first list used to appear as a 1 column list in extra-small view and under.
* The second list used 1 column in extra-extra-small view and 2 columns in extra-small view.

That setup caused the footer to look unusually long and inconsistent on smartphones in landscape orientation.

This resolves it by adjusting the first list to use the ``colcount-xs-2`` class.

Related changes:
* Add deprecated footer pages to demonstrate the old pattern
* Add short explanations to distinguish the old and new patterns

**Screenshots:**

* **Extra-small view...**
  * **Before:**
gcweb-footer-xs-before
  * **After:**
gcweb-footer-xs-after

**Bonus screenshots (for reference):**
* **Extra-extra-small view (Before + After)...**
gcweb-footer-xxs
* **Small view (Before + After)...**
gcweb-footer-sm

### DIFF
--- a/docs/static-header-footer/site-assets/x-bootstrap-4.css
+++ b/docs/static-header-footer/site-assets/x-bootstrap-4.css
@@ -263,6 +263,12 @@ a.btn, .nav a {
 	margin-top: 20px;
 	width: auto;
 }
+@media screen and (min-width: 480px) {
+	.colcount-xs-2 {
+		-webkit-column-count: 2;
+		column-count: 2;
+	}
+}
 @media screen and (min-width: 768px) {
 	.colcount-sm-2 {
 		-webkit-column-count: 2;
--- a/sites/footers/index.json-ld
+++ b/sites/footers/index.json-ld
@@ -30,6 +30,16 @@
 				"path": "footers-fr.html"
 			},
 			{
+				"title": "Complete footer version 1.0 (deprecated)",
+				"language": "en",
+				"path": "deprecated/footers-v1-en.html"
+			},
+			{
+				"title": "Pied de page complet version 1.0 (dépréciée)",
+				"language": "fr",
+				"path": "deprecated/footers-v1-fr.html"
+			},
+			{
 				"title": "Hide 'About government' from footer",
 				"language": "en",
 				"path": "no-footer-about-en.html"
--- a/_data/sites.json
+++ b/_data/sites.json
@@ -7,14 +7,14 @@
 		"modified": "dct:modified"
 	},
 	"title": {
-		"en": "Site wide demoed feature",
-		"fr": "Démonstration de fonctionalité applicable à l'ensemble du site"
+		"en": "Archived - template",
+		"fr": "Archivé - Gabarit"
 	},
 	"description": {
 		"en": "Gobal demoed feature included but need to move in its own folder.",
 		"fr": "Fonctionalité globale inclus mais qu'il nécessite qu'ils soit déplacé dans son propre dossier."
 	},
-	"modified": "2021-02-12",
+	"modified": "2022-05-03",
 	"componentName": "archived",
 	"status": "stable",
 	"pages": {
@@ -96,6 +96,16 @@
 				"title": "Pied de page complet",
 				"language": "fr",
 				"path": "footers-fr.html"
+			},
+			{
+				"title": "Complete footer version 1.0 (deprecated)",
+				"language": "en",
+				"path": "deprecated/footers-v1-en.html"
+			},
+			{
+				"title": "Pied de page complet version 1.0 (dépréciée)",
+				"language": "fr",
+				"path": "deprecated/footers-v1-fr.html"
 			},
 			{
 				"title": "Hide 'About government' from footer",
--- a/docs/static-header-footer/bootstrap-3.html
+++ b/docs/static-header-footer/bootstrap-3.html
@@ -173,7 +173,7 @@
 		<dl id="wb-dtmd">
 			<dt>Date modified:</dt>
 			<dd>
-				<time>2019-09-12</time>
+				<time>2022-05-18</time>
 			</dd>
 		</dl>
 	</div>
@@ -184,7 +184,7 @@
 	<div class="landscape">
 		<nav class="container wb-navcurr">
 			<h2 class="wb-inv">About government</h2>
-			<ul class="list-unstyled colcount-sm-2 colcount-md-3">
+			<ul class="list-unstyled colcount-xs-2 colcount-md-3">
 				<li><a href="https://www.canada.ca/en/contact.html">Contact us</a></li>
 				<li><a href="https://www.canada.ca/en/government/dept.html">Departments and agencies</a></li>
 				<li><a href="https://www.canada.ca/en/government/publicservice.html">Public service and military</a></li>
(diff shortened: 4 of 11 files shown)


docs/static-header-footer/site-assets/x-bootstrap-4.css
```diff
@@ -263,8 +263,8 @@ a.btn, .nav a {
 	margin-top: 20px;
 	width: auto;
 }
-@media screen and (min-width: 768px) {
-	.colcount-sm-2 {
+@media screen and (min-width: 480px) {
+	.colcount-xs-2 {
```
duboisp, line 267: keep both
* colcount-sm-2
* colcount-xs-2

duboisp:
After discussion with DTO, we will keep support for both pattern in order to keep this change "minor".

That will requires to keep or add an example where it used "col-count-sm-2" and example that use "col-count-xs-2".

EricDunsworth:
Just force-pushed a revised commit with the following updates:
* Updated the Bootstrap 4 CSS files (unmin + minified) to address @duboisp's comment
* Updated some modified dates
* Added these deprecated footer pages:
  * sites/footers/footers-deprecated-en.html
  * sites/footers/footers-deprecated-fr.html
* Updated sites/footers/index.json-ld
* Added a tiny bit of content to "sites/footers/footers-\*.html" and "sites/footers/footers-deprecated-\*.html" to explain how they differ from each other

sites/footers/index.json-ld
```diff
@@ -29,6 +29,16 @@
 				"language": "fr",
 				"path": "footers-fr.html"
 			},
+			{
+				"title": "Complete footer (deprecated)",
+				"language": "en",
+				"path": "footers-en.html"
+			},
+			{
+				"title": "Pied de page complet (d�conseill�)",
```
duboisp, line 38: There is an UTF-8 encoding issue here.

duboisp (changes requested):
The "footers-depreated" files, we will need to figure out how to provide a working example not just the code.
To work around this, you can use the "nocontainer" layout or the "core" layout. 

Or to add a special template feature to replace the whole footer. But a such feature will need to be clearly identified as a test deprecated code feature. I don't want regular users of our GCWeb-jekyll to use a such functionality.

EricDunsworth:
Btw I'll also need to revise my commit message to reflect my last set of updates + upcoming ones.

duboisp:
Let's create a new layout in GCWeb (/_layout) not in gcweb-jekyll to illustrate this deprecated use case.

docs/static-header-footer/site-assets/x-bootstrap-4.css
```diff
@@ -263,6 +263,12 @@ a.btn, .nav a {
 	margin-top: 20px;
 	width: auto;
 }
+@media screen and (min-width: 480px) {
+	.colcount-xs-2 {
+		-webkit-column-count: 2;
```
duboisp, line 268: I do think our post-css will fix this browser CSS

duboisp (changes requested):

docs/static-header-footer/site-assets/x-bootstrap-4.css
EricDunsworth: Do you mean that autoprefixer might do it as a part of the postcss task? I just tried removing that line and ran grunt to no avail (the line wasn't re-inserted).

PostCSS only deals with the dist folder's compiled CSS files. This experimental Bootstrap 4 file isn't situated there. It's a "loose" CSS file in the repo's source files.

sites/footers/index.json-ld
EricDunsworth: Aww lame lol.

Since this isn't the first time ANSI has messed up one of my GCWeb PRs, I decided to check out what was going on...

Turns out that even though GCWeb's .editorconfig file sets the default charset to UTF-8, Notepad++'s editorconfig plugin doesn't currently support that property (see https://github.com/editorconfig/editorconfig-notepad-plus-plus/issues/41#issuecomment-959124854). That leaves it up to Notepad++ to auto-detect file encoding. Since this file didn't previously contain any French characters in it, Notepad++ auto-detected it as ANSI when I opened it for editing. So it saved my accented characters as ANSI.

GitHub's diff viewer probably correctly handles the .editorconfig file's charset property and consequently showed my ANSI characters as messed-up.

EricDunsworth:
Just forced-pushed a revised commit with the following changes:
* Updated commit message
* Updated modified dates
* Fixed encoding issues in ``sites/footers/index.json-ld``
* Added a ``footerDeprecated`` variable:
  * Set its value to "true" in ``sites/footers/footers-deprecated-*.html``
  * Added liquid logic in ``sites/footers/includes/footer.html`` to toggle between ``colcount-xs-2`` and ``colcount-sm-2``
* Fixed broken language toggle links in the deprecated footer page

sites/footers/footers-deprecated-en.html
```diff
@@ -0,0 +1,51 @@
+---
+{
+	"title": "Content page including complete footer (deprecated)",
+	"language": "en",
+	"altLangPage": "footers-deprecated-fr.html",
+	"secondlevel": false,
+	"dateModified": "2022-03-18",
+	"share": "true",
+	"footerDeprecated": "true"
```
EricDunsworth, line 9: @duboisp What's your take on using a ``footerDeprecated`` variable and having the footer include (``sites/footers/includes/footer.html``) check on ``page.footerDeprecated``)? It'd be less repetitive than adding yet another layout file.

If you'd still prefer going with a dedicated layout file I can make further revisions to add one in.

duboisp: Let discuss at our next technical review or our next office hours.

May be we should have something that will allow us to deprecate it again because we know that DTO are currently looking at redesigning the site footers section.

EricDunsworth: Just throwing this out there, but maybe that could be achieved with more specific variable/file names? For example, using a naming scheme along the lines of ``old[Topic][Change]`` or ``old[Topic][YYYYMMDD]`` for variables and file names? If one of those naming schemes were to be applied to this PR, we'd end up with ``oldFooterColcountXs2`` or ``oldFooter20220510``.

EricDunsworth: Ended up scrapping the ``footerDeprecated`` variable and doing something very similar to #1903 and #1962's deprecated page samples.

EricDunsworth:
Just force-pushed some updates with the following changes:
* Scrapped the ``footerDeprecated`` variable
* Renamed/moved the deprecated footer pages
* Revised the content of the deprecated footer pages to be shorter and more to-the-point
* Based the deprecated footer pages on core.html and default.html (didn't add any extra indentation like what #1903 and #1962 did)
* Updated modified dates

duboisp (commented):
Approved upon successful local testing and validation of the current + deprecated code.

duboisp (commented):
Approved upon successful local testing and validation of the current + deprecated code.

duboisp (approved):
I did the local testing and it works as expected.

This is Major change for the footer site component, but minor for GCWeb because we are going to support both HTML structure for the footer site component.

Thanks @EricDunsworth

PR #1950, opened 2022-03-11
Index page improvement - Show working example first and source code

GormFrank:


components/_base.scss
```diff
@@ -173,3 +173,11 @@
...
 	zoom: 1;
 }
+
+/*
+ * WET overrides
+ */
+
+.wb-fltr-inpt {
+	min-height: 39px;
+}
```
duboisp, line 183: Can you move this in another PR?

duboisp (changes requested):

duboisp:
I tested locally and it looks great. 😃

duboisp (approved):

PR #1951, opened 2022-03-14
Remove Share this page button from Canada.ca home page

polmih:
modify the home template to Remove Share this page button.
as per WET-237 request.

### DIFF
--- a/templates/home/home-en.html
+++ b/templates/home/home-en.html
@@ -5,7 +5,6 @@ altLangPage: home-fr.html
 dateModified: 2021-07-07
 layout: home
 noReportProblem: true
-share: true
 ---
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
This is a major change. It will require to add migration instruction from old to new markup.

@polmih is there a way to avoid a Major change? Would be possible to update the share plugin instead? Or to apply an alternate solution for this special use case where there is no feedback and only a share this page link?

- Share plugin:  https://github.com/wet-boew/wet-boew/tree/master/src/plugins/share

duboisp:
This PR will need to be redone according the upcoming change from #1962. 

The change seems to be, after the PR #1962 would be merged, to add a new page property for the home page template.

If that is the expected change, it will be approved and merged upon successful re-review and local testing by myself.

duboisp (approved):

PR #1953, opened 2022-03-16
Filter: Increase input height to match adjacent button

GormFrank:
Padding bottom was useless for inline fieldflow pattern.

This PR can be tested on the GCWeb page with WET-BOEW compiled demos: https://wet-boew.github.io/gcweb-compiled-demos/wetboew-demos/filter/filter-en.html

Plus, it can be tested on the index page and on Fieldflow main working example page (inline pattern). 

Furthermore, as an additional test you can add the following blurb of HTML in the inspector:
```
<div class="input-group">
      <div class="input-group-addon">$</div>
      <input type="text" class="form-control" id="exampleInputAmount" placeholder="Amount">
      <div class="input-group-addon">.00</div>
</div>
```

**In addition to this, I added css include recent fix for Equal height CSS component.**

### DIFF
--- a/components/wb-fieldflow/_base.scss
+++ b/components/wb-fieldflow/_base.scss
@@ -10,8 +10,5 @@
 		.form-control:last-child {
 			border-radius: 4px 0 0 4px;
 		}
-		input[type=submit] {
-			padding-bottom: 7px;
-		}
 	}
 }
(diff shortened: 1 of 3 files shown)


duboisp:
This is a patch.

Code look good.

Todo:
* [ ] Test locally the alignment
* [ ] Test locally the equal height xs fix

duboisp (approved):
I tested both change and it work as expected.

duboisp:
@Ricokola FYI

PR #1960, opened 2022-04-11
Fieldflow: Escape quotes in value attributes

### LINKED ISSUE #1744: Fieldflow - apostrophe trimmed from input field value
Hi!

I want to flag this bug I found when the form input field value has a single quote in it (often used as an apostrophe). After the list is converted to a form by the WET framework, the "value" is trimmed by the single quote inside the text. We tried replacing the single quotes by double quotes inside out... no success there. The workaround for now is to replace the single quote in the text by an apostrophe character.

Thank you for looking into it.

[screenshot](https://imgur.com/a/f2kHyod)

```
<div id="services2" class="wb-frmvld" data-gc-analytics-collect="[{'value':'input','emptyField': 'N/A'}]" data-gc-analytics-formname="whoareyou">
<div id="question2" class="wb-fieldflow" data-wb-fieldflow="{
			'noForm':false,
                      	'renderas':'radio',
                      	'base': { 'live': false },
                      	'noreqlabel': true}">
<ul>
<li><a href="#1">l'enfant (single quote)</a></li>
<li><a href="#2">l&apos;enfant (apostrophe)</a></li>
</ul>
</div>
</div>
```

EricDunsworth:
Prevents single/double quotes from causing fieldflow to truncate its form fields' value attributes.

Examples of scenarios this helps with:
* i18n.btn variable containing double-quotes (fieldflow names its submit button after it)
* List items containing single quotes (fieldflow uses them as checkbox/option/radio button values)
* Form action attributes containing URL parameters with single quotes in their values (fieldflow creates hidden inputs for them)

Fixes #1744.

### DIFF
--- a/components/wb-fieldflow/fieldflow.js
+++ b/components/wb-fieldflow/fieldflow.js
@@ -150,10 +150,10 @@ var componentName = "wb-fieldflow",
 				$( formElm.parentElement ).addClass( formComponent );
 			} else if ( config.inline && !config.renderas ) {
 				stdOut = "<div class='wb-frmvld mrgn-bttm-md " + formComponent + "'><form><div class='input-group'><div id='" + bodyID + "'>";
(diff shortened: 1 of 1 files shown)


duboisp (commented):
Approved after a successful local testing.

duboisp (approved):
I tested locally and it works as expected.

PR #1973, opened 2022-05-20
Update and-or.css

Christopher-O:
Updated CSS to allow for using ul list without nested elements (e.g. header/paragraph tags).


duboisp (approved):
The corresponding working example are in PR #1974

PR #1975, opened 2022-05-20
Feedback: Config + deprecated working example fix

GormFrank:
We have to find a better way to define the feedbackPath config


duboisp:
FYI - https://github.com/wet-boew/GCWeb/blob/master/sites/includes/settings.liquid

_config.yml
```diff
@@ -6,8 +6,8 @@ website: "https://wet-boew.github.io/GCWeb/" # URL of your public facing website
 global:
   lang: fr
   feedbackPath:
-    en: "/sites/feedback/ajax/report-problem-en.html"
-    fr: "/sites/feedback/ajax/report-problem-fr.html"
+    en: "/GCWeb/sites/feedback/ajax/report-problem-en.html"
+    fr: "/GCWeb/sites/feedback/ajax/report-problem-fr.html"
```
EricDunsworth, line 10: FWIW I've just created #1989 to resolve this via Liquid's``relative_url`` filter and clean up a few related things.

IMO the ``relative_url`` filter is a better bet for this kind of thing since it works by deriving paths from a root folder (``site.baseurl``) that can represent either "/" in local Jekyll instances and "/GCWeb/" on GitHub Pages.

GormFrank:
Parts have been dealt with and others will be dealt with eventually; this one is not needed anymore.

PR #1976, opened 2022-05-24
Meli-melo: Create meli-melo 2022-05-tokens

delisma:
This pull request is to experiment GCDS design tokens

### DIFF
--- a/méli-mélo/2022-06-tokens/_variables.scss
+++ b/méli-mélo/2022-06-tokens/_variables.scss
@@ -1,0 +1,171 @@
+/*
+ Global variables for GCWeb
+ */
+// Canada.ca foundations (core)
+//   Typography
+//     Font families
+$gcds-core-font-families-heading: Lato;
+$gcds-core-font-families-body: "Noto Sans";
+$gcds-core-font-families-monospace: Menlo;
+$gcds-core-font-families-icons: "Glyphicons Halflings";
+
+//     Font sizes
+$gcds-core-font-size-base: 1.25; // Sets base font size to 20px
+$gcds-core-font-scale-base: 1.125;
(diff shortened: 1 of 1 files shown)


duboisp:
As discussed: https://github.com/wet-boew/GCWeb/blob/master/sites/theme.scss#L44-L48

EricDunsworth:
Just wondering if there's an estimated release timeline for this? Would it be better to resurrect #1684 instead?

My concern is that to this day there still isn't a "proper" way of creating standard-looking forms in GCWeb (over 4 years after its fonts were reworked in 2019). I'm not thinking about specialized components like [fieldflow](https://wet-boew.github.io/GCWeb/components/wb-fieldflow/fieldflow-en.html) or the [giant checkbox/radio button classes](https://wet-boew.github.io/GCWeb/components/gc-chckbxrdio/gc-chckbxrdio-en.html).

Anyone that tries creating "standard" forms in GCWeb that follow [WET's form validation plugin examples](https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html) and/or use [Bootstrap 3.4's form CSS patterns](https://getbootstrap.com/docs/3.4/css/#forms) are pretty much guaranteed to end up with tiny-looking forms.

Partial/Temporary workarounds like the ``gc-font-2019`` class aren't very ideal since they have a very narrow scope and are likely to  produce a strange mix of tiny and normal-sized form presentations. Buttons are even worse off (only way of getting properly-sized buttons is to deliberately misuse larger button classes or call to action buttons).

I guess the point I'm trying to make is that the piecemeal approach that's been used for forms in GCWeb up to this point isn't really working out... and I don't think it's due to lack of demand (this PR is the 4th attempt to deal with it after @jkshapiro tried in #1654 / #1676 and mine in #1684).

Thoughts?

PR #1978, opened 2022-05-25
Footer: Update top of page link's destination

EricDunsworth:
GCWeb's "Top of page"/"Haut de la page" link was previously pointing to ``#wb-cont``. That's the primary H1 heading's ID (usually top of the content area). The link text was misleading since it didn't actually go to the top of the page.

This resolves it by changing the link's destination to ``#wb-tphp``. That ID is fit for purpose. Older versions of WET also used top of page links that pointed to that ID.

### DIFF
--- a/docs/static-header-footer/bootstrap-3.html
+++ b/docs/static-header-footer/bootstrap-3.html
@@ -173,7 +173,7 @@
 		<dl id="wb-dtmd">
 			<dt>Date modified:</dt>
 			<dd>
-				<time>2022-05-18</time>
+				<time>2022-05-25</time>
 			</dd>
 		</dl>
 	</div>
@@ -210,7 +210,7 @@
 						<li><a href="https://www.canada.ca/en/transparency/privacy.html">Privacy</a></li>
 					</ul>
 				</nav>
-				<div class="col-xs-6 visible-sm visible-xs tofpg"> <a href="#wb-cont">Top of Page <span class="glyphicon glyphicon-chevron-up"></span></a> </div>
+				<div class="col-xs-6 visible-sm visible-xs tofpg"> <a href="#wb-tphp">Top of Page <span class="glyphicon glyphicon-chevron-up"></span></a> </div>
 				<div class="col-xs-6 col-md-3 col-lg-2 text-right"> <img src="https://wet-boew.github.io/themes-dist/GCWeb/GCWeb/assets/wmms-blk.svg" alt="Symbol of the Government of Canada"> </div>
 			</div>
 		</div>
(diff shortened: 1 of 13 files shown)


docs/static-header-footer/bootstrap-3.html
```diff
@@ -210,7 +210,7 @@ <h2 class="wb-inv">About this site</h2>
 						<li><a href="https://www.canada.ca/en/transparency/privacy.html">Privacy</a></li>
 					</ul>
 				</nav>
-				<div class="col-xs-6 visible-sm visible-xs tofpg"> <a href="#wb-cont">Top of Page <span class="glyphicon glyphicon-chevron-up"></span></a> </div>
+				<div class="col-xs-6 visible-sm visible-xs tofpg"> <a href="#wb-tphp">Top of Page <span class="glyphicon glyphicon-chevron-up"></span></a> </div>
```
EricDunsworth, line 213: The updated link destination would result in a broken link in this page since it lacks skip navigation links (the ``wb-tphp`` ID is usually set on the skipnav links' ``<ul>`` element).

@GormFrank Is that intentional? Should the skipnav links be re-added to this page?

EricDunsworth:
@GormFrank I noticed that [``docs/static-header-footer/bootstrap-4.html`` lacks a "Top of page" link](https://github.com/wet-boew/GCWeb/blob/1e1da58efdf844eae3617b04cb359837d02ab87f/docs/static-header-footer/bootstrap-4.html#L112-L113).

Is that by design? Should the skipnav links and "Top of page" link be re-added to that page too?

duboisp (commented):
We need to investigate if that was intentional or not.

EricDunsworth:
Gonna bring this out of draft to make it stand out more.

I don't have any further changes planned from my end. But still need @GormFrank / @delisma's feedback to be able to bring it forward.

delisma:
@EricDunsworth I don't really have any kind of opinion on this. If we want to bring people to the top let's link to `#wb-tphp`. Otherwise if we want to go to the first content header `#wb-cont`, we could change the text link for "top of content / haut de contenu", which sounds weird.

duboisp:
Once the PR #1990 would be merged, the "top of the page" link in the far far end of page will be removed. At lot of device support a mechanism to go on the top of the page. (ex. "home" keyboard key) Instead, if there is a documented use, we can consider to re-add may be in the main footer and keep that link working for all view ports not only mobile.

duboisp:
The top of the page was removed with the release of the site footer component version 4 in April 2023.

PR #1992, opened 2022-07-07
Lists: Adding flexbox list columns utility classes

Garneauma:
Adding utility classes to be able to display lists in a column format distributed horizontally. Possibility to have 1, 2, 3 or 4 columns.

### DIFF
--- a/components/list/list-en.html
+++ b/components/list/list-en.html
@@ -1,18 +1,18 @@
 ---
 {
-	"title": "List - Additional style - GCWeb theme V5",
+	"title": "List - Additional styles - GCWeb theme V5",
 	"language": "en",
 	"altLangPage": "list-fr.html",
 	"breadcrumbs": [
 		{ "title": "GCWeb home", "link": "https://wet-boew.github.io/themes-dist/GCWeb/index-en.html" }
 	],
 	"secondlevel": false,
-	"dateModified": "2018-11-07",
+	"dateModified": "2022-07-07",
 	"share": "true"
 }
 ---
 
-<p>The following style is an addition to list styling availble in <a href="https://wet-boew.github.io/wet-boew-styleguide/design/lists-en.html">WET-BOEW Style Guide</a> and it is exclusive to GCWeb v5 theme</p>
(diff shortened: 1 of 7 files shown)


sites/baseline/_screen-sm-min.scss
```diff
@@ -0,0 +1,9 @@
+/*
```
duboisp, line 1: Because we have a bunch of other file in "baseline" we will need to make it easier to understand from a file system perspective.
So I propose either:
* Prefix the screen specific CSS with the component name, like "_lists-screen-sm-min.scss"; or
* Move all list related scss into a subfolder of baseline named "lists".

components/list/list-en.html
```diff
@@ -32,3 +32,39 @@ <h3>Source code</h3>
...
+	<li>Item</li>
+	<li>Item</li>
+	<li>Item</li>
+	<li>Item</li>
+	<li>Item</li>
+	<li>Item</li>
+</ul>
+<h3>Source code</h3>
+<pre><code>
+&lt;ul class="list-col-xs-1 list-col-sm-2 list-col-md-3 list-col-lg-4"&gt;
```
duboisp, line 56: Just add some emphasis on the CSS class name

```suggestion
&lt;ul class="<strong>list-col-xs-1 list-col-sm-2 list-col-md-3 list-col-lg-4</strong>"&gt;
```

duboisp (dismissed):
Note: This is an incomplete review.

components/list/list-en.html
Garneauma: > Just add some emphasis on the CSS class name

Done. :)

sites/baseline/_screen-sm-min.scss
Garneauma: > Because we have a bunch of other file in "baseline" we will need to make it easier to understand from a file system perspective. So I propose either:
> 
> * Prefix the screen specific CSS with the component name, like "_lists-screen-sm-min.scss"; or
> * Move all list related scss into a subfolder of baseline named "lists".

Added a "lists" folder and moved all lists associated styles into it.

GormFrank (approved):

duboisp:
@GormFrank In what browser have you tested? Please check all supported one.

GormFrank:
@duboisp Tested on Chrome, Safari, Firefox and Edge; works on all of them :)

@garneauma Thank you for your contribution!

PR #1993, opened 2022-07-07
Search template: Personal information notification

Ricokola:
Related to: WET-235

### DIFF
--- a/templates/search/results-contextual-en.html
+++ b/templates/search/results-contextual-en.html
@@ -5,7 +5,7 @@
 	"altLangPage": "results-contextual-fr.html",
 	"pageType": "search",
 	"pageclass": "page-type-search",
-	"dateModified": "2018-12-21",
+	"dateModified": "2022-08-02",
 	"share": "false",
 	"deptfeature": false
 }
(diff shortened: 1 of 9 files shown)


templates/search/results-contextual-en.html
```diff
@@ -14,7 +14,7 @@
 <div class="well">
 	<form class="mrgn-tp-sm">
 		<label class="wb-inv" for="sch-inp">Search Government of Canada websites</label>
-		<input class="form-control full-width" spellcheck="false" autocomplete="off" dir="ltr" id="sch-inp" name="q" type="search" data-fusion-query />
+		<input class="form-control full-width" spellcheck="false" autocomplete="off" dir="ltr" id="sch-inp" name="q" type="search" data-fusion-query aria-describedby="pers-info" />
```
duboisp, line 17: We will need to change the id and make it more specific where it self define it's context which is for "Canada.ca" and for "Search".

templates/search/results-contextual-en.html
```diff
@@ -29,11 +29,10 @@
 			</div>
 			<div class="clearfix"></div>
 		</div>
+		<p id="pers-info" class="mrgn-tp-md">Please <strong>do not include sensitive personal information</strong> in the search box, such as your social insurance number, personal finance data and medical or work history.</p>
```
duboisp, line 32: The `id` will need to modified to include more context. This is to avoid collision and quickly know it's context of use.

duboisp (changes requested):
This is an incomplete review.

templates/search/results-contextual-en.html
GormFrank: I recommend `#gc-pi`

CC: @Ricokola @duboisp

duboisp:
@Ricokola please rename this PR title. There is no "removal" but instead it add a message into our search pages.

Ricokola:
> @Ricokola please rename this PR title. There is no "removal" but instead it add a message into our search pages.

The title has been updated has requested

GormFrank (approved):

PR #1995, opened 2022-07-12
Build: Resolving build warnings

Garneauma:
- Resolving all includes warnings by either adding the correct included file or including the expected HTML directly (and image dependencies).
- Converting Handlebars syntax to Liquid syntax where applicable.
- Removing "_wetboew-demos" and "méli-mélo-demos" from Jekyll build to save time as they are not used (demos are fetched from the "gcweb-compiled-demos" project on deployment).


Garneauma:
@polmih Can you review this PR ?

CC @GormFrank

_includes/templates/corp-info-org.html
```diff
@@ -0,0 +1,63 @@
...
+
+++++++++++++++++++++++++++++++++++
++ List of social media links
++
++ Used by organizational and institutional page
+++++++++++++++++++++++++++++++++++
+
+{%- endcomment -%}
+
+<section class="gc-crprt">
```
duboisp, line 11: We will probably need to move the associated style of `gc-crprt` either with its associated template or as an independent component (like when it's reuse by multiple template).

see: https://github.com/wet-boew/GCWeb/blob/master/components/_base.scss#L57

We might reconsider to rename this file too. The goal is by their location and by their file name we can easily link them together without deep diving into the code and without causing any breaking change according to our public versioning API. (here the CSS class name that was used). 

FYI - We have the following word in this context which might require some adaptation : "corp-info-org" + "gc-crprt" + "Organizational/Institutional template" + "Social media links" (as line 4) + "Corporate information"

_includes/templates/corp-info-inst.html
```diff
@@ -0,0 +1,71 @@
+{%- comment -%}
+
+++++++++++++++++++++++++++++++++++
++ List of social media links
```
duboisp, line 4: Is it about social media links or corporation information?

_includes/templates/corp-info-inst.html
```diff
@@ -0,0 +1,71 @@
+{%- comment -%}
```
duboisp, line 1: How different this one is compared to "_includes/templates/corp-info-org.html" from a component perspective?

Like identifying what is different and can be adapated.

components/components-en.html
```diff
@@ -25,7 +25,20 @@ <h2 id="download">Download link</h2>
 	<p>Download links are used for referencing and linking to non-HTML files on Canada.ca web pages.</p>
 	<div class="row">
 		<div class="col-sm-4">
-			{{>comp-download}}
+			<a class="gc-dwnld-lnk" href="#">
```
duboisp, line 28: We might need to define it as a independent component : `gc-dwnld-lnk` ; `gc-dwnld`; `gc-dwnld-txt`; `gc-dwnld-info`

duboisp (changes requested):
This is an incomplete review. 

Wow, thanks @Garneauma for this

Before to merge it we will need to:
* [ ] Review the file naming to ensure they match their associated component
* [ ] Identify a list of components that we will need to document individually, like the "gc-dwld-link"
* [ ] Test in a local environment each modified working example

_includes/templates/corp-info-org.html
```diff
@@ -0,0 +1,63 @@
...
+	{%- if page.language == "fr" -%}
+	<h2>Renseignements sur l’organisation</h2>
+	{%- else -%}
+	<h2>Corporateinformation</h2>
```
polmih, line 15: Missing blank space
<h2>Corporate information</h2>

templates/news/news-fr.html
```diff
@@ -10,4 +10,109 @@
 	"pageType": "news"
 }
 ---
-{{>placeholdercontent}}
+<h2>Heading 2 (<code>h2</code>)</h2>
```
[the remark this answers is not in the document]
Garneauma: > 

I agree. However, this would probably be for another PR as it was like this before this PR and will require a lot of work.

_includes/templates/corp-info-inst.html
Garneauma: > How different this one is compared to "_includes/templates/corp-info-org.html" from a component perspective?
> 
> Like identifying what is different and can be adapated.

The content in the "col-md-8" is different. There is probably a way to merge them and have parameters in the include, but it should be handled in a different PR.

duboisp (commented):
Preapprove at the technical review - This is only web site content + documentation + Minor change to GCWeb-Jekyll

Todo:
* [ ] local testing + review

_config.yml
```diff
@@ -57,3 +57,5 @@ exclude:
 - vendor/cache/
 - vendor/gems/
 - vendor/ruby/
+- _wetboew-demos
+- méli-mélo-demos
```
duboisp, line 61: What was the error related for those 2 folder. Those folder need to be build by Jekyll in order to spin up compiled working example of the méli-mélo and idem for the wet-boew demos rendered with the GCWeb theme.

I do understand there is some link issue where the fix need to be about the link not excluding them.

_includes/templates/corp-info-inst.html
duboisp, line 1: This include file is a clone. It's formal location should be in the "sample" folder of its owner component.

Check the example of gc-servinfo component and the temporary "sample" folder in "templates" folder

duboisp (changes requested):
My review is incomplete. I need to check if each include files are the right one.
---
So far here some comment/item to address

* [ ] Please rebase and fix the conflict for our main page (index-*.md)
* [ ] Move/duplicate all include files into their appropriate "sample" folder
* [ ] Allow Jekyll to compile/build locally wet-boew demos and compiled méli-mélo project
* [ ] @duboisp to finish the review

templates/organizational/organizational-arms-en.html
```diff
@@ -13,14 +13,14 @@
...
-<h1 property="name" id="wb-cont">{{{title}}}</h1>
+<h1 property="name" id="wb-cont">{{ page.title }}</h1>
 <div class="row">
 	<div class="col-sm-7 pull-left">
 		<p class="gc-byline">We are a federal institution that operates at arm’s length from the Government of Canada.</p>
 		<p>[Institution name] Lorem ipsum dolor sit amet, consectetur adipiscing elit. Cras sagittis massa sed libero eleifend, ac iaculis urna tempor. Cras sagittis massa sed libero eleifend, ac iaculis urna tempor.</p>
 	</div>
 	<div class="col-sm-5 col-xs-12 pull-right">
-		{{>follow}}
+		{% include templates/follow.html %}
```
duboisp, line 23: "Follow us", "most requested" are reused components which should be defined in their own folder.

They are documented separately in the design system
* https://design.canada.ca/common-design-patterns/social-media-channels.html
* https://design.canada.ca/common-design-patterns/most-requested.html

So the proper fix will be to create 2 new components:
*  "/components/followus/includes/"
*  "/components/most-requested/includes/"

or create sample here:
*  "/components/followus/samples/"
*  "/components/most-requested/samples/"

---

Maybe we can consider to use "include_relative" instead of "include" for this kind of situation where template are using other components or using a component "samples".

duboisp (changes requested):
I do want this PR being merged but there is a lot of different components to review. I am thinking it will be probably easier if we split this PR into multiple smaller PR where it will resolve build warning by component. Or have a PR with 2-3 component max to keep it easy to review. @Garneauma what do you think?

PR #2003, opened 2022-08-17
Footer: Add a page details i18n variable

EricDunsworth:
"Page details" was previously hardcoded into the footer include's visually-hidden ``H2`` heading. As a result, French pages were using the English version of that heading.

This resolves it by adding an ``i18nText-pageDetails`` variable and using it in place of the English heading content.


polmih (approved):
tested on local environment.

duboisp (approved):

PR #2008, opened 2022-08-31
Homepage: Your government section bg image update

Garneauma:
Updating background image of the "Your government" section on the homepage.

Changes related to WET-260


duboisp (commented):
Test to perform
* [x] Ensure we test the view port limit to ensure the background image do not overlap with the text. Especially when the text is zoomed at 400%.

Rename the original file instead of creating a new one.

duboisp (changes requested):
Replace the existing image with the new one

Garneauma:
> Replace the existing image with the new one

Tested the background size increase in all viewports and zoomed to 400%.

Replaced the image in the DAM, and then reverted to the current one (v1.1) until this PR is merged and released. When it is, simply "revert" to version 1.2 of the image.

duboisp (approved):

PR #2009, opened 2022-09-07
Jekyll - removing 'inc' from includes files

duboisp:
Doing some cleaning in preparation of officially releasing GCWeb-Jekyll theme. Note that additional cleanup are required but it will be submitted in another PR.

* [x] Test - Build locally and modified files should remains the same as it was.


Ricokola (approved):
Build and tested locally - Pass

PR #2012, opened 2022-09-14
Creating direct parent/child dependencies

Christopher-O:
Fixing parent child relationship so child elements are directly relational only to the direct parent, thus could allow for nested and/or lists inside (for more complex explaination).  Will update example page with what that would look like when merged.


duboisp:
Where you are with DTO about this feature?

duboisp (approved):

PR #2013, opened 2022-09-14
keeping stacked "and" design same as side-by-side

Christopher-O:
Avoiding a "double line" look when stacked, making the "and" design borders the same as side-by-side (top/bottom rather than left/right)


duboisp (approved):

duboisp:
Once the build is completed. We can squash and merge. This project will need to be compiled in a new méli-mélo compilation.

PR #2024, opened 2022-09-28
Splash page - Create French version + assets cleanup

Ricokola:
- Addition of the French first version
- Creation of the documentation
- Update assets to use same as other content pages for the FIP and Canada wordmark
- Deletion of previous assets


duboisp (approved):

duboisp:
We need this minor template addition to support a priority request at PP.

I tested locally and validated the markup. It work as expected and it eliminate the need to use duplicate SVG images. 

Later, we will only need to automate the list of instances and include and apply some properties update at our GCWeb website splash page.

PR #2025, opened 2022-09-28
SVG Interactive Map

karinapereznolasco:


duboisp (approved):
Approved, there is no interference with wet-boew or GCWeb

PR #2034, opened 2022-10-17
Index: Improve card layout and make WET demos pop

GormFrank:
This PR is blocked by WET PR #[to-come]


duboisp:
Look promising, Thanks @GormFrank

duboisp:
Pre-approve upon successful testing to be completed after the latest WET-BOEW is released.

index-en.md
```diff
@@ -382,8 +382,19 @@ css:
...
+				<div class="col-xs-3 col-md-2 d-flex align-self-center">
+					<img src="https://raw.github.com/wet-boew/wet-boew-attachments/master/Promo/WET_Logo.png" class="full-width" alt="">
+				</div>
+				<div class="col-xs-9 col-md-10">
+					<h3 class="mrgn-tp-md" id="wet-boew-demos">WET features styled with Canada.ca theme</h3>
+					<p><a href="/gcweb-compiled-demos/index.html#wet-boew">view list of demos</a></p>
+				</div>
+			</div>
+		</div>
+	</div>
```
duboisp, line 397: nice look

image

It will be nice to associate the image along with the h3. May be with a `<section>` element?
And probably we can use the `stretched-link` too

duboisp (changes requested):
When I built and tested locally with the latest wet-boew this was happening.

It might need some adjustment.

**Before to be expand**
image

**when left is expand**
image

**When right is expand**
image

**When both are expand**
image

GormFrank:
> When I built and tested locally with the latest wet-boew this was happening.
> 
> It might need some adjustment.

This is actually the intended behavior.

index-en.md
GormFrank: @duboisp Yes it would make sense to add a `<section>` element. The stretch link however would not include the image in the clickable area because columns are relatively positioned and stop the propagation of the stretch link.

GormFrank:
@Ricokola We need a tie breaker for the card layout proposal; it's you!

Ricokola (commented):
After testing the proposed approach, I have a preference for the current behaviour. Having the drop-down menu going at the bottom for the smallest, having to scroll back up to view the heading or to reach the content can become awkward depending on the case.

GormFrank:
@Ricokola @duboisp Thank you for taking final decision. I removed the new behavior, while keeping the new styling for WET demos.

duboisp (approved):

PR #2037, opened 2022-10-24
Build - Improve docker and manual additional ci HTML lint check

duboisp:


duboisp (commented):
To test in difference environment
* [x] In a Windows docker environment
* [x] In a Mac docker environment

GormFrank (changes requested):
Environment variables are not taken into account with Docker compose

duboisp:
I made the adjustment and on my side I successfully tested the `docker-compose up --build` on my Mac and on my Windows 10 computer.

Ricokola (changes requested):
On the first round of testing, I received an error regarding a missing folder: `/~sites`. After creating the folder `~sites` manually, I've been able to successful build GCWeb locally.

docs/developing-en.md
```diff
@@ -75,6 +78,35 @@ Where:
...
+grunt debug
+docker-compose up
+
+```
+
+HTML link checker
+
+
+```
+docker exec -i gcweb_jekyll rake
```
GormFrank, line 95: My container was called `gcweb_jekyll_1` and not just `gcweb_jekyll`. Plus, could be important to note that the container must be running in order to execute this command line.

GormFrank (approved):
Love that there will be an equivalent to the dist folder (in WET) to simply grab and drop on a server to run a website quickly :)

THank you @duboisp, quick comment only.

docs/developing-en.md
duboisp: Yeah, we would need to update our instruction. 

Let's get it fix when we will start to implement that "rake" task onto our CID script.

duboisp:
TODO when we will implement the rake in our CID.



*  update our instruction on how to find the docker image name and ensure is running.

Let's get it fix when we will start to implement that "rake" task onto our CID script.

PR #2041, opened 2022-10-26
Footer: removing flexbox utility classes

Garneauma:
Removing the following flexbox utility classes:
- `.d-flex`
- `.align-items-center`
- `.align-self-end`

This PR is dependent on: https://github.com/wet-boew/wet-boew/pull/9461

Changes related to WET-289

### DIFF
--- a/sites/footers/_base.scss
+++ b/sites/footers/_base.scss
@@ -218,17 +218,3 @@
 		/* END */
 	}
 }
-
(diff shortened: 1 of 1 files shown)


duboisp (approved):
I validated and those CSS are now included in GCWeb v12.1.0 through wet-boew dependency. It is now safe to remove those duplicates.

PR #2050, opened 2022-11-08
Méli-mélo - 2021-05-steps - Update steps.css

Christopher-O:
Complete re-structure of CSS code, slight redesign of circle pattern and introduction of potential alternative pattern.

PR #2052, opened 2022-11-08
Update index.html

Christopher-O:
Added new lead zero pattern as possible alternative or semi-replacement for existing pattern.  Governance will come soon.


duboisp:
Change in this PR was merged with #2050

PR #2054, opened 2022-11-10
Templates - Splash page - Added deprecated images

duboisp:
Added back the deprecated images for backward compatibility with the version 1 of the splash page.

Moving forward, we are going to include in our public versioning API the list of files that are directly linked by a GCWeb template.

Note: After one year from the date we merged this PR, we will be able to remove those 4 deprecated images at the following major build of GCWeb.


Ricokola (approved):
All images are properly named to match file names in previous versions.

PR #2058, opened 2022-11-25
GC-Subway: Stabilizing GC-Subway pattern

Garneauma:
Changes related to WET-197

### DIFF
--- a/components/gc-subway/_base.scss
+++ b/components/gc-subway/_base.scss
@@ -3,134 +3,157 @@
  * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  *
- *
- *	Provisional GC Subway map menu
 */
 
-@mixin dt-li($paddingBottom, $fcb-left, $fcb-width, $ahb-height, $ahb-left, $ahb-width ) {
-	border-left: 4px solid #26374a;
-	padding: 0px 20px $paddingBottom 1em;
-
-	a {
-		&::before {
-			background-color: #fff;
-			border: 3px solid #26374a;
-			border-radius: 50%;
-			content: "";
-			height: 1.2em;
-			left: $fcb-left;
-			position: absolute;
-			transition: width .2s, height .2s, left .2s, margin-top .2s;
-			width: $fcb-width;
-		}
-		&.active {
-			color: $text-color;
-			text-decoration: none;
-
-			&::before {
-				background-color: #26374a;
-			}
-
-			&:hover,
-			&:focus {
-				color: $link-hover-color;
-				text-decoration: underline;
-			}
-		}
-		&:not(.active) {
-			&:hover::before,
-			&:focus::before {
-				height: 1.5em;
-				left: -.1em;
-				margin-top: -.1em;
-				width: 1.5em;
-			}
-		}
-	}
+@mixin subway-dot {
+	background-color: $accent-blue;
+	border: 3px solid $accent-blue;
+	border-radius: 50%;
+	box-shadow: 0 0 0 10px #fff inset;
+	content: "";
+	height: 1.2em;
+	left: -.7em;
+	position: absolute;
+	top: 0;
+	transition: box-shadow .25s ease;
+	width: 1.2em;
 }
 
-.provisional {
+.gc-subway.provisional {
+	&:not(.gc-subway-index) {
+		border: 4px solid $accent-blue;
+		border-radius: 6px;
+		margin: 2em 0 0 .5em;
+		padding-bottom: 1em;
+		position: relative;
 
-	/*** GC Subway mobile first styles ***/
-	&.gc-subway {
-		border-radius: 0px 6px 6px 0px;
-		border-right: 4px solid #26374a;
-		border-top: 4px solid #26374a;
-		margin-top: 38px;
-
-		ul,
-		dl {
+		ul {
 			clear: both;
 			list-style: none;
-			padding-left: .57em;
-			padding-top: 10px;
-			position: relative;
-
-			dt {
-				@include dt-li(3px, 0em, 1.3em, 1.5em, -.1em, 1.5em);
-				font-weight: normal;
-				&:last-of-type {
-					border-left: 5px solid transparent;
-					padding-bottom: 0;
-				}
-			}
-
-			dd {
-				border-left: 4px solid #26374a;
-				line-height: inherit;
-				margin-bottom: 0px;
-				margin-top: -3px;
-				padding: 0px 20px 20px 1em;
-
-				&:last-child {
-					border-left: 5px solid transparent;
-					padding-bottom: 0;
-				}
-			}
+			margin-bottom: 0;
+			margin-left: -.685em;
+			padding-left: .5em;
 
 			li {
-				@include dt-li(30px, .05em, 1.2em, 1.4em, -.05em, 1.4em);
-				&::first-line {
-					line-height: 1;
-				}
+				border-left: 4px solid $accent-blue;
+				line-height: 1.25em;
+				padding-bottom: 1.25em;
+				padding-left: 1em;
+				position: relative;
+
 				&:last-child {
-					border-bottom: 4px solid #26374a;
-					border-bottom-left-radius: 6px;
-					border-left: 4px solid #26374a;
+					border-left-color: transparent;
+					padding-bottom: 0;
+				}
+
+				a {
+					display: inline-block;
+
+					&::before {
+						@include subway-dot;
+					}
+
+					&.active {
+						color: #333;
+						cursor: default;
+						text-decoration: none;
+
+						&::before {
+							box-shadow: 0 0 0 10px $accent-blue inset;
+						}
+
+						&:hover, &:focus {
+							color: #333;
+							text-decoration: none;
+						}
+					}
+
+					&:not(.active):hover::before,
+					&:not(.active):focus::before {
+						box-shadow: 0 0 0 4px #fff inset;
+					}
 				}
 
 				ul {
-					margin-top: 20px;
-					padding-left: .55em;
+					margin: 1em 0 0;
 
-					li:last-child {
-						border-bottom-width: 0px;
-						padding-bottom: 0px;
+					li {
+						&:last-child {
+							padding-bottom: 0;
+						}
+					}
+
+					&.noline {
+						li {
+							border-left-color: transparent;
+
+							&::before {
+								display: none;
+							}
+
+							a.active::after {
+								background-color: $accent-blue;
+								content: "";
+								display: block;
+								height: 4px;
+								left: -1.75em;
+								position: absolute;
+								top: .5em;
+								width: 1.125em;
+							}
+						}
 					}
 				}
 			}
 		}
-
-		h1 {
-			float: left;
-		}
 	}
+	&.gc-subway-index {
+		h2 {
+			position: static;
+		}
 
-	/* Double H1 */
-	&.gc-subway h1,
-	&.gc-subway-section .gc-subway-h1 {
-		background-color: #fff;
-		border-bottom-width: 0px;
-		color: #555;
-		font-size: 1.3em;
-		margin-right: 20px;
-		margin-top: -19px;
-		padding: 0px 20px 10px 0px;
+		dl {
+			margin-left: .5em;
+
+			dt, dd {
+				border-left: 4px solid $accent-blue;
+				font-weight: normal;
+				margin: 0;
+				padding-left: 1em;
+				position: relative;
+
+				&:last-of-type {
+					border-left-color: transparent;
+					padding-bottom: 0;
+				}
+			}
+
+			dt {
+				a::before {
+					@include subway-dot;
+				}
+			}
+
+			dd {
+				padding-bottom: 1.25em;
+				padding-top: .25em;
+			}
+		}
 	}
 }
 
-// Remove the border for index pages, when an h1 precede the subway nav.
-h1 ~ .provisional.gc-subway {
-	border-right: none;
-	border-top: none;
+.gc-subway-section {
+	hgroup {
+		p {
+			display: none;
+		}
+	}
+}
+
+.gc-subway-support {
+	margin-top: 3em;
+}
+
+.gc-subway-landmark-end {
+	@extend .clearfix;
 }
--- a/components/gc-subway/_screen-md-min.scss
+++ b/components/gc-subway/_screen-md-min.scss
@@ -1,54 +1,82 @@
 /*
- *
  * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  *
- *
- *	Provisional GC Subway map menu
+ * Medium view and over (screen only) - Variable $screen-md-min >= 992px
 */
 
-.provisional {
+/*** GC subway desktop styles ***/
 
-	/*** GC subway desktop styles ***/
-	&.gc-subway {
-		border-right: 0px;
-		border-top: 0px;
-		display: none;
-		margin-top: 25px;
-		padding-left: 15px;
-
+.gc-subway.provisional {
+	&:not(.gc-subway-index) {
 		&.no-blink {
 			display: block;
 		}
-		.gc-subway-menu-nav {
-			float: right;
-			width: 33.33%;
-		}
-		ul li:last-child {
-			border-bottom: 0px;
-			border-left: 4px solid transparent;
-		}
-	}
 
-	/* Double H1 */
-	&.gc-subway-section {
-		padding-right: 15px;
-		width: 66.66%;
+		border-color: transparent;
+		display: none;
+		float: right;
+		width: calc(33.33% - .5em);
 
-		h1, .gc-subway-h1 {
-			margin-top: 0px;
+		.wb-disable & {
+			display: block;
 		}
-		.gc-subway-h1 {
-			font-family: Lato, sans-serif;
-			font-weight: inherit;
-			margin-bottom: 0px;
-			margin-right: 0px;
-			padding-bottom: 0px !important;
-			padding-left: 0px;
+
+		hgroup {
+			left: -.5em;
+			position: absolute;
+			top: -2em;
+
+			h1 {
+				clip: rect(1px, 1px, 1px, 1px);
+				height: 1px;
+				margin: 0;
+				overflow: hidden;
+				position: absolute;
+				width: 1px;
+			}
+		}
+
+		ul {
+			li {
+				&:last-child:has(ul) {
+					border-left: 4px solid $accent-blue;
+					padding-bottom: 1.25em;
+
+					&::after {
+						background-color: $accent-blue;
+						bottom: 0;
+						content: "";
+						height: 4px;
+						left: -.45em;
+						position: absolute;
+						width: .75em;
+					}
+				}
+			}
 		}
 	}
 }
 
-.wb-disable .provisional.gc-subway {
-	display: block;
+.gc-subway-section {
+	float: left;
+	padding-right: 30px;
+	width: 66.66%;
+
+	hgroup:first-of-type {
+		h1 {
+			margin-top: 0;
+		}
+
+		p {
+			display: block;
+			font-size: 1.25em;
+			margin-bottom: 0;
+		}
+	}
+}
+
+.gc-subway-support {
+	float: right;
+	width: 33.33%;
 }
--- a/components/gc-subway/_screen-sm-max.scss
+++ b/components/gc-subway/_screen-sm-max.scss
@@ -2,15 +2,24 @@
  * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  *
- * Small view and under (screen only) - Variable $screen-sm-max <= 991 px
- *
- *	Provisional GC Subway map menu
+ * Small view and under (screen only) - Variable $screen-sm-max <= 991px
 */
 
-.provisional {
-	&.gc-subway {
-		dt, li {
-			max-width: initial;
+.gc-subway.provisional {
+	&:not(.gc-subway-index) {
+		h1 {
+			background-color: #fff;
+			border-bottom: none;
+			color: #555;
+			float: left;
+			font-size: 1.3em;
+			margin-left: -.5em;
+			margin-top: -.75em;
+			padding: 0px 20px 10px 0px;
+		}
+
+		ul {
+			padding-top: .5em;
 		}
 	}
 }
--- a/components/gc-subway/gc-subway.js
+++ b/components/gc-subway/gc-subway.js
@@ -10,20 +10,12 @@
 var $document = wb.doc,
 	componentName = "gc-subway",
 	selector = ".provisional." + componentName,
-	initEvent = "wb-init ." + componentName,
-	views = {
-		xxs: "xxsmallview",
-		xs: "xsmallview",
-		sm: "smallview",
-		md: "mediumview",
-		lg: "largeview",
-		xl: "xlargeview"
-	},
+	initEvent = "wb-init " + selector,
 	mainClass = "gc-subway-section",
-	toggleClass = "wb-inv",
-	desktopInited = false,
-	$html = wb.html,
-	$h1, $h2, $h1Copy, $menu, $main,
+	indexClass = "gc-subway-index",
+	sectionsTitle,
+	$navH1, $pageH1,
+	$subwayLinks,
 
 	/**
 	 * @method init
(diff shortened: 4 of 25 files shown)


components/gc-subway/index-en.html
```diff
@@ -27,56 +21,20 @@ <h1 property="name" id="wb-cont" class="gc-thickline">GC Subway Index</h1>
...
-		</dd>
-		<dt>
-			<a href="#">
-				[Page 6]
-			</a>
-		</dt>
-		<dd>
-			Page description. Lorem ipsum dolor sit amet, consectetur adipiscing elit
-		</dd>
+		<dt><a href="gc-subway-en.html">[Page 1]</a></dt>
```
GormFrank, line 27: Those are broken links, please link to different kinds of examples and add description to help support the link.

GormFrank (dismissed):
Quick review; some changes needed.

Merci @Garneauma :)

Garneauma:
To do:

- [ ] Translate documentation once review has passed.

duboisp (changes requested):
@Garneauma @GormFrank we will need to review and update our current internal guidance for double h1. The endless discussion on that topic with the web community continued since our last check and it seems to be resolved now. I do think the html spec guidance on [document outline](https://html.spec.whatwg.org/multipage/sections.html#headings-and-outlines-2) was updated too.

* [ ] Review and do research and update the [Design Decision 12](https://wet-boew.github.io/wet-boew-documentation/decision/12.html)
* [ ] Adapt this plugin based on the updated research

Garneauma:
> @Garneauma @GormFrank we will need to review and update our current internal guidance for double h1. The endless discussion on that topic with the web community continued since our last check and it seems to be resolved now. I do think the html spec guidance on [document outline](https://html.spec.whatwg.org/multipage/sections.html#headings-and-outlines-2) was updated too.
> 
> * [x]  Review and do research and update the [Design Decision 12](https://wet-boew.github.io/wet-boew-documentation/decision/12.html)
> * [x]  Adapt this plugin based on the updated research

Reviewed the reference for multiple H1 tags and it still supports it. It mentions: "If a document has one or more headings at least a single heading within the outline should have a heading level of 1". It also mentions the following: "A document can contain multiple top-level headings".

components/gc-subway/gc-subway.js
```diff
@@ -35,93 +29,82 @@ var $document = wb.doc,
...
+				if ( h1 ) {
+					h1.id = h1.id || wb.getId(); // Ensure the element has an ID
+					wb.addSkipLink( wb.i18n( "skip-prefix" ) + " " + h1.textContent, { href: "#" + h1.id } );
+				}
+
+				// Wrap all content until it hits either: ".pagedetails, .gc-subway-section-end, .gc-subway-support"
+				$elm.nextUntil( ".pagedetails, .gc-subway-section-end, .gc-subway-support" ).wrapAll( "<section class='" + mainClass + "'>" );
+
+				$main = $elm.next( "." + mainClass );
+				$( "<p class='gc-subway-h1' aria-hidden='true'>" + h1.textContent + "</p>" ).prependTo( $main );
```
duboisp, line 61: We would need to use a pattern that leverage the `<hgroup>` element

duboisp (changes requested):

Garneauma:
@duboisp GC Subway is now ready for a full review.

duboisp:
I will do the review + testing but we might need to technical documents (and review existing documentation) on how to code stack header and double h1 prior to mark it "stable".

Meanwhile, we would review and accept those change but keep it "provisional" for now until the technical guidance is completed.

Garneauma:
@duboisp J'ai remis le plugin en "provisional".

components/gc-subway/section-end-fr.html
```diff
@@ -1,58 +1,42 @@
...
 
-<ul class="list-inline">
-	<li><a class="btn btn-primary" href="docs-fr.html">Documentation</a></li>
-	<li><a class="btn btn-primary" href="https://github.com/wet-boew/GCWeb/issues/new?title=gc-subway">Questions ou commentaires?</a></li>
-</ul>
+{%- include gc-subway/docs-questions.html -%}
 
-<nav class="provisional gc-subway">
-	<h1 id="gc-document-nav">[Nom du service]</h1>
+<nav class="gc-subway provisional" data-gc-subway='{"subtitle": "My sections"}'>
```
duboisp, line 14:
```suggestion
<nav class="gc-subway provisional">
```

Is `data-gc-subway` value can be extracted or it need to be provided by the author?

components/gc-subway/section-end-fr.html
```diff
@@ -1,58 +1,42 @@
...
 <p>Sed consectetur eu ligula a molestie. Vivamus convallis libero malesuada pharetra suscipit. In a pulvinar mi, quis aliquet mauris. Duis convallis nunc nunc, in euismod nisi volutpat sit amet. Integer convallis lacus non orci imperdiet, ac convallis massa mollis. Aliquam erat volutpat. Ut maximus euismod auctor. Ut ac gravida nunc. Nam non efficitur neque. Pellentesque tincidunt, libero luctus condimentum laoreet, turpis magna maximus nibh, at cursus lectus tellus in augue. Aenean scelerisque eros dui, at tincidunt eros tristique nec.</p>
 <p>Duis vel lacus elit. Mauris rhoncus urna at enim fermentum, eget fermentum est ornare. Pellentesque porttitor bibendum venenatis. Nam tempor vel felis a fringilla. Integer dui nisl, sagittis nec risus a, fermentum euismod purus. Vivamus dictum, elit quis varius eleifend, purus leo scelerisque metus, id fermentum nulla neque eget elit. Etiam vel vestibulum justo, a scelerisque eros.</p>
+
+<div class="gc-subway-section-end"></div>
```
duboisp, line 32: What is the purpose of this marker, should use semantic?

components/gc-subway/section-end-fr.html
```diff
@@ -1,58 +1,42 @@
...
-	"altLangPage": "gc-subway-en.html",
-	"pageclass": "cnt-wdth-lmtd",
+	"altLangPage": "section-end-en.html",
 	"layout": "without-h1",
-	"dateModified": "2022-08-25",
+	"dateModified": "2023-01-04",
 	"share": "true"
 }
 ---
-{% include alert-provisional.html provisionaldir="../" %}
```
duboisp, line 12: Missing the provisional alert:

```
{% include alert-provisional.html siteroot="../../" %}
```

components/gc-subway/section-rel-info-en.html
```diff
@@ -0,0 +1,47 @@
...
+	"title": "[Page 1] - GC Subway section page example with related information",
+	"language": "en",
+	"altLangPage": "section-rel-info-fr.html",
+	"layout": "without-h1",
+	"dateModified": "2023-01-04",
+	"share": "true"
+}
+---
+
+{%- include gc-subway/docs-questions.html -%}
```
duboisp, line 12: Missing the provisional alert:

```
{% include alert-provisional.html siteroot="../../" %}
```

components/gc-subway/includes/docs-questions.html
```diff
@@ -0,0 +1,9 @@
+<ul class="list-inline">
```
duboisp, line 1: No sure if this file fit under "includes" neither under "samples". But it do help to maintain repetition on the sub-pages. Let's think where it will be fit best, for sure not in the "includes" folder.

components/gc-subway/index-en.html
```diff
@@ -1,24 +1,18 @@
...
-{% include alert-provisional.html provisionaldir="../" %}
+<h1 property="name" id="wb-cont" class="gc-thickline">GC Subway index</h1>
 
-<h1 property="name" id="wb-cont" class="gc-thickline">GC Subway Index</h1>
-
-<ul class="list-inline">
-	<li><a class="btn btn-primary" href="docs-en.html">Documentation</a></li>
-	<li><a class="btn btn-primary" href="https://github.com/wet-boew/GCWeb/issues/new?title=gc-subway">Questions or comments?</a></li>
-</ul>
+{%- include gc-subway/docs-questions.html -%}
```
duboisp, line 15: Use include relative

components/gc-subway/section-rel-info-en.html
```diff
@@ -0,0 +1,47 @@
...
+<p>Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Nunc aliquam felis massa, ut congue enim faucibus in. Vivamus rhoncus elit in pharetra bibendum. In fringilla justo non consectetur feugiat. Ut bibendum nisi lacus, eu dapibus massa semper sit amet. Fusce condimentum pretium libero eget fermentum. Duis feugiat velit at commodo accumsan. Nunc in felis leo.</p>
+
+<nav class="gc-subway-pagination">
+	<h2 class="wb-inv">Document navigation</h2>
+	<ul class="pager">
+		<li class="next"><a href="#" rel="next"><span class="wb-inv">Next: </span>[Page 2]</a></li>
+	</ul>
+</nav>
+
+<aside class="gc-subway-support">
```
duboisp, line 41: Food for thought: Do we need a special CSS class or can we leverage "pull-right" + the grid.

components/gc-subway/section-rel-info-fr.html
```diff
@@ -0,0 +1,47 @@
...
+	"title": "[Page 1] - GC Métro - Exemple de page de section avec information additionnelle",
+	"language": "fr",
+	"altLangPage": "section-rel-info-en.html",
+	"layout": "without-h1",
+	"dateModified": "2023-01-04",
+	"share": "true"
+}
+---
+
+{%- include gc-subway/docs-questions.html -%}
```
duboisp, line 12: Missing the provisional alert:

```
{% include alert-provisional.html siteroot="../../" %}
```

duboisp (changes requested):
The CSS need be reinforced by "provisional" for ".gc-subway-index" and ".gc-subway" or it need to make the plugin unusable if the "provisional" are not provided.

components/gc-subway/section-end-en.html
```diff
@@ -1,58 +1,42 @@
...
 <p>Sed consectetur eu ligula a molestie. Vivamus convallis libero malesuada pharetra suscipit. In a pulvinar mi, quis aliquet mauris. Duis convallis nunc nunc, in euismod nisi volutpat sit amet. Integer convallis lacus non orci imperdiet, ac convallis massa mollis. Aliquam erat volutpat. Ut maximus euismod auctor. Ut ac gravida nunc. Nam non efficitur neque. Pellentesque tincidunt, libero luctus condimentum laoreet, turpis magna maximus nibh, at cursus lectus tellus in augue. Aenean scelerisque eros dui, at tincidunt eros tristique nec.</p>
 <p>Duis vel lacus elit. Mauris rhoncus urna at enim fermentum, eget fermentum est ornare. Pellentesque porttitor bibendum venenatis. Nam tempor vel felis a fringilla. Integer dui nisl, sagittis nec risus a, fermentum euismod purus. Vivamus dictum, elit quis varius eleifend, purus leo scelerisque metus, id fermentum nulla neque eget elit. Etiam vel vestibulum justo, a scelerisque eros.</p>
+
+<div class="gc-subway-section-end"></div>
```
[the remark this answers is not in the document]
duboisp: Is it an "landmark"?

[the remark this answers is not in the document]
duboisp: Or can we leverage and kind of enforce the content width limited feature, CSS: `cnt-wdth-lmtd`

components/gc-subway/section-end-en.html
```diff
@@ -1,58 +1,42 @@
...
+
 <p>Nam elementum sit amet sem id finibus. Vivamus et sagittis dolor. Curabitur semper libero ac maximus viverra. Sed sed mollis nulla. Quisque ornare sapien dictum elit fermentum, ut efficitur sem placerat. Praesent nulla enim, venenatis at finibus quis, semper eu nisl. Nam finibus ante vel lacus accumsan, vitae tincidunt massa porta. Donec ullamcorper pharetra ex et dapibus. Pellentesque et viverra nibh. Pellentesque non ante sodales, interdum sapien vitae, gravida risus.</p>
 <p>Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Nunc aliquam felis massa, ut congue enim faucibus in. Vivamus rhoncus elit in pharetra bibendum. In fringilla justo non consectetur feugiat. Ut bibendum nisi lacus, eu dapibus massa semper sit amet. Fusce condimentum pretium libero eget fermentum. Duis feugiat velit at commodo accumsan. Nunc in felis leo.</p>
 
-<nav class="mrgn-bttm-lg mrgn-tp-lg">
+<nav class="gc-subway-pagination">
```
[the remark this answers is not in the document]
duboisp: Check the navigation pattern used by the step form plugin.

components/gc-subway/section-end-fr.html
duboisp: Est-ce qu'on peut modifié le design pattern pour l'inclure dans le HTML à la place d'une configuration

Example/idée lorsque Enhanced:
```
<h1 class="h3"> <span class="hidden-md visible-xs">[Service name]</span> <span class="landmark-name hidden-xs visible-md visible-lg mrgn-tp-0">Sections</span></h1>
```


Example/idée lorsque coded by author:
```
<h1>[Service name] <span class="landmark-name">Sections</span></h1>
```
où que la class "landmark-name" serait caché par défaut (invisible)

Je sais qu'on a une demande pour identifier les landmark, peut-être: (Id or attribute+JS)
```
<nav aria-labeledby="gc-subway-landmark-name">
    <h1>[Service name] <span id="gc-subway-landmark-name" data-landmark>Sections</span></h1>
</nav>
```

components/gc-subway/includes/docs-questions.html
duboisp: What about "fragments" folder and use "include_relative"

components/gc-subway/section-rel-info-en.html
duboisp: This might create challenges about meaningful sequence SC

components/gc-subway/section-end-en.html
```diff
@@ -1,58 +1,44 @@
...
+
+<div class="gc-subway-landmark-end"></div>
+
 <p>Nam elementum sit amet sem id finibus. Vivamus et sagittis dolor. Curabitur semper libero ac maximus viverra. Sed sed mollis nulla. Quisque ornare sapien dictum elit fermentum, ut efficitur sem placerat. Praesent nulla enim, venenatis at finibus quis, semper eu nisl. Nam finibus ante vel lacus accumsan, vitae tincidunt massa porta. Donec ullamcorper pharetra ex et dapibus. Pellentesque et viverra nibh. Pellentesque non ante sodales, interdum sapien vitae, gravida risus.</p>
 <p>Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Nunc aliquam felis massa, ut congue enim faucibus in. Vivamus rhoncus elit in pharetra bibendum. In fringilla justo non consectetur feugiat. Ut bibendum nisi lacus, eu dapibus massa semper sit amet. Fusce condimentum pretium libero eget fermentum. Duis feugiat velit at commodo accumsan. Nunc in felis leo.</p>
 
-<nav class="mrgn-bttm-lg mrgn-tp-lg">
+<nav class="gc-pagination">
```
duboisp, line 39: If this is it's own component, we will need to create a "gc-pagination" component

components/gc-subway/section-end-fr.html
duboisp: Should that title be connected to the h1 instead of the nav? That is essentially an alternate title of when it get combined with the other h1.

components/gc-subway/_screen-md-min.scss
```diff
@@ -1,54 +1,83 @@
...
- *
- *	Provisional GC Subway map menu
+ * Medium view and over (screen only) - Variable $screen-md-min >= 992px
 */
 
-.provisional {
+/*** GC subway desktop styles ***/
+
+.gc-subway {
+	display: none;
```
duboisp, line 11: This line seem to break the index page

duboisp (changes requested):
Todo:
* [ ] The index page are broken, the plugin don't transform as expected
  Seem to be caused by
```css
@media screen and (min-width: 992px)
.gc-subway {
  display: none;
}
```
---

Note to ourselves: Update the broken link in the provisional alert include

---

Can be done later, but meanwhile it might create confusion is we demo it like this.
* [ ] Make a "gc-pagination" component
* [ ] Move/implement alternate for: `data-gc-subway='{"subtitle": "My sections"}'`. Like get it semantically connected with the h1 as an alternate title when the h1 "[Nom du service]" are merged with the other h1 "[Page 1]". OR should we ask the author to code an "hgroup" inside that nav where it identify the paragraph as being the subway alternate title.
* [ ] Explore if we can remove the need for the landmark: "gc-subway-landmark-end"
* [ ] Explore if `gc-subway-support` can leverage only CSS utilities instead of being fully integrated into the component pattern. Also validate if this cause a meaningful sequence issue.

components/gc-subway/section-end-en.html
Garneauma: I have reverted. Let's explore this when stabilizing the plugin.

duboisp (approved):
Thanks for the quick edits.

I reviewed and tested and I do confirm the index page is working as expected. The `gc-pagination` component was removed in favours of CSS utilities. And the CSS was re-enforced with the `provisional` CSS class.

duboisp (approved):
Last commit are only about spacing cleanup which don't impact my previous review.

components/gc-subway/_screen-sm-max.scss
```diff
@@ -2,15 +2,24 @@
...
+			color: #555;
+			float: left;
+			font-size: 1.3em;
+			margin-left: -.5em;
+			margin-top: -.75em;
+			padding: 0px 20px 10px 0px;
+		}
+
+		ul {
+			padding-top: .5em;
```
Christopher-O, line 22: Is this needed?  It adds a "hat" to the first item in mobile, removing it still has a "hat" but smaller

components/gc-subway/_screen-sm-max.scss
```diff
@@ -2,15 +2,24 @@
...
+	&:not(.gc-subway-index) {
+		h1 {
+			background-color: #fff;
+			border-bottom: none;
+			color: #555;
+			float: left;
+			font-size: 1.3em;
+			margin-left: -.5em;
+			margin-top: -.75em;
+			padding: 0px 20px 10px 0px;
```
Christopher-O, line 18: please add margin-right: 0.5em to avoid title cutting off entire top  bar if title word wraps
image

components/gc-subway/_base.scss
```diff
@@ -3,134 +3,157 @@
...
+		}
 	}
 }
 
-// Remove the border for index pages, when an h1 precede the subway nav.
-h1 ~ .provisional.gc-subway {
-	border-right: none;
-	border-top: none;
+.gc-subway-support {
+	margin-top: 3em;
```
Christopher-O, line 154: Can this be removed, or the amount shrunk significantly?  Adds a huge amount of space between menu and related information. 
image

components/gc-subway/section-rel-info-en.html
```diff
@@ -0,0 +1,49 @@
...
+<p>Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Nunc aliquam felis massa, ut congue enim faucibus in. Vivamus rhoncus elit in pharetra bibendum. In fringilla justo non consectetur feugiat. Ut bibendum nisi lacus, eu dapibus massa semper sit amet. Fusce condimentum pretium libero eget fermentum. Duis feugiat velit at commodo accumsan. Nunc in felis leo.</p>
+
+<nav class="mrgn-bttm-lg mrgn-tp-lg">
+	<h2 class="wb-inv">Document navigation</h2>
+	<ul class="pager">
+		<li class="next"><a href="#" rel="next"><span class="wb-inv">Next: </span>[Page 2]</a></li>
+	</ul>
+</nav>
+
+<aside class="gc-subway-support">
```
Christopher-O, line 43: My original update to the script moved the dom order of the related information to either below the section menu (desktop) or below the content in mobile to keep with expected dom order.  Was there a reason that was not implemented?   Does this pass WCAG 2.2?

components/gc-subway/section-end-en.html
```diff
@@ -1,58 +1,44 @@
 ---
```
Christopher-O, line 1: What is the purpose of this function?  The subway pattern is usually self contained and should not need to be escaped.  The example makes the pattern look broken.  could the content still be forced into the same column width? 
image

components/gc-subway/gc-subway.js
```diff
@@ -35,93 +27,67 @@ var $document = wb.doc,
...
+				$subwayLinks = $( selector + " a, ." + mainClass + " .gc-pagination a" );
 
-		// Desktop view, setup and mutate H1s
-		if ( $html.hasClass( views.md ) || $html.hasClass( views.lg ) ||
-			$html.hasClass( views.xl ) ) {
+				// Duplicating GC-Subway links for single-page application feel on mobile
+				$subwayLinks.each( function( i, el ) {
+					let $el = $( el ),
+						elHref = $el.attr( "href" ),
+						cloneHref = elHref.includes( "#" ) ? elHref : elHref += "#wb-cont";
```
Christopher-O, line 70: Should this also build a secondary link in the pagination to go to the second h1?

Christopher-O (commented):
Added comments on some of the implementation of design or function.

duboisp:
FYI - Chris suggestion/review internal PP reference number is wet-314

components/gc-subway/_screen-sm-max.scss
Garneauma: @Christopher-O  I actually think it looks better that way.

Christopher-O: @Garneauma I was mentioning to someone that out of all the comments this one was the only one I wouldn't mind as much if it stayed.  However, it does creates more inconsistency with the desktop version that does not have that "hat".

components/gc-subway/gc-subway.js
Garneauma: @Christopher-O You are right, it should. As you see on line 64, it gets the links within the pagination. Unfortunately, I removed the class not remembering it was used in the JS. We'll fix it.

components/gc-subway/_base.scss
Christopher-O: Can we add a border, or something else?  the padding looks like it's broken.  Also, per my comments below, there was a script that moved the dom order so that was never an issue

components/gc-subway/section-end-en.html
Christopher-O: @Garneauma could you let me know some examples of what elements broke the plugin?  This is the first I'm hearing/knowing about it.   I do hope it does get patched because again, it looks like its broken here.  If maybe there is a class that can keep the width the same, without resorting to col-md-8/col-md-4?

components/gc-subway/_screen-sm-max.scss
Garneauma: @Christopher-O  I see your point. However, the mobile version also has a border that goes around the whole nav, so I don't see this "hat" as an inconsistency. Another option would be to remove that border altogether on mobile, that way it would really be consistent with the desktop version.

components/gc-subway/section-end-en.html
Christopher-O: @Garneauma  Okay thanks, because I realized that if you have this "end" feature, but also want to add related information, it doesn't seem to work because semantically you would put the related information after the pagination links, but if they are outside of the subway design, they no longer "float right".

components/gc-subway/_screen-sm-max.scss
Christopher-O: @Garneauma I'm okay with leaving that in if it's agreed upon.

components/gc-subway/section-end-en.html
Garneauma: @Christopher-O This is why it is specified in the documentation that those two features cannot be implemented together: https://github.com/wet-boew/GCWeb/pull/2058/files#diff-588c6c74983f3776844bf5576148e39badcf927214777b3873ef5388e76b4938R116-R117

Christopher-O: I just saw that in passing as I was reviewing the governance, and why I added a phrase in my governance notes that it's important to review the governance before proceeding to the examples first.  Hopefully that can be rectified before becoming stable.

components/gc-subway/_base.scss
GormFrank: Hi @Christopher-O, we had consensus at PP that a reasonable gap was enough to break the sequence while not looking odd in terms of usability. I agree that we could look into shrinking a bit that gap, but all in all for performances reasons, as well as simplicity and accessibility, we decided that the gap is the way to go unless proven otherwise through something like usability testing.

duboisp: May be we can explore a few alternative, like Chad proposed to have a smaller gap or use a top border (hr section separator)

Like proposing a few prototype for discussion on alternative or to keep this one with the 3em space in between.

@GormFrank how much performance hit this will create? Like if we duplicate the "related links" in the right column and use CSS visibility to show/hide it in the different view ports.

PR #2064, opened 2022-12-02
Fieldflow: Support empty values in query actions

EricDunsworth:
Providing a ``value`` in the ``query`` action causes fieldflow to set it as the ``select`` option's value when that action runs. But it previously wasn't working for empty values.

This fixes it by changing a couple of truthy ``if`` conditions into ``typeof`` string checks.

### DIFF
--- a/components/wb-fieldflow/alternative-en.html
+++ b/components/wb-fieldflow/alternative-en.html
@@ -183,3 +183,151 @@ dateModified: "2021-04-28"
 		&lt;/ul&gt;
 	&lt;/div&gt;</code></pre>
 </section>
+
+<section>
+	<h2><code>query</code> action, use of empty value</h2>
+	<h3>For individual <code>query</code> action</h2>
+	<div class="wb-frmvld">
+		<form action="alternative-en.html" method="get">
+			<div class="wb-fieldflow hidden" data-wb-fieldflow='{
+				"noForm": true,
+				"defaultselectedlabel": false
+			}'>
+				<p>Pick an option</p>
+				<ul>
+					<li data-wb-fieldflow='[
+						{ "action": "query", "name": "q1", "value": "" }
+					]'>Custom: Make your selection...</li>
+					<li data-wb-fieldflow='[
+						{ "action": "query", "name": "q1", "value": "opt1" }
+					]'>Option 1</li>
+					<li data-wb-fieldflow='[
+						{ "action": "query", "name": "q1", "value": "opt2" }
+					]'>Option 2</li>
+				</ul>
+			</div>
+			<button>Submit</button>
+		</form>
+	</div>
+	<pre class="mrgn-tp-md"><code>&lt;div class=&quot;wb-frmvld&quot;&gt;
+	&lt;form action=&quot;alternative-en.html&quot; method=&quot;get&quot;&gt;
+		&lt;div class=&quot;wb-fieldflow hidden&quot; data-wb-fieldflow='{
+			&quot;noForm&quot;: true,
+			&quot;defaultselectedlabel&quot;: false
+		}'&gt;
+			&lt;p&gt;Pick an option&lt;/p&gt;
+			&lt;ul&gt;
+				&lt;li data-wb-fieldflow='[
+					{ &quot;action&quot;: &quot;query&quot;, &quot;name&quot;: &quot;q1&quot;, &quot;value&quot;: &quot;&quot; }
+				]'&gt;Custom: Make your selection...&lt;/li&gt;
+				&lt;li data-wb-fieldflow='[
+					{ &quot;action&quot;: &quot;query&quot;, &quot;name&quot;: &quot;q1&quot;, &quot;value&quot;: &quot;opt1&quot; }
+				]'&gt;Option 1&lt;/li&gt;
+				&lt;li data-wb-fieldflow='[
+					{ &quot;action&quot;: &quot;query&quot;, &quot;name&quot;: &quot;q1&quot;, &quot;value&quot;: &quot;opt2&quot; }
+				]'&gt;Option 2&lt;/li&gt;
+			&lt;/ul&gt;
+		&lt;/div&gt;
+		&lt;button&gt;Submit&lt;/button&gt;
+	&lt;/form&gt;
+&lt;/div&gt;</code></pre>
+
+	<h3>When the new default global action is <code>query</code></h3>
+	<div class="wb-frmvld">
+		<form action="alternative-en.html" method="get">
+			<div class="wb-fieldflow hidden" data-wb-fieldflow='{
+				"noForm": true,
+				"defaultselectedlabel": false,
+				"action": "query", "name": "q2", "prop": "value"
+			}'>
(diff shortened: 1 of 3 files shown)


EricDunsworth:
I've provided some basic examples for this PR's use case below.

They're using [``"defaultselectedlabel": false`` to disable the default option](https://wet-boew.github.io/GCWeb/components/wb-fieldflow/basic-en.html#noemptydefault) and substituting it with a "custom" default option. The advantage of doing it that way is that it can allow other actions to be tied to the default option (not shown in the examples).

**Testing:**
Feel free to use the examples to compare what happens before and after this PR...
* **Before this PR:**
  * The custom default option's ``value`` attribute will show up as ``value="CUSTOM: Make your selection..."``
  * Selecting option 1 or 2 and then going back to the default option **won't** show a form validation error
  * The form **will** be submittable with the default option selected (even though the fieldflow question is required)
* **After this PR:**
  * The custom default option's ``value`` attribute will show up as ``value=""``
  * Selecting option 1 or 2 and then re-selecting the default option **will** show a form validation error
  * The form **won't** be submittable unless option 1 or 2 are selected

**Examples:**
```html
<h2>Example 1 (query actions)</h2>
<div class="wb-frmvld">
	<form action="http://localhost/" method="post">
		<div class="wb-fieldflow hidden" data-wb-fieldflow='{
			"noForm": true,
			"defaultselectedlabel": false
		}'>
			<p>Pick an option</p>
			<ul>
				<li data-wb-fieldflow='[
					{ "action": "query", "name": "q1", "value": "" }
				]'>CUSTOM: Make your selection...</li>
				<li data-wb-fieldflow='[
					{ "action": "query", "name": "q1", "value": "opt1" }
				]'>Option 1</li>
				<li data-wb-fieldflow='[
					{ "action": "query", "name": "q1", "value": "opt2" }
				]'>Option 2</li>
			</ul>
		</div>
		<button>Submit</button>
	</form>
</div>

<h2>Example 2 (query action with prop values)</h2>
<div class="wb-frmvld">
	<form action="http://localhost/" method="post">
		<div class="wb-fieldflow hidden" data-wb-fieldflow='{
			"noForm": true,
			"defaultselectedlabel": false,
			"action": "query", "name": "q2", "prop": "value"
		}'>
			<p>Pick an option</p>
			<ul>
				<li data-wb-fieldflow=''>CUSTOM: Make your selection...</li>
				<li data-wb-fieldflow='opt1'>Option 1</li>
				<li data-wb-fieldflow='opt2'>Option 2</li>
			</ul>
		</div>
		<button>Submit</button>
	</form>
</div>
```

duboisp:
Pre-approved upon code review, local testing and after the example (in comments) is added to GCWeb like here: https://wet-boew.github.io/GCWeb/components/wb-fieldflow/alternative-en.html.

duboisp:
@EricDunsworth 
> The advantage of doing it that way is that it can allow other actions to be tied to the default option (not shown in the examples).

Can you provide an example of that situation? I would like to adapt the working example for that specific edge case, otherwise setting a value label to the properties `defaultselectedlabel` it is already supported by the plugin and will do what you expect.

Do you mean to do some grouping with the option? if so that is already supported too.

Other than the use case that need to be clarified, the propose change (code review + local testing) work as expected and I partially added + translated the 2 working example you suggested.

EricDunsworth:
> Can you provide an example of that situation? I would like to adapt the working example for that specific edge case

Here's another example that demonstrates extra actions tied to a custom default option:
```html
<h2>Example 3 (query/addClass/removeClass actions)</h2>
<div class="wb-frmvld">
	<form action="http://localhost/" method="post">
		<div class="wb-fieldflow hidden" data-wb-fieldflow='{
			"noForm": true,
			"defaultselectedlabel": false
		}'>
			<p>Pick an option</p>
			<ul>
				<li data-wb-fieldflow='[
					{ "action": "query", "name": "q3", "value": "" },
					{ "action": "addClass", "source": ".redgreen", "class": "text-danger", "live": true },
					{ "action": "removeClass", "source": ".redgreen", "class": "text-success", "live": true }
				]'>CUSTOM: Make your selection...</li>
				<li data-wb-fieldflow='[
					{ "action": "query", "name": "q3", "value": "opt1" },
					{ "action": "addClass", "source": ".redgreen", "class": "text-success", "live": true },
					{ "action": "removeClass", "source": ".redgreen", "class": "text-danger", "live": true }
				]'>Option 1</li>
				<li data-wb-fieldflow='[
					{ "action": "query", "name": "q3", "value": "opt2" },
					{ "action": "addClass", "source": ".redgreen", "class": "text-success", "live": true },
					{ "action": "removeClass", "source": ".redgreen", "class": "text-danger", "live": true }
				]'>Option 2</li>
			</ul>
		</div>
		<p class="redgreen">Red when default option is selected <strong>or</strong> green if a real option selected.</p>
		<button>Submit</button>
	</form>
</div>
```

Example 3 fails WCAG 2.x and doesn't represent a real-world use case. But it shows off the basic concept.

There could potentially be legit use cases for tying actions to a custom default option. But I can't think of a really good example off the top of my head.

What brought about this PR in the first place was that I was trying to prototype a fieldflow form that used the ``addClass``/``removeClass`` actions to toggle a ``nojs-show`` class that unveiled extra optional fields. I was trying to use the ``removeClass`` action on a custom default option to hide all the extra fields if a user switched back to it. But I later realized that the ``toggle`` action can produce similar results without needing a custom default option.

In any case, I felt like the status quo was a bug and decided to send this PR to fix it.

> otherwise setting a value label to the properties ``defaultselectedlabel`` it is already supported by the plugin and will do what you expect.

It's currently possible to do something along the lines of ``"defaultselectedlabel": "Customized make your selection text"`` to override the "Make your selection..." text's label and keep its empty value. But it isn't possible to tie actions to the built-in default option.

If I try something along the lines of ``"defaultselectedlabel": false`` to disable the built-in default option, fieldflow will show my first hardcoded option as the default (which makes sense). I can tie actions to that. But fieldflow won't allow me to set the ``query`` action's value to an empty string in my hardcoded option.

Truthy ``if`` conditions in fieldflow's logic will cause my empty value string to be treated as falsy and result in fieldflow re-using my label as the value. End result is that required field validation stops working properly since every option has a non-empty value. The non-empty string may also appear in form submission data if the fieldflow dropdown is set to optional (``"isoptional": true``).

This PR fixes that bug and in turn makes it possible to create a fully custom default option.

duboisp:
Thanks for the example @EricDunsworth I will add it and then we will merge this PR.

Would the "reset" configuration for fieldflow might fit your immediate need: https://wet-boew.github.io/GCWeb/components/wb-fieldflow/advanced-en.html#baseAction

duboisp:
Pre-approved after review and after I added the working example that illustrate the use case.

duboisp:
pre-approved after the working is added

duboisp:
I added the working example and re-tested and i confirm it work as expected.

Let's merge it once the cid script pass as this PR was pre-approved.

duboisp (approved):
Approved at the technical review.

PR #2068, opened 2022-12-07
Build: Adding dependencyChecker when running grunt

Garneauma:
Adding dependencyChecker task when running `grunt` (or `grunt debug`), `grunt dist` or `grunt méli-mélo`.  This should fix package update issues when rebasing or switching to a new branch.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -29,6 +29,8 @@ module.exports = (grunt) ->
 		"dist"
 		"Build distribution files ready for production"
 		[
+			"checkDependencies"
+			"check-wet-version"
 			"test"
 			"jekyll-theme"
 			"core-dist-PROD"
@@ -41,6 +43,8 @@ module.exports = (grunt) ->
 		"debug"
 		"Build a local working copy"
 		[
+			"checkDependencies"
+			"check-wet-version"
 			"jekyll-theme"
 			"jekyll-theme-runLocal"
 			"core-dist-DEBUG"
@@ -52,6 +56,8 @@ module.exports = (grunt) ->
 		"méli-mélo"
 		"Build méli-mélo files and run it in-place"
 		[
+			"checkDependencies"
+			"check-wet-version"
 			"jekyll-theme"
 			"jekyll-theme-runLocal"
 			"clean:mélimélo"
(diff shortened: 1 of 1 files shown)


duboisp (changes requested):
I tested, and it don't seem to detect our wet-boew patches neither minors changes neither any change when we update the wet-boew version in GCWeb. 

I also tried with the following 2 options for checkDependencies in our Gruntfile.coffee but it did nothing
* checkGitUrls: true
* checkCustomPackageNames: true

My test was the following:

**Test 1 - Check if detect patch changes**
I did:
1. git fetch --all
2. git reset --hard v12.2.0
3. npm install
4. git stash save
5. git stash drop
6. 5. git rebase v12.4.0
7. git cherry-pick c748893
8. grunt checkDependencies

Result expected: Step 6 (grunt) fail because locally I have an old version of wet-boew installed v4.0.56 which should be expected to be v4.0.56.3

Result obtained:
* The grunt was working without requiring to update my local version of wet-boew 
* I tried "grunt checkDependencies" with the config options changes above but the grunt build didn't failed


**Test 2 - Check if detect minor changes**
I did:
1. git fetch --all
2. git reset --hard v12.1.0
3. npm install
4. git stash save
5. git stash drop
6. git rebase upstream/master
7. git cherry-pick c748893
9. grunt checkDependencies

Result expected: Step 6 (grunt) fail because locally I have an old version of wet-boew installed v4.0.55 which should be expected to be v4.0.56.3 or greater

Result obtained:
* The grunt was working without requiring to update my local version of wet-boew 
* I tried "grunt checkDependencies" with the config options changes above but the grunt build didn't failed

---

I am not sure if it is because we are using a Github URL or if we should publish wet-boew on npm or both or something else.

The goal with this PR is to have a automated way to remain us to update our local version of wet-boew when building with grunt. If the grunt action "checkDependencies" don't work, let explore other options like writing our special dedicated function to meet our need.

FYI - We can know the installed dependencies by looking at the value of the property "_from" in this local file "node_modules/wet-boew/package.json" which must match the value of the property "dependencies/wet-boew" in this local file "package.json".

duboisp:

duboisp:
I tested locally and it works as expected. It detect when my wet-boew version is outdated and when it match

duboisp (approved):

PR #2069, opened 2022-12-08
DoAction: Ability to search on the whole table

Garneauma:
Adding the ability to filter on whole table when the column property is not defined. If a column if not defined, the value will be added to the datatables search box (default datatables behaviour).

Changes related to WET-297


duboisp:
Pre-approved upon successful local testing

duboisp (dismissed):
Reviewed and tested the code locally, work as expected.

I am going to add the working example I used for testing.

duboisp (approved):
Reviewed the code and tested locally. It all work as expected

PR #2071, opened 2022-12-12
Footer: customizable transparency links

Garneauma:
Transparency links in sub-footer band can now be customized either globally or on a per-page basis. If the links are neither defined on the page or globally, they will default to the Canada.ca transparency URLs.

### DIFF
--- a/sites/footers/includes/footer.html
+++ b/sites/footers/includes/footer.html
@@ -14,6 +14,26 @@
 	{%- endif -%}
 {%- endcapture -%}
 
+{%- capture termsUrl -%}
+	{%- if page.termsUrl -%}
+		{{ page.termsUrl }}
+	{%- elsif site.global.termsUrl[i18nText-lang] -%}
+		{{ site.global.termsUrl[i18nText-lang] }}
+	{%- else -%}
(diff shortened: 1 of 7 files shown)


sites/footers/includes/footer.html
```diff
@@ -14,6 +14,34 @@
...
+{%- capture termsUrl -%}
+	{%- if page.termsUrl -%}
+		{{ page.termsUrl }}
+	{%- elsif site.global.termsUrl[i18nText-lang] -%}
+		{{ site.global.termsUrl[i18nText-lang] }}
+	{%- else -%}
+		{%- if i18nText-lang == "fr" -%}
+		https://www.canada.ca/fr/transparence/avis.html
+		{%- elsif i18nText-lang == "en" -%}
+		https://www.canada.ca/en/transparency/terms.html
```
duboisp, line 26: Est-ce que tu peut mettre le default dans le i18n liquid include

duboisp (changes requested):
Pre-approved upon successful testing and after the default are moved into our centralized i18n liquid setting include.

duboisp (approved):
Reviewed the code and tested locally, work as expected

PR #2077, opened 2022-12-30
Maintenance - Removal of the 'misc' folder

duboisp:
To improve and reduce ambiguity in our code base IA by removing the "misc" folder.

This change is a **patch** according to the latest draft of our public versioning API as defined by the [Design Decision 14](https://wet-boew.github.io/wet-boew-documentation/decision/14.html).

**Technical change in this PR:** Only move the existing CSS declaration into the component they belong to. There is no functionality removal and no addition.

cc @delisma


Ricokola (approved):
Tested localy and seems to work as expected - PASS

PR #2086, opened 2023-01-18
Build warnings: Fixing advanced service pages

Garneauma:
Resolving build warnings part 3:
- Adding service side panel to all advanced service pages (and service page)
- Fixing next and previous buttons in advanced service pages navigation
- Fixing subtitle in advanced service pages


Ricokola:

duboisp (changes requested):
The files in the "includes" folder look like to be more a code sample rather than an reusable/configurable include file when compared to all the other includes we do have in gcweb-jekyll.

* [ ] Rename the folder "includes" for "samples" (+ do reference adjustment).

I did a local testing and reviewed all the other change and it looks good to go after the folder renaming. I confirm this change are only a Patch that fix the old service initiation reference implementation template.

Garneauma:
> * [x]  Rename the folder "includes" for "samples" (+ do reference adjustment).


Done ! :)

duboisp (approved):
Completed the review and local testing

PR #2090, opened 2023-01-20
Build warnings: contact us, corp info, supportedby

Garneauma:
Resolving build warnings part 7:
- Adding institutional information include file
- Adding organizational information include file
- Fixing reference to institutional info include file
- Fixing reference to organizational info include file
- Integrating "supported-by" include file content into files calling it


Ricokola:

templates/institutional/includes/inst-info.html
```diff
@@ -0,0 +1,71 @@
...
+
+++++++++++++++++++++++++++++++++++
++ List of institutional links and head of institution image
++
++ Used by institutional pages
+++++++++++++++++++++++++++++++++++
+
+{%- endcomment -%}
+
+<section class="gc-crprt">
```
duboisp, line 11: Would this code sample be in the "components" under "gc-crprt". It look very similar to the other code sample in organisation template.

I suggest to put both content sample under "components/gc-crprt/samples/" folder

templates/organizational/includes/org-info.html
```diff
@@ -0,0 +1,63 @@
...
+
+++++++++++++++++++++++++++++++++++
++ List of organization links and head of organization image
++
++ Used by organizational pages
+++++++++++++++++++++++++++++++++++
+
+{%- endcomment -%}
+
+<section class="gc-crprt">
```
duboisp, line 11: Would this code sample be in the "components" under "gc-crprt"

templates/localnav/index-en.html
```diff
@@ -48,7 +48,14 @@ <h3 class="h5"><a href="./task4/index-en.html">[Task 4]</a></h3>
 		</div>
 	</section>
 	<div class="pull-right col-xs-12 col-md-4">
-		{{>supported-by}}
+		<section class="lnkbx">
```
duboisp, line 51: Comment: `lnkbx` CSS seem to be an utility component by itself

duboisp (changes requested):
* [ ] Code sample should be in the "samples" folder instead of "includes"
* [ ] Move `gc-crprt` as a components outside templates

Garneauma:
> * [x]  Code sample should be in the "samples" folder instead of "includes"
> * [x]  Move `gc-crprt` as a components outside templates

Done ! :)

duboisp (approved):

PR #2092, opened 2023-01-23
Release v12.6.1

polmih:
Release GCWeb v12.6.1


duboisp (commented):
Finally, we didn't merged a change that directly impacted our GCWeb versioning API. We are going to wait until next week.

duboisp (approved):

PR #2093, opened 2023-01-24
Theme: Feedback - Jekyll, No ajax when link are undefined

duboisp:
Test, when there is no site config or when the page feedback property are set to an empty string.


duboisp:
Pre-approved upon successful testing and code review

Garneauma (approved):
Fix is working as expected.

However, the template seemed to be working the same without this change. The only difference is that, before, the HTML showed an empty `data-ajax-replace` attribute (which did not cause any error).

PR #2094, opened 2023-01-27
gc-servinfo: Adding reference to base scss file

Garneauma:
components/gc-servinfo/_base.scss was missing from theme.scss

Changes related to WET-305

### DIFF
--- a/components/gc-servinfo/_base.scss
+++ b/components/gc-servinfo/_base.scss
@@ -1,6 +1,6 @@
 /*
  *
- * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
Other than the encoding issue, it's all good. I reviewed and tested locally and it work as expected.

duboisp (approved):
Approved, it work as expected and I am going to make another PR to fix the UTF-8 caracter encoding issue considering this only impact a non-rendered SCSS comment.

duboisp:
Regarding the file with the UTF-8 encoding issue, it seems that GitHub auto-fixed the issue. When I fetch the repository and manually check, there is no encoding issue. Neither in the GitHub source code view here: https://github.com/wet-boew/GCWeb/blob/master/components/gc-servinfo/_base.scss

So no further action is required

PR #2101, opened 2023-02-15
Ministerial Profile - Fixing minor typos

Ricokola:
Content changes only


duboisp (approved):

Patch - Content change

PR #2102, opened 2023-02-15
index page : fix typo in meaning of status

ouafaaetta:
change as to at


duboisp (approved):

Patch - Content change

PR #2104, opened 2023-02-17
header: Fix broken links in docs

ouafaaetta:
en and fr


duboisp (approved):

PR #2108, opened 2023-02-22
index: switch from javascript  to CSS

ouafaaetta:
switch to CSS like the en index page


duboisp:
Pre-approved upon successful local testing.

duboisp (approved):
Tested locally and work as expected.

PR #2098, opened 2023-02-10
Feedback: implementing Page Feedback Tool

Garneauma:
- Implementing Page feedback tool with JSON Manager
- Updating Report a problem to use JSON Manager
- Implementing Iteration feature for variants for both

Changes related to WET-219 and WET-249

PR #2110, opened 2023-02-28
Feedback: Adding JSON Manager to RAP

Garneauma:
- Feedback: Adding JSON Manager Extractor and Data-JSON to RAP to generate hidden inputs automatically.
- Page details: switching `<footer>` to `<section>` element.

Changes related to WET-249.

### DIFF
--- a/sites/feedback/ajax/report-problem-en.html
+++ b/sites/feedback/ajax/report-problem-en.html
@@ -1,78 +1,93 @@
-<div class="row row-no-gutters"><div class="col-sm-9 col-md-6 col-lg-5">
-<details class="brdr-0">
-	<summary class="btn btn-default text-center">Report a problem on this page</summary>
-	<div class="well row">
-		<div class="gc-rprt-prblm">
-			<div class="gc-rprt-prblm-frm gc-rprt-prblm-tggl">
-				<form action="/sites/feedback/feedback-form-destination.html" id="gc-rprt-prblm-form" class="wb-postback" data-wb-postback="{&quot;success&quot;:&quot;.success-message&quot;,&quot;failure&quot;:&quot;.failure-message&quot;}" >
-					<!--
-
-					Insert the form submission configuration here
-
-					-->
-					<input type="hidden" name="externalReferer" value="">
-					<input type="hidden" name="pageTitle" value=""> <!-- Copy of the title of the page -->
-					<input type="hidden" name="submissionPage" value=""> <!-- Copy of the page URL -->
-					<input type="hidden" name="lang" value="en">
-					<input type="hidden" name="pageOwner" value="gc:institutions/"> <!-- Department name from canada.ca controlled vocabulary -->
-					<input name="subject" type="hidden" value="Report a problem or mistake on this page">
-					<fieldset>
-						<legend>
-							<span class="field-name">Please select all that apply:</span>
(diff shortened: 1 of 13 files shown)


duboisp:
pre-approve upon review and testing

sites/feedback/ajax/report-problem-en.html
```diff
@@ -1,78 +1,93 @@
...
+								"success": ".success-message",
+								"failure": ".failure-message"}'
+							data-wb-jsonmanager='{
+								"name": "rap",
+								"extractor": [
+									{ "interface": "referer", "path": "externalReferer" },
+									{ "selector": "title", "path": "pageTitle" },
+									{ "interface": "locationHref", "path": "submissionPage" },
+									{ "selector": "html", "attr": "lang", "path": "lang" },
+									{ "selector": "meta[name=\"author\"]", "attr": "content", "path": "pageOwner" },
```
duboisp, line 19: Let's use the dublin core metadata corresponding value of `dcterms.creator` as that is the metadata value required according to our [TBS metadata standard in the Appendix C](https://www.tbs-sct.canada.ca/pol/doc-eng.aspx?id=18909)

For the meta value, should we support official alternative according to HTML5 spec? like `dc.creator`, `author`. Except the JSON-LD or/and RDFa equivalent as it might require an improved data extractor.

duboisp (changes requested):
I reviewed the code and tested locally, All the content update are good idem the code is functional as expected except:

* [ ] Use the primary value associated with `dcterms.creator` instead of `author`. As specified by the [Web Resource Discovery Metadata Requirements](https://www.tbs-sct.canada.ca/pol/doc-eng.aspx?id=18909)

For discussion:

* [ ] Should we support metadata alternative to obtain the value, like a fallback mechanism. For example, let grab `meta[name=dcterms.creator]`, then `meta[name=dc.creator]`, then `meta[name=author]`, then `meta[property=dcterms:creator]`, then `meta[property=dc:creator]`, then `meta[property=dc11:creator]`.


Note (Proposal not required for this PR scope):

* [ ] Add something at the top of the documentation that show there is 2 sub component. One is for the integration into the page template and another one which is the design pattern of the inserted (ajaxed) form.

Garneauma:
> * [x]  Use the primary value associated with `dcterms.creator` instead of `author`. As specified by the [Web Resource Discovery Metadata Requirements](https://www.tbs-sct.canada.ca/pol/doc-eng.aspx?id=18909)

Changed the meta target to "dcterms.creator".

> * [ ]  Should we support metadata alternative to obtain the value, like a fallback mechanism. For example, let grab `meta[name=dcterms.creator]`, then `meta[name=dc.creator]`, then `meta[name=author]`, then `meta[property=dcterms:creator]`, then `meta[property=dc:creator]`, then `meta[property=dc11:creator]`.

I agree it would be nice to have a fallback. This would require enhancements to the "JSON Manager" plugin. Let's discuss this at our next team meeting.

> * [ ]  Add something at the top of the documentation that show there is 2 sub component. One is for the integration into the page template and another one which is the design pattern of the inserted (ajaxed) form.

This can be done in PR #2098 as it completely rewrites the Page Feedback documentation.

duboisp (approved):
Reviewed and tested.

Work as expected

PR #2124, opened 2023-03-09
wcag upgrade- scaffolding -add feature link assessment

ralph82:


duboisp:
There is a JSON validation issue in the ARC, see: https://github.com/wet-boew/GCWeb/actions/runs/4377200423/jobs/7660359116#step:7:13

PR #2126, opened 2023-03-10
GCWeb menu documentation - Fixing broken links

Ricokola:
Content change - Fixing broken links


duboisp (approved):
Review and confirm this is only content change

PR #2127, opened 2023-03-10
wcag upgrade- scaffolding accessibility assessment

ralph82:


duboisp:
This report need to be submitted into https://github.com/wet-boew/wet-boew repository in this folder: /src/plugins/footnotes/reports/

PR #2125, opened 2023-03-09
wcag upgrade- scaffolding -add text-level assessment

ralph82:

PR #2130, opened 2023-03-11
wcag upgrade- scaffolding -add text-fr-level assessment

ralph82:


duboisp (changes requested):
This PR content have no indicator to mark the content as being in French, as suggested by the PR title. There is no language tag neither a mention saying it need to be translated.

duboisp:
This PR seem to be a duplicate of PR #2125 but I think the intend is to provide a French version of the propose change in #2125

Please review and if applicable move this change request in the PR #2125

duboisp:
Change was merged via #2321

PR #2131, opened 2023-03-11
wcag upgrade- scaffolding -add group-fr-level assessment

ralph82:


duboisp (changes requested):
In a few location inside the HTML the markup need to be updated according to the success criterion 1.3.1, please review and apply the required changes.

I don't think we do need to use the line break (`<br>`) element everywhere
Please use the proper markup to identify when there is a fragment of computer code in the content (`<code>` element).
We can not have flow content inside paragraph. For example this is invalid: `<p><h3>` and should be instead `<h3>`

PR #2134, opened 2023-03-22
Templates: Fix flawed H1 heading setups

EricDunsworth:
Several page templates had flawed H1 heading implementations. Probably mostly-caused by handlebars leftovers.

Specifically:
* Some templates showed duplicate H1s:
  * Caused by using both a hardcoded H1 and a default Jekyll layout (which already comes with a built-in H1)
* Some templates showed one H1:
  * Using a standard-looking hardcoded H1 (not in a grid) and the "without-h1" Jekyll layout... i.e. unnecessarily-complicated setup

This resolves it by:
* Removing hardcoded standard-looking H1s from pages that use the default Jekyll layout
* Adding the "without-h1" Jekyll layout to pages that use a hardcoded H1 in a grid
* Removing both hardcoded H1s and the "without-h1" layout from pages that don't actually need them

Note:
* Didn't update modified dates since these changes just restore intended H1 presentations. Plus the dates weren't changed when the presentations broke in the first place.

### DIFF
--- a/templates/institutional/institution-arms-en.html
+++ b/templates/institutional/institution-arms-en.html
@@ -1,7 +1,6 @@
 ---
 {
 	"title": "[Institution Name]",
-	"layout": "without-h1",
 	"language": "en",
 	"altLangPage": "institution-arms-fr.html",
 	"pageType": "institution",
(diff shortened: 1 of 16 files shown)


duboisp (approved):
Reviewed and tested locally,

This is only a content change and work as expected.

PR #2149, opened 2023-04-18
WCAG upgrade - Create test page for: Color (foreground/background)

BrahimMahadi:
This pull request includes a test page with working examples filled with lorem ipsum text for Utilities/helpers. More info can be found by referencing WET-319.


sites/footers/reports/a11y-1-en.hbs
duboisp, line 1: The propose change for this file need to be put in another PR (if that not already the case)

sites/footers/reports/a11y-1-fr.hbs
duboisp, line 1: The propose change for this file need to be put in another PR (if that not already the case)

sites/footers/reports/a11y-1.json
duboisp, line 1: The propose change for this file need to be put in another PR (if that not already the case)

duboisp (changes requested):
Here a quick review

* [ ] The site footer ACR would need to be removed from this PR. The easiest way to make this change will be to use a local git client to make the changes.
* [ ] Missing the French content equivalent page
* [ ] Missing the index.json-ld which will contains the metadata for this common component
* [ ] I quickly saw that the heading structure for the "colour.html" page need to be adjusted like at line 62 where it should be an h5 instead of being an h4

sites/footers/index.json-ld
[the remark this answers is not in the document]
BrahimMahadi: Removed footer files from this PR.

sites/footers/reports/a11y-1-en.hbs
BrahimMahadi: Removed footer files from this PR.

sites/footers/reports/a11y-1.json
BrahimMahadi: Removed footer files from this PR.

sites/footers/reports/a11y-1-fr.hbs
BrahimMahadi: Removed footer files from this PR.

PR #2156, opened 2023-05-02
Update for Canada Day 2023

nfrenette:


duboisp (approved):
I reviewed and new added feature are demoed and designed according to PE.

Although I found the working example page structure (heading sequence + associated content) may need to be reviewed. Considering that you are going to be at the office hours tomorrow, let fix that content during that scheduled time.
* line 131 missing an heading level 4 or move the code block before the row at line 113
* section at line 67-72 contains an h2 but it seem to be under the heading level 3. Considering this heading structure is inconsistent with the rest of that page structure, I recommend to re-organise that example and if needed to move it onto another separate page.

PR #2160, opened 2023-05-11
Stabilizing follow-us plugin

jmealing:
- Remove provisional classes
- Adjustment to margins in lists to have better aligned spacing and touch target areas
- Remove margin on h2 that is causing it to have a large gap on mobile

To do:
* [x] DTO to update html file for the Institutional Landing page to add a margin to make things align better
* [x] DTO to create guidance page 
* DTO EN: PR https://github.com/canada-ca/design-system/pull/308
* DTO FR: PR in progress


jmealing:
Unfortunately I can't get the local preview working on my devices, but I coded the page elsewhere to validate these edits should work to improve the design spacing

duboisp:
@Garneauma when you have time this week, can you build a site demo + link to where to download the compiled version in .zip.

Garneauma:
@jmealing Unfortunately, some of the changes you made are breaking the older pattern of the follow us plugin (`.followus`). Is the intent to stop supporting the older pattern?

Secondly, I think this stabilization will require us to create a documentation for the follow us plugin in order for us to test and track upcoming changes. Are you interested in creating this documentation or are you relying on us to create it? If you are relying on us, we will most likely create a separate PR that would include your few change suggestions regarding margins.

Finally, could you send the link to the guidance for this plugin?

Thank you!

jmealing:
@Garneauma The intention is to stop supporting the older pattern and replace it with this one

I'll discuss the other items with the DTO team and get back to you

Garneauma:
> @Garneauma Could you clarify what changes would be breaking between the new pattern and the old? They use two entirely different classes `gc-followus` vs `followus`

They indeed use a different class. However, they use the same pseudo-class to build the styles in SASS (`%social-media-icons-followus-logo-properties`). Since you removed "provisional" from the pseudo-class name at line 71, it now overwrites the pseudo-class defined at line 5 (which now has the exact same name). I have suggested changes to fix the issue.

Garneauma (changes requested):

Garneauma:
@jmealing The old pattern looks good now. Thank you.

I was comparing the output of GCWeb with your latest changes vs the guidance and noticed the spacing is a little different for the stacked version and very different for the inline version. Is this intentional?

GCWeb stacked:
Screen Shot 2023-12-07 at 12 56 10 PM

Guidance stacked:


GCWeb inline:
Screen Shot 2023-12-07 at 12 56 34 PM

Guidance inline:

jmealing:
@Garneauma You're right, there shouldn't be that much spacing between the heading and the icons for inline. I am not able to preview my edits reliably which makes it a challenge. It's about more than double what it should be.

The guidance draft image does need to be updated too. The slightly increased spacing between the individual icons themselves is correct, and a bit more space between header and icons than that pictured is intended.

Garneauma (dismissed):
@jmealing The following change should fix the issue.

Garneauma:
@jmealing This component will need a documentation page and working examples (for stacked and inline).

Are you willing to try the new data-first approach for the documentation? If so, you can start with the sites/date-modified files and edit it for gc-follow-us. You will only need the 2 HTML files and the JSON-LD file. Don't bother with the "changesets" and "iterations" properties as they can be a bit tricky. I can suggest the content for those once you are done.

If not, we can discuss another approach. Thanks!

jmealing:
@Garneauma guidance and examples have been created in the `sites` folder. Please let me know if we need additional changes. I flagged some items to send to translation

Garneauma:
@jmealing I don't see the guidance and examples in this PR. Did you add them in another PR ?

jmealing:
It's all in the` sites `folder. The guidance is there and the examples in the includes. 
Screenshot 2024-01-17 at 8 53 15 AM

Garneauma:
I'm sorry @jmealing but as far as this PR is concerned, all the files about gc-follow-us remain unchanged and under /components. The only change made on this PR is in /components/gc-follow-us/_base.scss.

jmealing:
Screenshot 2024-01-18 at 1 25 35 PM
@Garneauma I'm not sure what's going on, as I did edit and push the files successfully from what I can see. Any ideas @delisma ?

duboisp:

duboisp (dismissed):
Reviewed and tested, everything work as expected.

The following will need to be completed, we will do it after this is merged considering this is only documentation related
* [ ] Documents the to be deprecated asset "twitter.svg"
* [ ] Document the new asset "x.svg"

duboisp (approved):

Garneauma (approved):

PR #2169, opened 2023-05-31
apply h1 thickline globally

jmealing:
As per DTO email, this is the update for the H1 globally:

- Removed gc-thickline class and css
- Modified heading base to apply the thick line styles globally


duboisp (changes requested):
We will need to check all the proper documentation + working example is also there and up to date.

duboisp (commented):
Move the h1 CSS and create example in this base folder: /sites/h1

Like the breadcrumb folder structure: https://github.com/wet-boew/GCWeb/tree/master/sites/breadcrumbs

1. Create a single SCSS file for the h1 style
2. Remove the old h1 style (like you did)
3. Link that newly SCSS in this main SCSS file, near the other heading style - https://github.com/wet-boew/GCWeb/blob/master/sites/theme.scss#LL35C6-L35C6
4. Create the dedicated working example (EN and FR) for the h1 
5. Document in those page the version and a code sample
    a. I do think it could be version 1.0.1, a patch change considering there is not change required for the author neither the implementor. We would need to check the public versioning API to be sure. It could be seen as a Minor/Major visual change.
   b. Code sample: I do think we have a variant with the [id=wb-cont] and another one without it. And there is some variant enhance with RDFa + schema.org
6. Create the index.json-ld file (the name of the componentName need to be same as the folder name.

sites/baseline/_heading.scss
```diff
@@ -8,10 +8,15 @@
...
-	padding-bottom: .2em;
+	border-bottom: none;
+
+	&::after {
+		border-bottom: .18em solid $border-red;
+		content: "";
+		display: block;
+		padding-bottom: 8px;
+		width: 70px;
+	}
```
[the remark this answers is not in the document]
EricDunsworth: This alternate method also reduces the space below the H1 vs what the pseudo-element produces.

When the pseudo-element is used, the H1's .2em ``padding-bottom`` and .2em ``margin-bottom`` appear back-to-back below the thickline - culminating to .4em of bottom spacing.

By comparison, in the pre-existing underline style and the alternate thickline method, the H1's .2em ``padding-bottom`` appears **above** the underline/thickline and the .2em ``margin-bottom`` appears below it.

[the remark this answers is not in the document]
duboisp: Thanks for sharing, we will need to compare both variation to determine the pros and cons of each variant.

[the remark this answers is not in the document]
delisma: @jmealing What do you think of @EricDunsworth proposition? I do like the idea to not rely on a pseudo-element but I don't know if there was a rationale beforehand.

sites/h1/h1-en.md
```diff
@@ -0,0 +1,94 @@
...
+<h3>Guidance</h3>
+<ul>
+	<li><a href="https://design.canada.ca/styles/typography.html">Typography</a></li>
+</ul>
+
+<h3>Variants</h3>
+<p>This component supports only 1 variant:</p>
+<ul>
+	<li><a href="#variantDefault">H1 (default)</a></li>
+</ul>
```
duboisp, line 45: @delisma as we discussed the other day, I renamed this "component variant" concept for "Change set"

duboisp (changes requested):
I suggest to reorganize the documentation according to our [latest documentation template prototype](https://universallabs.org/wet/labs/global-header/sites/header/header-docs-en.html).

You can see how the "Change set" and "Iteration" are documented under the  [technical documentation](https://universallabs.org/wet/labs/global-header/sites/header/header-docs-en.html#technical-docs) section.

sites/baseline/_heading.scss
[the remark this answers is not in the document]
delisma: Yes! I was working on the documentation. I'll move to clean-up the CSS next week.

sites/main-page-title/includes/main-page-title.html
```diff
@@ -0,0 +1,12 @@
+{% if include %}
+	{% if include.dir == 'ltr' or include.dir == 'rtl' %}
+		<h1 property="name"
+		{%- if include.rdfa -%}
+		typeof="WebPageElement"
+		{%- endif -%}
+		{%- if include.id -%}
+		id="wb-cont"
+		{%- endif -%}
+		dir="{{ include.dir }}">{{ page.title }}</h1>
```
duboisp, line 10: We don't need to include the direction `dir`. That is usually/should be done on an higher level like the `<html>` element.

We should allow any general attribute and we should document what is essential or what is optional like it apply some customization.

sites/main-page-title/includes/main-page-title.html
```diff
@@ -0,0 +1,12 @@
+{% if include %}
+	{% if include.dir == 'ltr' or include.dir == 'rtl' %}
+		<h1 property="name"
+		{%- if include.rdfa -%}
+		typeof="WebPageElement"
```
duboisp, line 5: The h1 don't have a `typeof` attribute. That `WebPageElement` must be defined on a element that wrap the whole content, like the `<main>` element.

sites/main-page-title/includes/main-page-title.html
```diff
@@ -0,0 +1,12 @@
...
+		{%- endif -%}
+		{%- if include.id -%}
+		id="wb-cont"
```
duboisp, line 8: I think that include is exclusively for the first level 1 heading in the content and I don't think that include should be used when there is additional level 1 heading through the page.

```suggestion
		id="wb-cont"
```

sites/main-page-title/includes/main-page-title.html
```diff
@@ -0,0 +1,12 @@
+{% if include %}
```
duboisp, line 1: Is that needed? This include files are going to be copied inside our gcweb-jekyll `_include` folder.

```suggestion
```

sites/main-page-title/_main-page-title.scss
duboisp, line 1: We need to add this file in our main theme.scss file here: https://github.com/wet-boew/GCWeb/blob/master/sites/theme.scss in the "Base site" section

sites/main-page-title/main-page-title-en.md
```diff
@@ -0,0 +1,145 @@
...
+altLangPage: "main-page-title-fr.html"
+breadcrumbs:
+  - title: "GCWeb home"
+    link": "https://wet-boew.github.io/GCWeb/index-en.html"
+dateModified: "2023-07-24"
+description: "Documentation on the default H1 use as the main title of a page with short bold red underline."
+language: "en"
+pageclass: "wb-prettify all-pre"
+title: "Main page title"
+---
```
duboisp, line 11: You should use our special page documentation template and remove all the body content unless if they are additional information not included in our default template.

See: 
* https://github.com/wet-boew/GCWeb/blob/master/sites/feedback/feedback-docs-en.html
* https://github.com/wet-boew/GCWeb/blob/master/sites/feedback/feedback-docs-fr.html


```suggestion
---
altLangPage: "main-page-title-fr.html"
breadcrumbs:
  - title: "GCWeb home"
    link": "https://wet-boew.github.io/GCWeb/index-en.html"
dateModified: "2023-07-24"
description: "Documentation on the default H1 use as the main title of a page with short bold red underline."
language: "en"
pageclass: "wb-prettify all-pre"
title: "Main page title"
layout: documentation
index_json: index.json-ld
---
```

sites/main-page-title/index.json-ld
duboisp, line 1: At the first run with our page template, there is an issue with this file. May be missing a property or something. I am going to look at that and propose the change via a subsequent review.

duboisp (changes requested):

delisma:
@duboisp Can you remove me and Eric as reviewers? Can you remove Jennifer as an assignee?

sites/main-page-title/index.json-ld
duboisp: Finally I found the issue is related to the documentation template.

The documentation template try to map the value of the following 2 properties in a non-existing template. That generate the error where the element is null.
The problematic JSON property are:
* /implementation/[]/status
* /variation/[]/revisedFor
The problematic HTML CSS selector which are defined in our data-json configuration but are pointing nowhere are:
* [data-status-stable]
* [data-iteration]
And this following swap do nothing, it silently fail
* { "op": "wb-swap", "path": "/revisedFor", "ref": "/iteration" },

@Garneauma FYI - We will need to look at those issue eventually and update our documentation template accordingly.

sites/_variables.scss
```diff
@@ -14,6 +14,11 @@ $search-gray: #e0e0e0;
 $profile-gray: #eaebed;
 $gray-light: #6f6f6f; // Light grey colour (Bootstrap override).
 
+// Brand variables
```
duboisp, line 17: Those brand variable seems to be specific for the main page title. It might be better to change the comment to said for what component those value are created for. 

```suggestion
// Main page title - Brand variables
```

sites/_variables.scss
```diff
@@ -14,6 +14,11 @@ $search-gray: #e0e0e0;
...
+$rule-width: 71px;
+$rule-thickness: .18em;
+$rule-color: $border-red;
```
duboisp, line 20: Their technical SASS name can be adjusted accordingly too. I don't this following rule is only the default border width related to the first h1 and nothing else. And that width is only applicable to the border not to the h1 content.

```suggestion
$rule-sites-mainh1-border-width: 71px;
$rule-sites-mainh1-border-thickness: .18em;
$rule-sites-mainh1-border-color: $border-red;
```

_data/sites.json
duboisp, line 1: Not an issue, this file would need to be regenerated. We can do that after the merge

duboisp (changes requested):
We should clean up, which can be done on or immediately after, remove any use of "gc-thickline" in our pages.

Do we have a note about that somewhere saying that CSS class is useless because it is now the default for h1?

sites/main-page-title/main-page-title-en.md
```diff
@@ -0,0 +1,13 @@
...
+altLangPage: "main-page-title-fr.html"
+breadcrumbs:
+  - title: "GCWeb home"
+    link": "https://wet-boew.github.io/GCWeb/index-en.html"
```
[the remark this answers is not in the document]
delisma: True, but I noticed that it links back to the splash page instead of the home page.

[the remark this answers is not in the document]
duboisp: It should not, we do have two different breadcrumb config and those link are prepend with the site base path which is specially adjusted for Github or for when running the Jekyll site locally via Docker.

FYI - 

https://github.com/wet-boew/GCWeb/blob/master/_config.yml#L36-L42

```yml
  breadcrumbs:
    en:
      - link: "index-en.html"
        title: "GCWeb"
    fr:
      - link: "index-fr.html"
        title: "GCWeb"
```

https://github.com/wet-boew/GCWeb/blob/master/sites/breadcrumbs/includes/breadcrumbs.html#L17

```html
<li property="itemListElement" typeof="ListItem">
	<a property="item" typeof="WebPage" href="{% unless breadcrumb.link contains "http" %}{{ setting-siteBasePath }}{% endunless %}{{ breadcrumb.link }}">
		<span property="name">{{ breadcrumb.title }}</span>
	</a>
	<meta property="position" content="{{ breadcrumbCounter }}">
</li>
```

GormFrank:
Pre-approved for this week provided @delisma makes the changes requested by @duboisp, which are small changes that don't affect the core purpose of this PR.

sites/main-page-title/includes/main-page-title.html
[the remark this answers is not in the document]
delisma: Not yet. Let's wait to see what kind of logic we would need for the "_stacked h1s_" first.

_data/sites.json
delisma: I have removed it from the PR

sites/_variables.scss
delisma: Done

sites/main-page-title/main-page-title-en.md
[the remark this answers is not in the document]
delisma: Done

sites/main-page-title/main-page-title.json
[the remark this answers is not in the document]
delisma: Removed

components/baseline/_base.scss
```diff
@@ -7,19 +7,6 @@
 	background-color: #355688;
 }
 
-/* H1 short bold red underline */
-h1.gc-thickline {
```
duboisp, line 11: Comment/Note: We will need a technical note about that CSS class being obsoleted because its effect are now the new default. And that it is the new default, that is not a breaking change because any content that was specifying the `gc-thickline` will continue to appear and be rendered as it was before.

Not a requirement to get this PR merge, but that note must be added to GCWeb before the subsequent release. May be the best location for that note will be in the "index.json-ld" of the new "main-page-title" site component.

duboisp (approved):
Review and tested. It does work as expected.

Once this PR is merged, the following 2 actions is required immediatly after or they can be included with the release PR.
* [ ] Add the technical note about the obsoleted `gc-thickline` in the index.json-ld
* [ ] Regenerate the GCWeb sites assets (ex: _data/sites.json ...) to keep GCWeb web site functional.

duboisp:
Pre-approved upon quick review of the unresolved conversation by @duboisp

PR #2172, opened 2023-06-13
Build - Update dockerfile with ruby version 2.7.3

duboisp:


duboisp:
@delisma @jmealing FYI - Here the fix to run the site locally via Docker

GormFrank (approved):

PR #2181, opened 2023-06-30
Step by step navigation: adding documentation

Garneauma:
Adding .gc-stp-stp documentation and example.

### DIFF
--- a/components/gc-stp-stp/gc-stp-stp-en.html
+++ b/components/gc-stp-stp/gc-stp-stp-en.html
@@ -1,25 +1,20 @@
 ---
 {
-	"title": "[Service name] - 1. [Step / section page name]",
-	"secondTitle": "1. [Step / section page name]",
+	"title": "[Service name] - 2. [Step / section page name]",
+	"secondTitle": "2. [Step / section page name]",
 	"language": "en",
-	"altLangPage": "index-fr.html",
+	"altLangPage": "gc-stp-stp-fr.html",
 	"breadcrumbs": [
 		{ "title": "[Institution name]", "link": "#" },
 		{ "title": "[Topic name]", "link": "#" },
 		{ "title": "[Service name]", "link": "#" }
 	],
 	"secondlevel": false,
-	"dateModified": "2017-04-07",
+	"dateModified": "2023-09-05",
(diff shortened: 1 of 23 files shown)


components/gc-stp-stp/gc-stp-stp-en.html
```diff
@@ -0,0 +1,63 @@
...
+		<ul class="toc lst-spcd col-md-12">
+			<li class="col-md-4 col-sm-6"><a class="list-group-item" href="index-en.html">1. [Step / section page name]</a></li>
+			<li class="col-md-4 col-sm-6"><a class="list-group-item active">2. [Step / section page name]</a></li>
+			<li class="col-md-4 col-sm-6 clr-lft-sm"><a class="list-group-item" href="page3-en.html">3. [Step / section page name]</a></li>
+			<li class="col-md-4 col-sm-6 clr-lft-md clr-lft-lg"><a class="list-group-item" href="page4-en.html">4. [Step / section page name]</a></li>
+			<li class="col-md-4 col-sm-6 clr-lft-sm"><a class="list-group-item" href="page5-en.html">5. [Step / section page name]</a></li>
+			<li class="col-md-4 col-sm-6"><a class="list-group-item" href="page6-en.html">6. [Step / section page name]</a></li>
+		</ul>
+	</div>
+</div>
```
duboisp, line 30: This don't like the version 2 of the component.

This is not a "toc" (table of content) if we compare on how the "toc" CSS is currently used for the in-page table of content component.

duboisp (changes requested):
Would you like to create a include for this component and to update the "advanced service" template to use that include?

And I can see there is new CSS added for this component, is that necessary or probably a minor/patch change to the component?

duboisp:
Pre-approved upon review and local testing.

To be confirm: The CSS code for the download link might need to be removed as it was already moved I think.

Garneauma:
Removed the .gc-dwnld CSS. It was added by mistake during merge conflict resolution.

duboisp (approved):
Review and tested locally,

Work as expected. 

This is a minor change because it add another supported pattern for the gc step step which is the same as the version 1 but simpler to code for an author.

PR #2187, opened 2023-07-06
pre-assesment gc-freatures

ipaksc:


components/gc-features/reports/a11y-1-fr.html
```diff
@@ -0,0 +1,16 @@
...
+	"language": "fr",
+	"description": "Partielle. Pré-évaluation des critères de succès aux niveaux A et AA du WCAG 2.1",
+	"tag": "gc-features",
+	"parentdir": "gc-features",
+	"parentPage": "Promotions contextuelles",
+	"parentPageURL": "gc-features",
+	"altLangPage": "a11y-1-en.html",
+	"dateModified": "2023-07-04",
+	"layout": "assessment_wrote_en-fr",
+	"reportURL": "a11y-1.json"
```
duboisp, line 13: The name need to be adjusted.

```suggestion
	"reportURL": "pre-a11y-1.json"
```

duboisp (changes requested):
I reviewed and tested locally.

Other than the little change, everything look good once that change is completed.

components/gc-features/reports/a11y-1-fr.html
ipaksc: Fixed

duboisp (approved):
Reviewed and tested locally. Work as expected.

Thanks @ipaksc for the pre-assessment and the assessment

PR #2188, opened 2023-07-06
Supports: Adding CSS feature support detection architecture

Garneauma:
- Adding the ability to define CSS feature detection files
- Adding ":has(*)" selector support detection (to be removed once Firefox fully supports ":has(*)" for at least 2 versions)

Dependent on https://github.com/wet-boew/wet-boew/pull/9595 being released.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -183,6 +183,7 @@ module.exports = (grunt) ->
 		[
 			"clean:dist"
 			"sass:all"
+			"concat:supports"
 			"concat:plugins"
 			"copy:assets"
 			"copy:fonts"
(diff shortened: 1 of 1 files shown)


Gruntfile.coffee
```diff
@@ -507,6 +508,14 @@ module.exports = (grunt) ->
 					separator: ","
 				src: "sites/**/index.json-ld"
 				dest: "_data/sites.json"
+			supports:
+				options:
+					stripBanners: false
+				src: [
+					"<%= themeDist %>/css/theme.css"
+					"sites/supports/*.css"
```
duboisp, line 516: You can probably get it from the "node_modules/wet-boew"

duboisp (changes requested):

duboisp (approved):
Reviewed and tested locally the change. it work as expected.

This change goes along with the related build script change that happen to tag filter which leverage the :has css selector.

PR #2191, opened 2023-07-10
Provisional alert include: Simplify link destination

EricDunsworth:
* Use the ``relative_url`` Liquid filter for the provisional page's link destination (makes it relative to ``baseurl``)
* Remove the ``provisionaldir`` and ``siteroot`` variables
* Fixes broken links caused by incorrect relative paths
* Spinoff of #2179


EricDunsworth:
I haven't tested this, but the change is super straightforward and should work correctly. Something similar to this is already being used in GCWeb's breadcrumb include.

Liquid's ``relative_url`` filter uses ``site.baseurl`` (which resolves to ``/GCWeb`` on GitHub Pages) as its starting point. So the end result should be ``/GCWeb/components/index-en.html`` and  ``/GCWeb/components/index-fr.html``.

GormFrank:
Pre-approved upon succesful testing

duboisp:
Pre-approved upon local testing and gh-pages testing

Need to test the wet-boew-demos local example

duboisp (approved):
Reviewed, tested locally and tested in gh-pages. The link in the alert work as expected.

This is a content change

PR #2196, opened 2023-07-12
Theme: Remove hardcoded post-JS HTML markup

EricDunsworth:
* Remove the following hardcoded markup:
  * ``wb-init`` and ``wb-*-inited`` classes
  * ``wb-auto-*`` IDs
    * Note: Fixes broken local links caused by duplicate IDs
* Update modified dates of:
  * Pages containing affected code samples
  * Pages that reference updated samples/includes
* Related changes by @duboisp:
  * Revert several pages to the old incarnation of the follow-us include
  * Resurrect secondary menu examples/includes
* Spinoff of #2179


components/gc-follow-us/samples/follow.html
EricDunsworth, line 1: This file is also orphaned. It's used as the "source code" for the old follow-us include.

GCWeb's build system generates a copy of it in ``_includes/components/gc-follow-us/follow.html``. For some reason, the repo is setup to track both the sample and its compiled include :S.

GormFrank (dismissed):

_includes/components/gc-follow-us/follow.html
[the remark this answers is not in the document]
duboisp: This file is a clone of "components/gc-follow-us/follow.html" to allow to use it as a jekyll include. It is created by the build script.

components/gc-follow-us/samples/follow.html
duboisp: This file is cloned into "_includes/component/gc-follow-us/follow.html" to allow to use this sample code as a jekyll include in the working example. It is cloned by the build script.

But now that we do have the jekyll relative include, cloning it it might be not necessary.

templates/gc-audience-en.html
```diff
@@ -5,7 +5,7 @@
 	"altLangPage": "gc-audience-fr.html",
 	"pageType": "gcaudience",
 	"pageclass": "page-type-nav cnt-wdth-lmtd",
-	"dateModified": "2017-12-05",
+	"dateModified": "2023-07-12",
```
duboisp, line 8: Actually we didn't changed the template, we only fixed the broken Jekyll page. Idem for the other template.

_includes/components/gc-follow-us/follow.html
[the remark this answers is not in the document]
duboisp: We will need to check if we need to support both pattern, which is probably the current case.

The confusion is probably coming from the gc-follow-us documentation. The new method would try to complement the documentation by using/inserting the sample file instead of hard coding the source code in the documentation page.

components/gc-follow-us/samples/follow.html
duboisp: We can make that change. And maybe if we know a sample should be used outside its folder, like by another template, we can probably move their copies into an obvious folder name called like "cloned" or "duplicates" or "copies"... 

Note: the untracked folder "~jekyll-dist" is an workaround to run locally gcweb via Jekyll without any dependence and to avoid any interference during the development by the gcweb-jekyll theme.

duboisp (dismissed):
Knowing that you have found we updated by mistake the gc-followus pattern on page which shouldn't. I suggest to get that fixed before we forget about it. If I do have time today, I will add a new commit fixing this.

duboisp:
I am going to match the follow us pattern along with the current DTO documentation based on the wireframe which is currently published as today

By default all should use the old follow us include "gc-follow-us/follow.html"

Will use the new follow us "gc-follow-us/gc-follow-us.html"
* beta institution page
* beta theme/topic
* Intistitution contact page (I didn't found a wireframe showing the old follow us, so let keep it as is)

The template "localnav" look more as a variant of the topic page. Suggest to move it.

duboisp:
@EricDunsworth I fixed the follow-us include and added working example of secondary menu (by marking it deprecated)

Can you fix the merge conflicts? Feel free to replay or includes any changes that I made via the new commit.

_includes/components/gc-follow-us/follow.html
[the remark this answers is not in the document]
duboisp: I added a new commit where I did fixed the follow-us included sample file with the matching template, as per the wireframe found on DTO design website.

components/gc-follow-us/samples/follow.html
duboisp: This file is not orphaned anymore. It is used by the stable old template

polmih:
@EricDunsworth , wondering if the conflict with the master can be fixed? Thanks

EricDunsworth:
@polmih @duboisp I think this is ready to rock and roll now :).

Just did the following:
* Rebased/fixed merge conflicts
* Added a third commit to update modified dates
  * **Side note:** I decided against editing my initial commit for that... @duboisp's second commit expanded the scope of pages needing updated modified dates. So I felt like it'd of been inappropriate to retroactively add updated modified dates for those "extra" pages in the first commit.
* Updated the OP to cover changes from @duboisp's commit

duboisp:
@polmih ready for your review. Once reviewed and tested, this has been pre-approved during the technical review.

polmih (approved):
reviewed and locally tested.

templates/institutional/institution-arms-en.html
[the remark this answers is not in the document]
duboisp: Yes currently this the current stable style for the institutional profile page with Insignia. You can also see it from the wireframe after expanding the "Guidance for the stable institutional landing page" here: https://design.canada.ca/mandatory-templates/institutional-profile-pages.html

duboisp:
Thanks @polmih 

Change type:
GCWeb - Content - Adjusted the follow us component used on various page to match the stable example vs beta template, 
GCWeb - Content - Adding working example with secondary navigation (fixing the lowest-topic section example)
GCWeb Jekyll - Minor - Add basic support for content with a secondary menu

PR #2201, opened 2023-07-17
Utility: dl-inline

Garneauma:
Adding GCWeb example for newly added `.dl-inline` utility class.


duboisp (approved):
Reviewed and tested, work as expected

This is a content change

PR #2210, opened 2023-07-19
Paginate: porting paginate CSS to GCWeb

Garneauma:
Adding missing Paginate plugin CSS from WET to GCWeb.


duboisp (approved):
Tested locally and it work as expected.

PR #2216, opened 2023-07-25
Pre-accessibility assessment for panels

josephdiab:


duboisp (changes requested):
There is some issue with the index.json-ld + html file, but let fix them later.

For now, see the change requested in the pre-a11y-1.json file.

PR #2217, opened 2023-07-25
Pre-accessibility assessment for multimedia

josephdiab:


components/multimedia/reports/pre-a11y-1.json
```diff
@@ -0,0 +1,619 @@
...
+      "earl:mode": "earl:unknownMode",
+      "@type": [
+        "earl:TestResult",
+        "earl:Assertion"
+      ]
+    },
+    {
+      "earl:test": "WCAG21:three-flashes-or-below-threshold",
+      "earl:subject": "_:subject",
+      "earl:outcome": "earl:inapplicable",
```
duboisp, line 346: This is applicable. And even more when it concern an animation.

components/multimedia/reports/pre-a11y-1.json
```diff
@@ -0,0 +1,619 @@
...
+      "earl:mode": "earl:unknownMode",
+      "@type": [
+        "earl:TestResult",
+        "earl:Assertion"
+      ]
+    },
+    {
+      "earl:test": "WCAG21:on-input",
+      "earl:subject": "_:subject",
+      "earl:outcome": "earl:inapplicable",
```
duboisp, line 511: This is applicable because there is a few input control.

components/multimedia/reports/pre-a11y-1.json
```diff
@@ -0,0 +1,619 @@
...
+      "earl:mode": "earl:unknownMode",
+      "@type": [
+        "earl:TestResult",
+        "earl:Assertion"
+      ]
+    },
+    {
+      "earl:test": "WCAG21:error-identification",
+      "earl:subject": "_:subject",
+      "earl:outcome": "earl:inapplicable",
```
duboisp, line 544: Considering there is some input field. It will best to also evaluate this SC, at least do a few investigation to see if this is application and write our observation in the dct:description field.

components/multimedia/reports/pre-a11y-1.json
```diff
@@ -0,0 +1,619 @@
...
+      "earl:mode": "earl:unknownMode",
+      "@type": [
+        "earl:TestResult",
+        "earl:Assertion"
+      ]
+    },
+    {
+      "earl:test": "WCAG21:error-suggestion",
+      "earl:subject": "_:subject",
+      "earl:outcome": "earl:inapplicable",
```
duboisp, line 566: I don't think it is applicable, but we should just double check and note our observation in dct:description

duboisp (changes requested):
Note: This pre-assessment might move at this location: "/wet-boew/multimedia"

PR #2218, opened 2023-07-25
Pre Accessibility Assessment for overlay

josephdiab:
This was copied from wet-boew/src/plugins/overlay


components/overlay/reports/1-a11y-en.hbs
duboisp, line 1: There is no ".hbs" file in GCWeb. Those need to be renamed for ".html"

FYI - The site building software are different between wet-boew and gcweb github repository. GCWeb do use Jekyll + automated github pages and wet-boew use handlebar + manual gh-page.

components/overlay/reports/1-acr-wcag21.json
duboisp, line 1: Do not create the conformance report yet. That is something we are going to do later, after the majority of the  assessment would be done.

components/overlay/reports/1-a11y.json
```diff
@@ -0,0 +1,1416 @@
...
+      "foaf:name": "Hani Ahmed (Github: @heansih)",
+      "foaf:homepage": "https://github.com/heansih",
+      "@type": ["earl:Assertor", "foaf:Person"],
+      "foaf:member": {
+        "@id": "_:wave_browser_extension",
+        "doap:name": "WAVE web accessibility evaluation tool, browser extension",
+        "doap:homepage": "https://wave.webaim.org/extension/",
+        "@type": [ "earl:Assertor", "earl:Software" ]
+      }
+    }
```
duboisp, line 34: This should be updated with your name and github handle. It do identify who created this assessment report

```suggestion
      "foaf:name": "Your Name (Github: @yourGitHubHandle)",
      "foaf:homepage": "https://github.com/yourGitHubHandle",
      "@type": ["earl:Assertor", "foaf:Person"]
    }
```

components/overlay/reports/1-a11y.json
```diff
@@ -0,0 +1,1416 @@
...
+  },
+
+  "dct:date": "2023-01-06",
```
duboisp, line 37: This date must be updated to when you did the assessment. It can be in January 2023.

duboisp (changes requested):
Wait a minute, is this only the move of the overlay assessment from wet into GCWeb? It looks like it.

There is nothing that seems related to s pre-assessment of the overlay. And the move of the assessment + acr are going to happen only after the creation of a dedicated section for wet-boew.

@josephdiab let chat.

---
Note I was going to say:


Creating the pre-assessment should not have involved to add any css (scss) and js. neither to re-add the already existing working example.

Here the GCWeb working example of the overlay plugin:
* https://wet-boew.github.io/gcweb-compiled-demos/wetboew-demos/overlay/overlay-en.html
* https://wet-boew.github.io/gcweb-compiled-demos/wetboew-demos/overlay/overlay-fr.html

Remove the conformance report (file ending with -arc) because we would only do that at later date.

PR #2219, opened 2023-08-02
Update and-or.css

Christopher-O:
Updated CSS to allow brdr-0 to exist on all resolutions when the design is stacked, and not just "always stacked".

### DIFF
--- a/méli-mélo/2021-05-conjunction/edge-cases.html
+++ b/méli-mélo/2021-05-conjunction/edge-cases.html
@@ -16,8 +16,8 @@ css:
 		<li>Edge cases of And/Or decision points (Conjunctions)</li>
 	</ul>
 </nav>
-
-<h2>No border shall not work for responsive break point</h2>
+<p><span class="label label-info">Updated:</span> The design has been updated to allow brdr-0 on all resolutions and will be removed when they become stacked.</p>
+<s><h2>No border shall not work for responsive break point</h2>
 
 <p><strong>Test expectation:</strong> The border style <code>.brdr-0</code> is ignored when stack except for "Test 4".</p>
 
@@ -82,4 +82,4 @@ css:
 			<li class="cnjnctn-col">This is content for part B</li>
 		</ul>
 	</div>
-</section>
+</section></s>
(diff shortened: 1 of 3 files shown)


duboisp:
Pre-approved upon review and after confirming there is an example of this change.

duboisp:
pre-approved upon testing

Relates to 
* #2219 
* #2220 
* #2221

Christopher-O:
@duboisp there is an example of the update on index.html under the updated header of no border.

duboisp (approved):
Tested the CSS change which are demoed in PR 2220 and work as expected. All CSS is demoed as expected.

Todo:
* [ ] Merge all other PRs into a single one.

méli-mélo/2021-05-conjunction/edge-cases.html
```diff
@@ -16,8 +16,8 @@ <h2 class="h5">Table of content</h2>
 		<li>Edge cases of And/Or decision points (Conjunctions)</li>
 	</ul>
 </nav>
-
-<h2>No border shall not work for responsive break point</h2>
+<p><span class="label label-info">Updated:</span> The design has been updated to allow brdr-0 on all resolutions and will be removed when they become stacked.</p>
+<s><h2>No border shall not work for responsive break point</h2>
```
duboisp, line 20: The `<s>` element should be use in the phrasing context not as a flow content.

I suggest to instead put this content inside a details/summary with a title that inform the content are not long accurate. Or simply you can remove those outdated instruction.

What would you like to do?
* [ ] Replicate the `<s>` element for every phrasing context in the deprecated instruction block
* [ ] Move the content inside a details summary with a title saying it is deprecated instruction
* [ ] Remove the outdated instruction.

duboisp (changes requested):
Fix the HTML markup in the edge-cases.html. The [`<s>` element](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-s-element) can only be use in phrasing context according to HTML 5 spec.

Other than that, the rest is good to be squash and merge

méli-mélo/2021-05-conjunction/edge-cases.html
duboisp: I replicated the `<s>` on every sub element to ensure it is used as phrasing content.

duboisp (approved):
Tested and reviewed locally

PR #2221, opened 2023-08-02
Update edge-cases.html

Christopher-O:
Struck out text regarding brdr-0 not being available on side-by-side options when stacked, now is allowed and border will disappear once stacked


duboisp:
This PR change request was merged/moved with PR#2219

PR #2230, opened 2023-08-17
gc-features: removing 'text-decoration: none' on hover

Garneauma:
Re-applying the underline on hover of gc-features. New PR on WET applying box shadow to panels and wells containing a stretched-link.

Dependent on https://github.com/wet-boew/wet-boew/pull/9614 being merged.

### DIFF
--- a/components/gc-features/_base.scss
+++ b/components/gc-features/_base.scss
@@ -13,10 +13,6 @@
 
 	margin-bottom: $space-md;
 
-	a:hover {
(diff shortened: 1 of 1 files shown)


duboisp:
Pre-approved if PR #9614 is merged. And need to be in sync

duboisp (approved):
This is a patch.

There is going to be still a visual hover effect which would added by the `.stretchlink` used in panel and in a well.

PR #2231, opened 2023-08-18
Page feedback: fixing bad JSON array in report-problem include

Garneauma:
- Fixing bad JSON array issue in report-a-problem ajax files
- Removing lingering report-a-problem references in page-details


duboisp (approved):
Agree this change is a Patch change, that was a typo error.

PR #2232, opened 2023-08-21
Page feedback: moving ajaxed-in PFT file to /assets

Garneauma:
Related to WET-407


_config.yml
```diff
@@ -6,8 +6,8 @@ website: "https://wet-boew.github.io/GCWeb/" # URL of your public facing website
 global:
   lang: fr
   feedbackPath:
-    en: "/sites/feedback/ajax/page-feedback-en.html"
-    fr: "/sites/feedback/ajax/page-feedback-fr.html"
+    en: "/dist/GCWeb/assets/feedback/page-feedback-en.html"
+    fr: "/dist/GCWeb/assets/feedback/page-feedback-fr.html"
```
duboisp, line 10: We would need the absolute URL here, because it is broken in gh-page. 

Like: https://wet-boew.github.io/themes-dist/GCWeb/GCWeb/assets/feedback/page-feedback-en.html

duboisp (changes requested):

PR #2240, opened 2023-08-30
WET-409 Basic HTML WCAG Pre-assessment

jonasg-gc:


duboisp (commented):
@AtlakM for your action check
* The language links for the en & fr pages
* Ensure the report is populated as expected

duboisp (approved):
Thanks @AtlakM for the review

PR #2241, opened 2023-09-01
WET-347: Added relevant component discription

ouafaaetta:


duboisp (approved):
Reviewed, all good. This is only a content change

PR #2242, opened 2023-09-05
Content - Updating GCWeb web site indexes

duboisp:


Garneauma (approved):
Reviewed. This is only content change and should only affect index-en.md et index-fr.md.

PR #2244, opened 2023-09-08
Page feedback: hiding page feedback by default and updating doc

Garneauma:
- Updating Page feedback documentation to use new data-first layout.
- Page feedback is now hidden by default and can be manually added to every page
- Page feedback variation is now stable


duboisp:
Pre-approved upon local testing and review.

Note: This is a minor change for PFT because it change from provisional to stable while the RAP is still in a stable status

duboisp (changes requested):

duboisp (changes requested):
Reviewed and everything look good except 2 change request
* [ ] Remove the `<div>` wrapper workaround in includes files (feedback includes and deprecated rap-inline)
* [ ] Replace the bold by a strong element to have emphasis sementic

In summary:
* No issue with the gcweb-jekyll major change regarding the default behaviour of when the feedback is applied.
* No issue with the documentation provided, it is accurate
* The provisional alert was removed
* The component version has been increased by a minor version. The major will only occur when the RAP pattern would be officially deprecated.
* The change requested was addressed

The performance issue has been localized to the "content updated" function in the data-ajax plugin. We will do a deep dive and submit a fix to the data-ajax plugin next weeks.

Regarding the additional wrapping `<div>`, I am divided between removing them now or removing them next week. This is a bug in the data-ajax because it transform the received data as a jQuery object directly. The best solution will be to apply the "div" wrapper workaround directly in the data-ajax instead of altering the reference implementation design pattern. 
Let's remove the wrapper div here and submit a fix in wet-boew for next week.

Note:
We can do that in another content PR
* For "Iteration" show the 'breaking', 'additions' and 'fixes' details
* For Variations and implementations we can probably include the 'baseOn' iteration into a data-attribute. Eventually, I would like to use that value to set automatically the status on them.

duboisp:
FYI - here my code sample that seems to fix/apply a work around for the performance issue and fixing the div wrapper required to import content via data-ajax.

https://github.com/wet-boew/wet-boew/compare/master...duboisp:wet-boew:gcweb-pr2244

@Garneauma Let me know if you do conclude at a similar fix, along those line, for the performance issue and avoiding the wrapper div workaround.

duboisp (approved):
Reviewed and tested locally, work as expected.

Requested change was applied.

Know issue, which would be fix in the coming weeks if it is not by next week.
* Performance issue, we identified the cause and it is not caused by any change made in this PR. Par contre, cet example illustre très bien le problème de performance et elle pourra être utilisé afin de valider le correctif.
* Erreur du ajax-fetch car le contenu inséré et qui devrait être encodé n'est pas wrapper dans un div. Le PR avec ce correctif a déjà été soumis ou il le sera aujourd'hui.

PR #2251, opened 2023-09-13
GC Tables: Improve display for empty cells

GormFrank:
Fixed display issue when table cell is empty. Previously looked like this (with overlapping):
gc-table-before

Now looks like this with a new table row added for working example purposes:
gc-table-after

The `flow-root` property allows for an element to preallocate its height as-if there was always a text value in it, and is supported by all major browsers. I tested Safari, Chrome and Firefox and everything looked good. No side effect on the existing examples.

I kept the GC Tables as a Provisional feature.

### DIFF
--- a/components/gc-table/_screen-sm-max.scss
+++ b/components/gc-table/_screen-sm-max.scss
@@ -92,7 +92,7 @@ $gc-tables-tr-margin-vertical: .625em;
 			}
 		}
 		td {
-			display: block;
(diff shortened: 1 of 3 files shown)


duboisp:
Pre-approved upon review + testing

duboisp (approved):
Review and tested locally,

Work as expected, 

Note this is a provisional feature

PR #2254, opened 2023-09-20
Release v13.9.0

Ricokola:
Release v13.9.0


duboisp (changes requested):
It's look good,

We will only need to add the 2 new méli-mélo project that is going to be added once the CID script is completed

duboisp (approved):

PR #2255, opened 2023-09-20
Fonts: fixing font loading

Garneauma:
Fixing Lato font loading issue when cache is cleared or accessing site for the first time.

`font-display: fallback;` allows 3 seconds to load the font.


duboisp:
Pre-approved upon testing and review

duboisp (approved):
Review and tested locally, work as expected.

This a patch change where it do allow a short swap period (100 ms) for the font.

PR #2258, opened 2023-09-25
Feedback: updating doc to add migration from RAP to PFT

Garneauma:
Changes related to WET-416


duboisp:
Pre-approved - review + local testing.

duboisp (approved):
Review and tested locally. It look good.

This is only a content change

PR #2261, opened 2023-09-26
WET-347

ouafaaetta:


components/header-rwd/index.json-ld
```diff
@@ -7,14 +7,14 @@
 		"modified": "dct:modified"
 	},
 	"title": {
-		"en": "Well header responsive",
-		"fr": "Entête well responsive"
+		"en": "Context-specific feature",
+		"fr": "Liste - Style additionel"
```
duboisp, line 11: Ceci ne correspond pas au composant "well header responsive"

duboisp (changes requested):

components/header-rwd/index.json-ld
ouafaaetta: done

duboisp:
There is a merge conflict that need to be resolved

duboisp (dismissed):
Reviewed, the change looks good.

I will only update the _data/component.json generated file into this PR, so the change will be immediately effective.

duboisp (approved):

PR #2263, opened 2023-09-27
Page details: updating doc to new format

Garneauma:
- Updating doc to new data-first format
- Minor PFT changes


sites/page-details/index.json-ld
```diff
@@ -82,5 +83,346 @@
...
+		"Page feedback area",
+		"Share this page",
+		"Date modified"
+	],
+	"a11yGuidance": "no accessibility guidance",
+	"variations": [
+		{
+			"name": {
+				"en": "Page details with Page feedback tool",
+				"fr": "Détails de la page avec Outil de rétroaction sur la page"
```
duboisp, line 97: Not exactly accurate,

This variations seem to include 4 inner variation. It will be preferable to move those inner variation as normal variation.

sites/page-details/index.json-ld
duboisp, line 1: The page details is a container component, I think it will best if we do provide the generic container code first and then we can provide a example along with the inner component. Or find a way to add some emphasis the page details is a container and the inner component "pft" could evolve independently

sites/page-details/index.json-ld
```diff
@@ -82,5 +83,346 @@
...
+			},
+			"status": "stable",
+			"description": {
+				"en": "This component displays the Page details section in a <code>&lt;footer&gt;</code> tag. It contains the following components: Page feedback tool, Share this page, and Date modified.",
```
duboisp, line 101: This variation code sample are only showing an implement with the "section" not with the "footer". We probably need to re-write this description.

sites/page-details/index.json-ld
```diff
@@ -82,5 +83,346 @@
...
+				"Changed <footer> to <section> as <footer> element does not provide a semantic break."
+			],
+			"assets": [
+				{
+					"@type": "source-code",
+					"@language": "en",
+					"description": "Code sample",
+					"code": {
+						"@type": [ "rdf:HTML", "@id" ],
+						"@value": "samples/page-details-v3-en.html#full"
```
duboisp, line 306: The selector `#full` don't exist in the sample file `samples/page-details-v3-en.html`

sites/page-details/index.json-ld
```diff
@@ -82,5 +83,346 @@
...
+				"New invisible heading for accessibility."
+			],
+			"assets": [
+				{
+					"@type": "source-code",
+					"@language": "en",
+					"description": "Code sample",
+					"code": {
+						"@type": [ "rdf:HTML", "@id" ],
+						"@value": "deprecated/page-details-v2-en.html"
```
duboisp, line 342: This URL are broken, did you mean in the sample folder?

```suggestion
						"@value": "samples/page-details-v2-en.html"
```

sites/page-details/index.json-ld
```diff
@@ -82,5 +83,346 @@
...
+					"code": {
+						"@type": [ "rdf:HTML", "@id" ],
+						"@value": "deprecated/page-details-v1-fr.html"
+					}
+				}
+			]
+		}
+	],
+	"changesets": [
+		{
```
duboisp, line 385: We should add another change set for the version 1 of the page details which could be very briefly described. That way we would be able to announce it is deprecated.

duboisp (changes requested):
Incomplete review.

There is a few changes and notes. Let's discuss next week.

duboisp:
Pre-approved upon review + local testing

This is only a patch for GCWeb Jekyll, no impact on GCWeb versioning API, only content change for GCWeb.

duboisp (changes requested):
Incomplete review

sites/page-details/index.json-ld
```diff
@@ -32,55 +34,504 @@
...
+			"status": "stable",
+			"baseOnIteration": "_:iteration_pd_3",
+			"detectableBy": "section.pagedetails",
+			"layout": [
+				"Page feedback tool (if included) on the top left",
+				"Share this page button (if included) on the top right",
+				"Date modified on bottom left"
+			],
+			"style": "Not applicable",
+			"semantic": {
```
duboisp, line 483: As discussed,

Move the include in a new property like "GCWeb template | includes"
And redefine this value for something basic like "section, h2" + a mapping of the static data + dynamic data (schema) if applicable.

duboisp (changes requested):
Reviewed and tested locally,

Other than the small change, it will be good to merge once completed.

Side note: Regarding the property "assets" in the iteration, we might split it into 2 properties like "Code sample" + "Assets" but let keep that change for later.

duboisp (changes requested):
One little change, once completed we will be good to merge this PR.

duboisp (approved):
Tested and reviewed.

This is only a gcweb-jekyll patch because it did touch some includes file where their output didn't change. There is no impact regarding the GCWeb versionning API.

PR #2270, opened 2023-10-19
Documentation: Adding Design patterns component category

Garneauma:


duboisp:
This seems to be content change, pre-approved after review and local testing

duboisp (changes requested):
Tested locally and it do work as expected.

While we are waiting for the first design pattern, I recommend that we do create the container folder. You can do that by adding a `.gitkeep` file temporarily inside the folder which we can remove after there is one or a few design pattern.

Note: This PR is only a content change.

Garneauma:
> While we are waiting for the first design pattern, I recommend that we do create the container folder. You can do that by adding a `.gitkeep` file temporarily inside the folder which we can remove after there is one or a few design pattern.

Thank you @duboisp, I was not aware that such file existed!

duboisp (approved):
Reviewed and tested locally, work as expected

This is a content change

PR #2275, opened 2023-11-07
Content - Borders: Created Example Pages

SebastianBurke:
Accidentally deleted old PR while trying to amend.
Changes reflected here


duboisp (changes requested):
You will need to re-create the commit because it is tagged as a merge commit.

The easiest is to redo the change and then do a force push to your branch "spacing".

Let's connect to do that.

duboisp (changes requested):
As discussed, please add an example of each style documented in the card.

Thanks

common/spacing-borders/spacing-borders-en.html
```diff
@@ -0,0 +1,320 @@
...
+	<div class="col-md-4 brdr-lft brdr-rght brdr-tp brdr-bttm">
+		<div class="margin-top-medium">Medium top Margin</div>
+	</div>
+	<div class="col-md-4 brdr-lft brdr-rght brdr-tp brdr-bttm">
+		<div>Default top Margin</div>
+	</div>
+	<div class="col-md-4">
+		<pre><code>&lt;div class="margin-top-medium"&gt;&lt;/div&gt;</code></pre>
+	</div>
+</div>
```
duboisp, line 66: Can you move those example at the bottom of the page as they are not recommended. They were only added to ease a transition between WET 3 to WET 4.

Please put it in a section named "Spacing utility not recommended" followed by an expand/collapse to view those item where you will downgrade the h2 for an h3. Something like:

```
<h2>Spacing utilities that are not recommended</h2>
<details>
<summary>Only to ease fast content transition from WET 3 to WET 4</summary>

<h3><code>.margin-bottom-none</code></h3>
....

</details>
```

common/spacing-borders/spacing-borders-en.html
```diff
@@ -0,0 +1,320 @@
...
+		style="display: flex; flex-direction: column; height: 500px; background-color: #f0f0f0">
+		<div class="bg-info">Top Element: This is a standard flex item which
+			will stay at its natural position, which is at the top of the flex container.
+		</div>
+		<div class="mt-auto bg-info ">Bottom Element with .mt-auto:
+			By applying .mt-auto, this element takes up all available space on the top, effectively pushing
+			itself to the bottom of the flex container.
+		</div>
+	</div>
+</div>
```
duboisp, line 160: Move those items to be the first item in this example page.

common/spacing-borders/spacing-borders-en.html
```diff
@@ -0,0 +1,320 @@
...
+	</div>
+	<div class="col-md-4 brdr-lft brdr-rght brdr-tp brdr-bttm">
+		<div>No bottom Margin</div>
+	</div>
+	<div class="col-md-4">
+		<pre><code>&lt;div class="mb-sm-5"&gt;&lt;/div&gt;</code></pre>
+	</div>
+</div>
+<div>
+	<h2><code>.mrgn-*-*</code></h2>
```
duboisp, line 80: This change is optional for the purpose of this working example but if you do it you will need to update all the other heading to be consistant. And do the same in French.

Ideally we will have a more meaningful name like:

```suggestion
	<h2>Margin generic utilities - <code>.mrgn-*-*</code></h2>
```

duboisp (changes requested):
I tested locally and the content look good.

Please apply the content re-ordering proposition.

Optionally, it will be great if you can update the heading with a human readable text which describe the associated demo.

Garneauma (approved):

PR #2276, opened 2023-11-07
Archive notice update to fix issues with overlay

nfrenette:
Removed the overlay from the Archive notice, and the width now matches the width of header.

Instead of fixing the overlay and moving code around, we decided to remove the overlay entirely, so that we didn't have to fix the following 3 issues:

1. overlay must not cause a horizontal scroll bar
2. close button must be visible on all screen sizes
3. user must be able to use Tab key to close overlay

Issue 1: At around 800 pixel width, the overlay “x” button to close is not visible on certain devices.
 
Screenshot of missing close button
image
 
Screenshot of visible close button on smaller device/width 
image
 
Screenshot of visible close button on larger device/width
image
 
 
Issue 2: Horizontal scroll bar around the same screen size where the close button disappears.
image

### DIFF
--- a/sites/headers-includes/archive.html
+++ b/sites/headers-includes/archive.html
@@ -1,13 +1,4 @@
-<section class="gc-archv wb-inview modal-content show-none bar-demo" data-inview="top-bar">
-	<div class="container">
+<section class="container gc-archv wb-inview modal-content show-none bar-demo" data-inview="top-bar">
 	<h2>{{ i18nText-archived }}</h2>
 	<p>{{ i18nText-archived-desc }}</p>
-	</div>
-</section>
-
-<section id="top-bar" class="gc-archv wb-overlay modal-content overlay-def wb-bar-t">
-	<div class="container">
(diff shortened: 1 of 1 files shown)


duboisp:
Todo for wet-boew maintainer:
* [ ] Move the include file in the site component folder
* [ ] Create a new documentation page for this site component
* [ ] Need to document the current version 
* [ ] Document this change as a minor change where we technically support 2 different pattern.

(ref num: bsd-27705)

Garneauma:
@nfrenette I have opened a new PR that fixes the scrollbar issues and hidden close button (https://github.com/wet-boew/GCWeb/pull/2289). The only issue left from your list is the keyboard accessible button. I intend to fix that in another PR in WET-BOEW.

Here is a link to a demo of this fix: https://servicecanada.github.io/wet-boew-demos/gcweb-pr2289/sites/archived/archived-en.html

Please let me know if this would be a suitable fix for your issues.

PR #2277, opened 2023-11-08
Feedback: deprecating RAP and Page details 2.0

Garneauma:
- Deprecating RAP
- Deprecating Page details 1.1 and 2.0
- Updating code samples with correct characters

### DIFF
--- a/sites/feedback/index.json-ld
+++ b/sites/feedback/index.json-ld
@@ -17,7 +17,7 @@
 	"modified": "2023-01-10",
 	"componentName": "feedback",
 	"status": "stable",
-	"version": "1.5",
+	"version": "2.0",
 	"pages": {
 		"docs": [
 			{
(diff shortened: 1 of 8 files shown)


sites/page-details/index.json-ld
duboisp, line 1: Deprecate the version 2.x and make the version 2.1 to be the version 3.0 of the component.

Remove the implementation of version 2 and create its own change set by duplicating the one you created.

sites/feedback/index.json-ld
duboisp, line 1: Through the implementation, this "version 2.1" need to be change for "version 3.0" in regard of the page details.

duboisp (changes requested):
2 little change request. 

Other than that its all good. I reviewed all the code and tested locally.

duboisp (changes requested):
One little change because we released page-details version 3.

Garneauma:
> One little change because we released page-details version 3.

Done. Sorry, I missed that.

PR #2280, opened 2023-11-16
GCWeb Menu: fixing doc typo

Garneauma:
Fixing small typo in GCWeb Menu documentation where the sitemenu ajax file was updated to v3 when it should have remained v2.


duboisp (approved):
Reviewed, this is only a content change

PR #2281, opened 2023-11-16
Feedback: removing soft launch alert, iteration and changeset edit

Garneauma:
- Removing soft launch alerts
- Moving iterations and changesets so that the newest ones are at the top
- Moving content of "semantic" to "include" property and adding correct info in "semantic" property
- Small fix in documentation layout so that the "breaking", "additions", and "fixes" are displayed


duboisp:
Pre-approved upon review and local build. This seems to be a content change only.

duboisp (approved):
Reviewed and tested locally

Work as expected and this is only a content change.

PR #2287, opened 2023-11-22
Breadcrumbs: adding GCWeb Home globally

Garneauma:
Adding link to GCWeb homepage in breadcrumbs for all pages with a way to overwrite it using "`overwriteBreadcrumbs: true`".

Relevant files:
- Gruntfile.coffee
- _config.yml
- sites/breadcrumbs/includes/breadcrumbs.html
- sites/includes/settings.liquid


duboisp (approved):
Reviewed and tested locally, work as expected.

I confirm this is a content change and a minor change to GCWeb Jekyll. No impact on the versioning API of GCWeb product.

PR #2291, opened 2023-11-29
Field flow: Handle HTML tags in unlinked list items

EricDunsworth:
Previously, when the plugin encountered list items that didn't start off with an anchor, it used each item's first child node as a label/value. That was fine if the only child was a text node. But not when nested elements were in play...

For example:
* If a list item's first child node was an element node, ``wb.escapeAttribute()`` raised the following JS exception (since it expects strings - not element nodes):
  * Uncaught TypeError: can't access property "replace", str is null
* If a list item started off with a text node and was followed by any element nodes, only the text node's content would be retained... resulting in the rest of the content getting truncated

This fixes it by deriving list item labels from their inner HTML code (like what's already done for anchors), as well as stripping HTML tags out of ``option`` elements and ``value`` attributes. HTML formatting is retained for checkbox/radio button labels.

Nested structure that isn't meant to be shown in grouping (``ul``) and nesting (``.wb-fieldflow-sub``) scenarios is excluded.

Also updated examples to demonstrate affected scenarios:
* Redirection example:
  * Bold a link in a list item
* Other examples:
  * Add italics/bolding in all "(Set 5)"/"(Ensemble 5)" list items

### DIFF
--- a/components/wb-fieldflow/fieldflow-en.html
+++ b/components/wb-fieldflow/fieldflow-en.html
@@ -6,7 +6,7 @@ description: Transform a basic list into a selectable list.
 tag: fieldflow
 parentdir: fieldflow
 altLangPage: fieldflow-fr.html
-dateModified: 2016-11-30
+dateModified: 2023-12-01
 ---
 <ul class="list-inline">
 	<li><a class="btn btn-primary" href="fieldflow-doc-en.html">Documentation</a></li>
@@ -50,7 +50,7 @@ dateModified: 2016-11-30
 		<li><a href="https://wet-boew.github.io/v4.0-ci/demos/charts/charts-en.html">Draw charts</a></li>
 		<li><a href="https://wet-boew.github.io/v4.0-ci/demos/details/details-en.html">Expand and collapse content</a></li>
 		<li><a href="https://wet-boew.github.io/v4.0-ci/demos/equalheight/equalheight-en.html">Set a consistant height</a></li>
-		<li><a href="https://wet-boew.github.io/v4.0-ci/demos/overlay/overlay-en.html">Popup content</a></li>
+		<li><a href="https://wet-boew.github.io/v4.0-ci/demos/overlay/overlay-en.html"><strong>Popup</strong> content</a></li>
 	</ul>
 </div>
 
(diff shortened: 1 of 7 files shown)


components/wb-fieldflow/fieldflow.js
```diff
@@ -829,7 +829,7 @@ var componentName = "wb-fieldflow",
 			}
 
 			if ( !itmLabel ) {
-				itmLabel = firstNode.nodeValue;
+				itmLabel = $( itmCached ).html();
```
EricDunsworth, line 832: This matches what's being done for anchors at:
https://github.com/wet-boew/GCWeb/blob/74b26ebe2453008fb1c45b13f6562f2eef18bb9a/components/wb-fieldflow/fieldflow.js#L789

(I probably would've used the [``innerHTML`` property](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML) instead of [jQuery's ``html()`` method](https://api.jquery.com/html/) if there hadn't been any precedent)

EricDunsworth:
Gonna shift this back into a draft since what I currently have is probably unable to cope with named/numeric entities. So I think it'll need a bit more work.

EricDunsworth:
I think I'll put this back into ready for review "as-is".

I did more research/testing earlier. Looks like modern browser rendering engines (i.e. Gecko/Blink/etc) are already designed to automatically convert entities into their "native" characters. So there's no need for special handling of entities in my logic. Browser engines already do what I had in mind "for free" :).

duboisp:

components/wb-fieldflow/fieldflow-en.html
duboisp, line 1: There is a side effect for the "Grouping" working example. An adjustment to the updated code will be required.

image

duboisp, line 1: There is a side effect for the "Nesting" working example. An adjustment to the updated code will be required.

image

duboisp (dismissed):
I did a partial review since that I found that 2 working example are going to be broken.

See the "Grouping" and the "Nesting" working example. Extra text is added.

Ex:
image

EricDunsworth:
Just force-pushed an update to fix the grouping/nesting issues + rebased + updated the commit message/OP :).

@duboisp Thanks for catching those! I probably didn't think of testing them initially since my eyes were glued to the examples I modified :P.

Btw the nesting example's labels/values will now end with an extra line break and a couple of tab characters (since the text node that contains them is now being included). That isn't a problem and is in line with this PR's intended purpose. But it might be worthwhile to add some trimming logic in the future.

duboisp:
We will do a review + local testing.

Garneauma:
Pre-approved upon successful testing.

Garneauma:
@EricDunsworth Was there a reason you did not include logic to strip extra lines and tabs from the value attribute?

Garneauma (approved):
Went through the working examples and they all are working as expected with the value and labels fixed.

PR #2292, opened 2023-11-30
GC-Subway: fixing skip link href

Garneauma:


components/gc-subway/gc-subway.js
```diff
@@ -47,8 +47,7 @@ var $document = wb.doc,
 
 				// Add skip link to sections list
 				if ( $navH1 ) {
-					$navH1.id = $navH1.id || wb.getId(); // Ensure the element has an ID
-					wb.addSkipLink( wb.i18n( "skip-prefix" ) + " " + $navH1.text(), { href: "#" + $navH1.id } );
+					wb.addSkipLink( wb.i18n( "skip-prefix" ) + " " + $navH1.text(), { href: "#" + elm.id } );
```
duboisp, line 50: I suggest to link to the `h1` instead of the `nav` element.

The issue with the previous is:
* `$navH1 = $( "h1", $elm );` do return multiple h1
* `$navH1.id = $navH1.id || wb.getId();` are not setting the id on the h1. The `id` property is only available on HTML element. This can be alternatively done, where it do apply an id on the first h1: `$navH1.get( 0 ).id = $navH1.get( 0 ).id || wb.getId();`

For your info:

Here what it looks like when linking to the `nav` element
image


Here what it looks like when linking to the `h1` element

image

duboisp (changes requested):

duboisp (approved):
Reviewed and tested locally, 

It does work as expected.

PR #2295, opened 2023-12-05
Breadcrumbs: Adding Design system link

Garneauma:


duboisp:
@delisma let us know what you think about this change. Thanks

duboisp:
Cancelling this change as it is not recommended by DTO.

PR #2296, opened 2023-12-05
Méli-mélo - 2021-05 Steps - Update css

Christopher-O:
Update to avoid cut off of lead zero in expand/hides in mobile and only allow margin top 0 for first child items in striped pattern.


duboisp:

duboisp (approved):
Reviewed and tested locally

Still conform to our méli-mélo guidance and the updated CSS selector are demoed.

PR #2300, opened 2023-12-13
Add checklist funcitonality

fsnoddy:
Add GC-Checklist to GCWeb. 

Code was mostly written by David Elisma but I am willing to sponsor it.

### DIFF
--- a/index-en.md
+++ b/index-en.md
@@ -1,7 +1,7 @@
 ---
 title: GCWeb, the WET-BOEW Canada.ca theme
 altLangPage: index-fr.html
-dateModified: 2024-12-09
+dateModified: 2025-01-15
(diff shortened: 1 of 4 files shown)


Christopher-O:
We at CRA were working on a eligibility checklist function for GCWeb, so this is great that something is coming along.  I will ask the developer to team up with this group as we do require a bit of extended functionality for the feature.  The two main ones would be a conditional statement (e.g. if more than one, but not all checklist items are checked does the message change),  as well, some of our content are "or" conditions, and would require structured data (e.g. or pattern) or interactive elements which is not really supported with the label element, and may require some further thinking (e.g. using aria's, though they have their drawback as well).

PR #2306, opened 2024-01-03
Relocate JSON Manager and Data JSON under wet-boew folder

Ricokola:
This is content change only. The JSON Manager and Data JSON folders have been moved under the wet-boew folder located in GCWeb. The previous and existing files used as redirect have been deleted.


duboisp (approved):
Révisé et testé localement. Fonctionne tel que prévue et je confirme que c'est un changement de contenu seulement.

PR #2308, opened 2024-01-12
Meli melo 2023 09 menu

nfrenette:
Removed code that generates menu attributes, added data-ajax calls for both menus, made minor changes to CSS.


méli-mélo/2023-09-menu/campaign-menu.js
```diff
@@ -180,18 +126,30 @@
 
                         // Determine if the focus is on a link or an anchor to open a submenu
                         if (savedFocusHref.match("^#")) {
-                            let submenuId = savedFocusHref.substring(6);
-                            $(".gcweb-menu a[aria-controls='sub-menu-" + submenuId + "']").focus();
+                            let submenuId = savedFocusHref.substring(14);
```
duboisp, line 129:
```
.substring( 14 )
```
That instruction look like to be very chirurgical. Not an issue for this PR, but it will be preferable to calculate the length to be used by the substring instead of using a fix length.

méli-mélo/2023-09-menu/campaign-menu.js
```diff
@@ -203,8 +161,13 @@
                         // Set focus on submenu
                         if (savedFocusHref.match("^#")) {
                             let gcSubmenuFocus = $(savedAnchorFocus).attr("aria-controls");
-                            let submenuId = gcSubmenuFocus.substring(9);
-                            $(".wb-menu a[href='#menu-" + submenuId + "']").focus();
+                            let submenuId = gcSubmenuFocus.substring(12);
```
duboisp, line 164:
```
.substring( 12 )
```
That instruction look like to be very chirurgical. Not an issue for this PR, but it will be preferable to calculate the length to be used by the substring instead of using a fix length.

duboisp (approved):
Reviewed and tested, all those changes are in the scope of the experimental project guideline (méli-mélo project)

I made 2 comments, but those are not a blocker for this PR.

PR #2318, opened 2024-02-21
Follow us: replacing Twitter with X

Garneauma:
Changes related to WET-454


components/gc-follow-us/_base.scss
```diff
@@ -310,7 +314,7 @@ $social-media-icons-size: 38px;
 	}
 
 	&.twitter {
-		@extend %social-media-icons-logo-twitter;
+		@extend %social-media-icons-logo-x;
 	}
```
duboisp, line 319: Should we add a new style exclusively for the social media X? I am thinking a css class like `social-media-x` or something similar.

```suggestion

	&.social-media-x {
		@extend %social-media-icons-logo-twitter;
		@extend %social-media-icons-logo-x;
	}

```

components/gc-follow-us/followus.js
```diff
@@ -0,0 +1,43 @@
...
+		var elm = wb.init( event, componentName, selector ),
+			$elm;
+
+		if ( elm && event.currentTarget === event.target ) {
+			$elm = $( elm );
+
+			elm.querySelector( ".twitter .wb-inv" ).innerHTML = "X";
+
+			// Identify that initialization has completed
+			wb.ready( $elm, componentName );
```
duboisp, line 33: Considering we use the $elm only once, let's do the jQuery encapsulation only where it needed.

```suggestion
			;

		if ( elm && event.currentTarget === event.target ) {

			elm.querySelector( ".twitter .wb-inv" ).innerHTML = "X";

			// Identify that initialization has completed
			wb.ready( $( elm ), componentName );
```

components/gc-follow-us/_base.scss
```diff
@@ -24,6 +24,10 @@
...CNjAwMUMiIHhtcE1NOkRvY3VtZW50SUQ9InhtcC5kaWQ6RjEwNzlCNEM2Mjc3MTFFMzgzNjNDMkIzMzRCNjAwMUMiPiA8eG1wTU06RGVyaXZlZEZyb20gc3RSZWY6aW5zdGFuY2VJRD0ieG1wLmlpZDpGMTA3OUI0OTYyNzcxMUUzODM2M0MyQjMzNEI2MDAxQyIgc3RSZWY6ZG9jdW1lbnRJRD0ieG1wLmRpZDpGMTA3OUI0QTYyNzcxMUUzODM2M0MyQjMzNEI2MDAxQyIvPiA8L3JkZjpEZXNjcmlwdGlvbj4gPC9yZGY6UkRGPiA8L3g6eG1wbWV0YT4gPD94cGFja2V0IGVuZD0iciI/PiBBaDwAAAJWSURBVHja3JZLbExhGIZnqihCoqTVhroUC9Rlo5IGISFYWIiVxNLGhogNO4nEylqwkEiQ7qQblxAWhLjEJK4LhBKKShtUW4zxvPJWTkb/c07NZBbe5MlMp2fOe77L/32TLRQKmUqpKlNBVdSsuvV8Zynfb4XdsB5q4RmchFN+r2DmQyPkqmUIk/2PdyMwmg3rYIWNpDmwGl7DGKiHJfAEvsloEmzyE5yBBymMpsIe2AFjI59nYS3MhTswHX7Cfr0qmgnQBnvhALQ42pBGwQbYVmQUVRNsgZlw3691VU7fOIetC47AVpgSuFEDbHTqk1QHS+ELdCuCAXgVuaDN4a+EsyosfIZBKDjt9U5ZknrdLJdlKLNPcB0ewQJfpLB3wirn/p6LrNadBn2Qd0rj9BRO2ON3bfrhGrTDrkhnSYvMZkff64I3pYhKWXgRPctD3bjGN+gsMhtSbeDzJH2NRi/XiS74drd0uaRUd8OPqFkPXIAPboxyqQtuOro/ZvrjIhyFj2U008Pfhu/Fg1jn4JgNu9xppUqT6H1o6iu/t+BhNPR/lIZwh+ZhyEzva2AxjC/R7IrPbj5kptzehUspDmuccl4zPX/ts2HC3wePPb3VnTNiBm6xdPAPuxyJm1qn/iUc8lrIjWCbayAc9DwdHHZTB6ZFs5eg9tLoFEbP/YDt7uxMyGyWh2uDp/kyb995XjtxUl1uwHEPhv7Y3yCwEJZ7ITZ7mdYErh/wDftc19Nwzscm8WzK7KoL2mHTFps2Rpoj75roDL719freG2+BVMr+tz9SfwkwAFjOiZEdguqDAAAAAElFTkSuQmCC");
```
duboisp, line 24: After this change, do we still need to keep the twitter icon? Is it used elsewhere?

```suggestion
```

components/gc-follow-us/_base.scss
```diff
@@ -310,7 +314,7 @@ $social-media-icons-size: 38px;
 	}
 
 	&.twitter {
```
duboisp, line 316: Note: this style is used here: https://wet-boew.github.io/GCWeb/templates/widgets-en.html

May be we should move that old working example page within this component instead.

components/gc-follow-us/followus.js
```diff
@@ -0,0 +1,43 @@
...
+ * @overview Plugin used to replace Twitter with "X" - Deprecated
+ * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * @author @garneauma
+ */
+( function( $, window, wb ) {
+"use strict";
+
+var $document = wb.doc,
+	componentName = "followus",
+	selector = "." + componentName,
```
duboisp, line 12: We might need to support the conversion of the other use of the twitter icon. Like on this working example page: https://wet-boew.github.io/GCWeb/templates/widgets-en.html

So, this plugin will need to support also situation where ".icon.twitter" is used where the text is fully visible.

components/gc-follow-us/index.json-ld
duboisp, line 1: We might need to include in our legacy doc this example: https://wet-boew.github.io/GCWeb/templates/widgets-en.html

And the "followus" need a status set to "demoted" until we officially go ahead to deprecate it within our next major version of GCWeb.

components/gc-follow-us/followus.js
```diff
@@ -0,0 +1,43 @@
...
+		// Start initialization
+		// returns DOM object = proceed with init
+		// returns undefined = do not proceed with init (e.g., already initialized)
+		var elm = wb.init( event, componentName, selector ),
+			$elm;
+
+		if ( elm && event.currentTarget === event.target ) {
+			$elm = $( elm );
+
+			elm.querySelector( ".twitter .wb-inv" ).innerHTML = "X";
```
duboisp, line 30: Would be safer if we only replace the word "Twitter" by "X" instead of assuming the word "twitter" is there?

Something like:
```suggestion
			// Swapping the text "Twitter" for "X"
			var twitterElm = elm.querySelector( ".twitter" );
			twitterElm.innerHTML = twitterElm.innerHTML.replace( "Twitter", "X" );
```

duboisp (changes requested):
Reviewed and it do work as expected.

I found that we might miss-placed a working example "widgets-en.html" which may need to be moved along with the social media channel component. The documentation might need to be adjusted too.

Need to adjust the status of the followus for "demote" because deprecation only happen when releasing a major version.

We might don't need to keep the old twitter icon considering we are replacing it by "X".

components/gc-follow-us/followus.js
Garneauma: Agreed. I also added a condition to check if the twitter query returned something because otherwise it would throw an error.

components/gc-follow-us/_base.scss
Garneauma: Added ".x-social"

duboisp (approved):
Reviewed and tested locally. Everything work as expected.

The switch from existing Twitter to X is a Minor change only because it is visually perceptible. No action is required by the author.
The addition of the named social media "x-social" is a Minor change.

PR #2199, opened 2023-07-13
Images - add working examples from WET-BOEW style guide

Ricokola:
Addition of working examples of the Images component form the WET-BOEW Style Guide

---
close #2323

PR #2323, opened 2024-02-29
Wet 321 images with some changes

AtlakM:


duboisp (approved):
Reviewed and tested locally and everything work as expected.

I am going to move all those commits inside the PR #2199

PR #2324, opened 2024-03-05
Field flow: Trim auto-generated labels/values

EricDunsworth:
Leading/Trailing spaces are usually undesirable, so let's strip them out.

Doesn't affect labels that are explicitly-declared via the "wb-fieldflow-label" class. Those will be kept as-is so as not to interfere with an implementer's desire.

Affects grouping and nesting scenarios. Directly impacts the demo page's nesting example.

Spinoff of #2291.


Garneauma:
Pre-approved upon successful review of all working examples.

EricDunsworth:
Just force-pushed an updated commit message (added a spinoff comment). Left everything else as-is.

Garneauma (approved):
Tested all working examples locally. Everything works as expected.

PR #2327, opened 2024-03-06
Field flow: Cleanup optgroup labels

EricDunsworth:
Removes HTML tags and escapes quotes from ``optgroup`` ``label`` attributes.

Also adds a ``strong`` element to the grouping example to demonstrate it in action.

Overlooked leftover from #2291.


Garneauma:
Pre-approved upon successful review of all working examples (specifically optgroup).

PR #2328, opened 2024-03-06
Field flow: Detect anchors via firstElementChild

EricDunsworth:
Anchors were previously detected by evaluating ``firstChild``'s node name... which only worked if the very first node was an anchor element.

As a result, if the first node was something other than an element (e.g. comment or text node with empty space)... they'd be evaluated instead of the real anchor and consequently mess up the redirection action.

This resolves it by evaluating ``firstElementChild`` instead of ``firstChild``. Also adds a comment and empty space to some of the redirection examples.

### DIFF
--- a/components/wb-fieldflow/fieldflow-en.html
+++ b/components/wb-fieldflow/fieldflow-en.html
@@ -6,7 +6,7 @@ description: Transform a basic list into a selectable list.
 tag: fieldflow
 parentdir: fieldflow
 altLangPage: fieldflow-fr.html
-dateModified: 2024-03-06
+dateModified: 2024-04-08
 ---
 <ul class="list-inline">
 	<li><a class="btn btn-primary" href="fieldflow-doc-en.html">Documentation</a></li>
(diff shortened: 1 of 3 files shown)


EricDunsworth:
Just force-pushed a couple of super minor updates:
* Mention the comment example in the commit message (previously only mentioned empty space)
* Update modified dates

Garneauma (changes requested):
@EricDunsworth There is a javascript error: `theme.js:2550 Uncaught TypeError: Cannot read properties of null (reading 'nodeName')`

EricDunsworth:
@Garneauma Ugh sorry :(. Will fix it in a sec.

When I prepped/tested this PR on my end, I edited my "built" theme.js file directly and everything worked fine... except I forgot to copy a snippet of code into the src folder's fieldflow script 🙃.

My next force-push will change this:
```js
if ( firstElmNode.nodeName === "A" ) { ... }
```

Into this:
```js
if ( firstElmNode && firstElmNode.nodeName === "A" ) { ... }
```

EricDunsworth:
Just force-pushed the the aforementioned change and updated modified dates.

@Garneauma Thanks for catching that 👍!

EricDunsworth:
Just rebased, updated modified dates and force-pushed the outcome :).

Garneauma (dismissed):
Works very well. Thank you for this fix @EricDunsworth.

EricDunsworth:
Just rebased, updated modified dates and force-pushed :).

The merge conflicts turned out to be just the dates.

Garneauma (approved):

PR #2348, opened 2024-04-18
Instititional landing page: content change on French working example

Garneauma:


duboisp (approved):
Reviewed, and this is only a content change. 
I didn't test locally, but it should be fine technically, and I validated that the json file should exist at the specified location.

So all good to merge.

PR #2354, opened 2024-04-30
add accessible name to search form

shawnthompson:
change method to get


duboisp:
We will need to documents this change and review the propose change.

@shawnthompson let discuss how we can do that. Can you join us at our weekly office hours?

shawnthompson:
@duboisp can you send me the invite, I no longer have it in my calendar. I'll Teams you my personal email.

PR #2355, opened 2024-05-01
Tagfilter: adding demo/doc links and partial a11y report

Garneauma:
Related to WET-466


duboisp (approved):
Reviewed and tested locally. Everything work as expected.

This is a content change.

PR #2357, opened 2024-05-01
Update campaign-menu.css

gjb-pch:
Update menu background color to black

### DIFF
--- a/méli-mélo/2023-09-menu/campaign-menu.css
+++ b/méli-mélo/2023-09-menu/campaign-menu.css
@@ -148,15 +148,15 @@
 /*** Secondary colour cm-bg-darker ***/
 /* add background and text color for menu */
 #wb-sm.campaign-menu.cm-bg-darker {
-    background: red;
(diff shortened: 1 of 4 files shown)


Garneauma:
Pre-approved upon successful review and assuming it still follows the implementation plan.

méli-mélo/2023-09-menu/campaign-menu.css
```diff
@@ -1,7 +1,7 @@
 /*** Mega Menu ***/
 /* add background and text color for menu */
 #wb-sm.campaign-menu {
-    background: #003c77;
+    background: #000;
```
duboisp, line 4: After discussion,

Let keep the blue as the default color and you can add a feature which are going to support a generic "bg-darker" variant. We will need to have an example of both style.

nfrenette: Fixed: GC theme blue is the default colour, and black is available as a feature/option.

PR #2369, opened 2024-05-22
Services and information: updating doc format and adding translation

Garneauma:


duboisp (changes requested):
2 small change.

duboisp (commented):
Reviewed the rest and everything look good. This is a content change.

Great work on this

duboisp (changes requested):
You will need to update the template page that are referencing this component via a jekyll include like:

* gc-audience
* organizational
* institution-arms
* institution-landing
* organizational-arms
* old institution-landing

Other than that, everything look good.

duboisp (approved):
Reviewed and approved.

Work as expected.

This a content change. 
Note the SASS file has be renamed, but their content hasn't been changed and the result of the compiled CSS should be rendered exactly the same as before this renaming.

PR #2377, opened 2024-06-04
Page details: remove share this page

Garneauma:
Changes related to WET-458


Garneauma:

Ricokola (approved):
Reviewed and tested locally and seems to works and implemented as expected.

duboisp (changes requested):
À première vue, je crois qu'il nous manques un "deprecated" example qui contient le share this page. Je vois le code mais pas l'exemple associé.

duboisp (changes requested):
In addition to the demoted working example, can you add a quick implementation for "Migration to v4" in the documentation? and mention this is the removal of "share this page" in the intro.

I completed the review and local testing. Other than that it should be good.

duboisp (dismissed):
Reviewed and tested locally

All is documented, demoed and work as expected.

duboisp:
This is a minor change because we demoted the pattern. We will deprecate officially that pattern at our next major release.

duboisp (approved):
Thanks for changing the version number for a minor change.

PR #2378, opened 2024-06-04
Breadcrumbs: fix GCWeb Jekyll breadcrumbs link when setting relative path

Garneauma:


Ricokola (approved):
Reviewed and tested locally! Works and behave as expected.

duboisp (approved):
I tested locally and it is working as expected.

I observe the "/" is not prepend after this fix is applied. So the breadcrumb link do work as expected.

duboisp:
This change is out of scope of GCWeb product. This is only impacting GCWeb Jekyll as a patch change.

PR #2379, opened 2024-06-07
Updating TC links to https://tc.canada.ca/ instead of https://www.tc.gc.ca/

alexandre-trottier:
The menu does not use the new tc.canada.ca domain and might stop working in the future.


Garneauma:

Garneauma (approved):

duboisp:
@alexandre-trottier For you information, we also updated the links in the ajax top menu hosted on Canada.ca

PR #2382, opened 2024-06-12
GC-intro-block: Added half-width image

BrahimMahadi:
Added the half-width image to the GC-intro-block, as well as updated the related files.


Garneauma (changes requested):
Here's a list of what I quickly see needs to be done:
- [ ] Move _base.scss out of the samples folder.
- [ ] Add _base.scss to theme.scss, otherwise those styles will never be applied.
- [ ] Move the working example for the half-width image variant to its own page.
- [ ] Adjust the include in "gc-intro-image.html" to point to the correct file name.
- [ ] Rename "gc-intro-image-en/fr.html" to "gc-intro-full-width-image-en/fr.html"
- [ ] Run grunt locally before commit to include the changes to data files and includes.

design-patterns/gc-intro/_base.scss
```diff
@@ -0,0 +1,5 @@
+/* REMOVE ONCE IMPLEMENTED IN WET-BOEW */
```
duboisp, line 1: This can also be implemented as an utilities in GCWeb. If that option it's chosen, then that new feature should go in the common component list under "colour" along with the other background CSS selector.

See: http://localhost:4000/common/colour/colour-en.html

design-patterns/gc-intro/_base.scss
```diff
@@ -0,0 +1,5 @@
...
+.bg-half-width-right {
```
duboisp, line 2: As I proposed earlier

```suggestion
.bg-half-right {
```

And let also add the `bg-half-left` to have a complete utilities suite.

design-patterns/gc-intro/gc-intro-image-en.html
```diff
@@ -1,20 +0,0 @@
...
-	"dateModified": "2024-03-14",
-	"breadcrumbs": [
-		{ "title": "Introduction block - documentation", "link": "design-patterns/gc-intro/gc-intro-doc-en.html" }
-	],
-	"layout": "no-container"
-}
----
-<div class="wb-prettify all-pre"></div>
-
-{% include_relative samples/gc-intro-image.html %}
```
duboisp, line 15: This sample `samples/gc-intro-image.html` is not removed by this PR?

design-patterns/gc-intro/samples/gc-intro-full-width-image.html
duboisp, line 1: Are this replacing "gc-intro-image.html" sample?

design-patterns/gc-intro/gc-intro-full-width-image-fr.html
```diff
@@ -0,0 +1,28 @@
...
+	<h2>Code</h2>
+	{% highlight html %}{% include_relative samples/gc-intro-full-width-image.html %}{% endhighlight %}
+</div>
+
+{% include_relative samples/gc-intro-half-width-image.html %}
+
+<div class="container">
+	<h2>Code</h2>
+	{% highlight html %}{% include_relative samples/gc-intro-half-width-image.html %}{% endhighlight %}
+</div>
```
duboisp, line 28: Is this an extra content duplicate?

_includes/components/gc-intro/gc-intro-full-width-image.html
duboisp, line 1: FYI - You should update those file by modifying the file in the "sample" folder. The file in the "_include" folder are then management by our grunt script. Although you might need to manually delete the unused sample file.

design-patterns/gc-intro/index.json-ld
```diff
@@ -94,38 +93,66 @@
...
+			"name": {
+				"en": "Introduction block with half-width image",
+				"fr": "Bloc d'introduction avec image demi-largeur"
+			},
+			"description": {
+				"en": "The introduction block pattern introduces the content of a landing page. This variation contains an image.",
+				"fr": "La configuration de conception du bloc d’introduction introduit le contenu d’une page d’accueil. Cette variation contient une image."
+			},
+			"guidance": {
+				"en": "https://design.canada.ca/common-design-patterns/intro-block.html",
```
duboisp, line 133: Question for DTO: Do they want us to apply the fading out middle transition by using CSS or should we let the author to do that by using image editing tool?

duboisp (changes requested):
I completed a partial review and I partially tested locally.

See my inline comment.

There is one question for DTO regarding their guidance vs our documentation which is should we apply the middle image transition by using CSS or is that should be handle by the author? 
Preferably, that would be managed by the author otherwise it might require us to create a complementary CSS to apply that middle transition effect.

_includes/design-patterns/gc-intro/gc-intro-half-width-image.html
```diff
@@ -0,0 +1,11 @@
+<div class="bg-norepeat bg-half-width-right" data-bgimg-srcset="https://wet-boew.github.io/vocab/wb/utilities#no-image 991w, https://placehold.co/800x500 992w">
```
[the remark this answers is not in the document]
Garneauma: Not needed anymore.

design-patterns/gc-intro/_base.scss
Garneauma: Not needed anymore.

design-patterns/gc-intro/samples/gc-intro-full-width-image.html
Garneauma: Yes it replaces "gc-intro-image.html".

design-patterns/gc-intro/index.json-ld
Garneauma: Not relevant anymore as DTO decided to use the current intro block pattern for the Topic template.

design-patterns/gc-intro/_base.scss
Garneauma: Not needed anymore.

design-patterns/gc-intro/index.json-ld
```diff
@@ -209,14 +245,69 @@
...
+						"code": "<div class=\"row\">\n\t<div class=\"col-md-6\">\n\t\t<h1 property=\"name\" id=\"wb-cont\">Introduction block with half-width image</h1>\n\t\t<p>The introduction block pattern introduces the content of a landing page.</p>\n\t\t<p><a class=\"btn btn-call-to-action\" href=\"#\">Supertask button</a></p>\n\t</div>\n\t<div class=\"col-md-6\">\n\t\t<img src=\"https://via.placeholder.com/520x200/000000/FFFFFF.png\" alt=\"\" class=\"img-responsive\">\n\t</div>\n</div>"
+					}
+				],
+				"fr": [
+					{
+						"@type": "source-code",
+						"description": "Exemple de code :",
+						"code": "<div class=\"row\">\n\t<div class=\"col-md-6\">\n\t\t<h1 property=\"name\" id=\"wb-cont\">Bloc d'introduction avec image demi-largeur</h1>\n\t\t<p>La configuration de conception du bloc d'introduction introduit le contenu d'une page de destination.</p>\n\t\t<p><a class=\"btn btn-call-to-action\" href=\"#\">Bouton de super-tâche</a></p>\n\t</div>\n\t<div class=\"col-md-6\">\n\t\t<img src=\"https://via.placeholder.com/520x200/000000/FFFFFF.png\" alt=\"\" class=\"img-responsive\">\n\t</div>\n</div>"
```
duboisp, line 255: This code sample don't match the full image width example

design-patterns/gc-intro/index.json-ld
```diff
@@ -209,14 +245,69 @@
...
+				]
+			},
+			"techConsiderations": {
+				"en": [
+					"The half-width image size recommended is 520px by 200px. For more information on the half-width image, please consult the official guidelines."
+				],
+				"fr": [
+					"La taille d’image demi-largeur recommandée est de 520 px par 200 px. Pour plus d’informations sur l’image demi-largeur, veuillez consulter les directives officielles."
+				]
+			},
```
duboisp, line 289: There is an accessibility consideration to add. If we don't have that section, you can add it into the technical consideration section.

Accessibility consideration:
* If the image is decorative where it doesn't provide any message and don't contains any text, ensure the text alternative is empty on the image.
* Otherwise, provide a text alternative that represent the message illustrated by the image and a text alternative that contains every words, written as is, in the text alternative.

duboisp (changes requested):
There is a few change requested inline. Like the gc-intro component in the "component" need to be cleaned in our "_includes" folder.

There is a jekyll error for the `templates/institutional-landing/institutional-landing-en.html` and the ILP working example pages because this file is removed/missing => `design-patterns/gc-intro/samples/gc-intro-image.html`. That error is probably caused because the code was not rebased to the upstream.

Important: Please add a note about the accessibility consideration of using the half-width images.

cc @Garneauma

_includes/components/gc-intro/gc-intro-half-width-image.html
```diff
@@ -0,0 +1,14 @@
+<div class="gc-intro half-width-image">
```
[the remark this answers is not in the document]
Garneauma: Deleted the file.

Garneauma:
The new intro block variation was approved by DTO, so it is ready to be merged once it passes the technical review.

duboisp (approved):
Reviewed and tested locally. Everything look good.

:-)

PR #2388, opened 2024-07-05
Topic page: stabilizing topic page template v2.0

Garneauma:
Introducing the version 2 of the Topic page template.


components/gc-features/samples/one-tile-hidden-title.html
```diff
@@ -0,0 +1,20 @@
+{%- comment -%}
+
+++++++++++++++++++++++++++++++++++
++ List of features - One tile (invisible title)
```
duboisp, line 4: That are not matching the "one tile feature" pattern. But do you mean the "cohabit one tile" design pattern where it depend on a special "context of use"?

If not, then this will need to be added as an additional pattern for the gc-features component

duboisp (changes requested):
Note: Partial and incomplete review,

templates/topic/topic-en.html
```diff
@@ -1,95 +1,37 @@
...
+
+<div class="container">
+	{%- include /components/gc-srvinfo/3col-hidden-title.html -%}
+
+	<div class="row mrgn-tp-xl">
+		<div class="col-md-8">
+			{%- include /components/gc-features/one-tile-hidden-title.html -%}
+		</div>
+		<div class="col-md-4">
+			{%- include gc-follow-us/gc-follow-us.html -%}
```
duboisp, line 32: Take a look at that include, it use a popup for when there is multiple account (see facebook) and now that pattern has changed via the "more" last link in that list.

templates/topic/index.json-ld
```diff
@@ -28,19 +29,195 @@
...
+		}
+	],
+	"implementation": [
+		{
+			"@id": "_:implement_topic",
+			"iteration": "_:iteration_topic_3",
+			"name": {
+				"en": "Standard",
+				"fr": "Standard"
+			},
```
duboisp, line 115: No code sample? or at least where to put the component (with their version #) + the page class to add.

templates/topic/index.json-ld
```diff
@@ -28,19 +29,195 @@
...
+				"en": "AEM users",
+				"fr": "Utilisateurs AEM"
+			},
+			"introduction": {
+				"en": "This implementation is meant for publishers adding the template manually into an AEM page.",
+				"fr": "Cette implémentation est destinée aux éditeurs qui ajoutent le gabarit manuellement à une page d'AEM."
+			},
+			"instructions": {
+				"en": [
+					"TODO"
```
duboisp, line 144:
```suggestion
					"See the AEM documentation on GCpedia."
```

templates/topic/index.json-ld
```diff
@@ -28,19 +29,195 @@
...
+				],
+				"fr": [
+					"TODO"
```
duboisp, line 147:
```suggestion
					""Consulter la documenation de AEM sur GCpedia.""
```

templates/topic/index.json-ld
```diff
@@ -28,19 +29,195 @@
...
+				]
+			},
+			"ajaxSourceCode": {
+				"en": "topic-en.html main > *:not(.pagedetails)",
+				"fr": "topic-fr.html main > *:not(.pagedetails)"
```
duboisp, line 152: I suggest to move that code sample for the "Standard" while we are waiting for the instruction for AEM.

templates/topic/index.json-ld
```diff
@@ -28,19 +29,195 @@
...
+					"en": "June 2024 - Stabilization of the version 2 of the page template.",
+					"fr": "Juin 2024 - Stabilisation de la version 2 du gabarit de page."
+				},
+				{
+					"en": "January 2021 - Provisional implementation of the version 2 of the page template.",
+					"fr": "Janvier 2021 - Implémentation provisoire de la version 2 du gabarit de page."
+				},
+				{
+					"en": "August 2014 - Implementation of the version 1 of the page template.",
+					"fr": "Août 2014 - Implémentation de la version 1 du gabarit de page."
```
duboisp, line 103: The versioning of the topic page has not been well recorded until Jan 2018 and it did get a few breaking change during that period. 
May be it will be best to say: As per the template state on Jan 2018, or something similar.

templates/topic/index.json-ld
```diff
@@ -28,19 +29,195 @@
...
+			"@id": "_:cs_topic_2",
+			"name": "Topic page",
+			"status": "stable",
+			"baseOnIteration": "_:iteration_topic_2",
+			"detectableBy": ".page-type-topic",
+			"layout": [
+				"Intro block",
+				"Most requested - version 1",
+				"Services and information - version 3",
+				"Context-specific features - version 5.0 [.col-md-8] + Social media channels - version 1 [.col-md-4]",
```
duboisp, line 218: "Social media" is not the "follow us" component? do we have 2 distinct component? I think that should be noted as being the version 2.

For the "features" see my note to confirm or not if we are using the co-habit pattern.

templates/topic/topic-en.html
```diff
@@ -1,95 +1,37 @@
...
-	],
-	"dateModified": "2023-09-01",
-	"share": "true",
-	"deptfeature": true
-}
+title: "[Topic title]"
+layout: no-container
+language: "en"
+altLangPage: "topic-fr.html"
+pageType: "topic"
```
duboisp, line 6: Do we use that page property? I think it is only useful when we are archiving a page

```suggestion
```

duboisp (changes requested):
The features component would need to be incremented by a minor change. It should be set to 5.1.0. Idem under the "template" version + code sample. But are you sure this is not the co-habit pattern that it is used? It looks like it in the topic working example page

The lowest-topic edge case need to be moved into the demoted too.

Is there any specific style that was only applicable to the topic page? if so where they are located?

templates/topic/index.json-ld
Garneauma: We did not have a code sample or anything like this for the ILP. Should I add that to both?

templates/topic/index.json-ld
Garneauma: You are right, it should be version 2.1.

components/gc-features/gc-features-doc-en.html
duboisp, line 1: Please revert back those change. It was the working example that didn't implemented the correct markup.

Please fix also the co-habit working example.

duboisp (changes requested):

templates/topic/index.json-ld
```diff
@@ -30,17 +31,215 @@
...
+			"iteration": "_:iteration_topic_3",
+			"example": [
+				{
+					"en": { "href": "topic-en.html", "text": "Topic page" },
+					"fr": { "href": "topic-fr.html", "text": "Page de sujet" }
+				},
+				{
+					"en": { "href": "demoted/topic-en.html", "text": "Topic page version 1 - Demoted" },
+					"fr": { "href": "demoted/topic-fr.html", "text": "Page de sujet version 1 - Rétrogradée" }
+				}
```
duboisp, line 110: Remove the demoted working example from this variation

```suggestion
				}
```

templates/topic/index.json-ld
```diff
@@ -30,17 +31,215 @@
...
+			"@id": "_:cs_topic_2",
+			"name": "Topic page",
+			"status": "stable",
+			"baseOnIteration": "_:iteration_topic_2",
+			"detectableBy": ".page-type-topic",
+			"layout": [
+				"Intro block",
+				"Most requested - version 1",
+				"Services and information - version 3",
+				"Context-specific features - version 5.1 [.col-md-8] + Social media channels - version 2.1 [.col-md-4]",
```
duboisp, line 240:
```suggestion
				"Context-specific features - version 5.0 [.col-md-8] + Social media channels - version 2.1 [.col-md-4]",
```

templates/topic/index.json-ld
```diff
@@ -30,17 +31,215 @@
...
+			}
+		}
+	],
+	"changesets": [
+		{
+			"@id": "_:cs_topic_2",
+			"name": "Topic page",
+			"status": "stable",
+			"baseOnIteration": "_:iteration_topic_2",
+			"detectableBy": ".page-type-topic",
```
duboisp, line 235:
```suggestion
			"detectableBy": "By the sequence of the mandatory implemented design pattern and components and limited to its optional components",
```

templates/topic/index.json-ld
```diff
@@ -30,17 +31,215 @@
...
+			"layout": [
+				"Intro block",
+				"Most requested - version 1",
+				"Services and information - version 3",
+				"Context-specific features - version 5.1 [.col-md-8] + Social media channels - version 2.1 [.col-md-4]",
+				"Contributors - version 1.0"
```
duboisp, line 241:
```suggestion
				"Intro block (mandatory)",
				"Most requested (optional) - version 1",
				"Services and information (mandatory) - version 3",
				"Context-specific features (optional) - version 5.1 [.col-md-8] + Social media channels (optional) - version 2.1 [.col-md-4]",
				"Contributors (optional) - version 1.0"
```

templates/topic/index.json-ld
```diff
@@ -30,17 +31,215 @@
...
+					"fr": { "href": "topic-fr.html", "text": "Page de sujet" }
+				},
+				{
+					"en": { "href": "demoted/topic-en.html", "text": "Topic page version 1 - Demoted" },
+					"fr": { "href": "demoted/topic-fr.html", "text": "Page de sujet version 1 - Rétrogradée" }
+				}
+			],
+			"implementation": [
+				"_:implement_topic",
+				"_:implement_topic_aem"
```
duboisp, line 114: Add an short implementation for upgrading to version 2.

Add something like that a full content review is required and the upgrade can not be automated.

components/gc-features/gc-features-doc-en.html
duboisp, line 1: Revert those change and update the working example to match the HTML structure documented

duboisp (changes requested):

duboisp (approved):
Reviewed and tested. All good.

This is a minor change to "topic" template and a content change to "gc-features"

PR #2395, opened 2024-07-11
Date modified: Fix iteration 1's code samples

EricDunsworth:
That iteration's code samples had the following issues:
* Opening tags were coded as named entities (but not closing tags)... resulting in the named entities being literally shown in the samples
* French code sample wasn't categorized properly:
  * Language was set to English
  * Description was called "Code sample with contact link" (probably copied from one of the Page Feedback Tool [PFT]'s samples)

This resolves it by re-adding real opening tags and correcting French categorizations.


Garneauma:
Pre-approved upon successful review. This is only a content change.

Garneauma (approved):

PR #2405, opened 2024-07-26
WET-469 -> gc-srvinfo -> adding example with invisible header

SebastianBurke:


Garneauma:

Garneauma (approved):

PR #2407, opened 2024-07-26
Field flow: Keep URL params matching HTML attributes

### LINKED ISSUE #2406: wb-fieldflow dropdown list removes destination page URL query parameters
## Required information

**Describe the bug**

As the title says, the wb-fieldflow list dropdown removes the URL query parameters from URLs appearing in the list.

In our case, because the language query parameter ?lang=fra is stripped out, our web server will enforce the English query parameter ?lang=eng so it is impossible to link to the French version of a page from the dropdown list. This explains why you end up on an English page using the example below.

Live example
[https://international.canada.ca/fr/analytics/test-pulldown-clicks](https://international.canada.ca/fr/analytics/test-pulldown-clicks)

```
<h2>UNSORTED LIST WITH CLASS wb-fieldflow</h2>
<div data-gc-analytics-collect="[ {&quot;value&quot;:&quot;select&quot;,&quot;emptyField&quot;: &quot;N/A&quot;}]" data-gc-analytics-formname="GAC-Test|Drop-Down:Test 2">
<div class="wb-fieldflow" data-wb-fieldflow="{&quot;inline&quot;: true}">
<p>Find the plugin for the action you need:</p>
<ul>
<li><a href="https://www.canada.ca/fr/services/emplois.html">Test Passed - Link WITHOUT query parameters in the URL</a></li>
<li><a href="https://www.international.gc.ca/transparency-transparence/strategic-environmental-assessment-evaluations-environnementales-strategiques/2022-indo-pacific-pacifique.aspx?lang=fra">Test Failed - Link WITH query parameters in the URL</a></li>
</ul>
</div>
</div>
```

**To Reproduce**

Add a URL with a query parameter like a UTM campaign code or a language parameter ?lang=fra in our case

**Expected behaviour**

Maintains query parameters of URLs

**Who shall do the work?**
I am asking for Principal publisher to please do the work

## Additional information (optional)

**Version of WET-BOEW/GCWeb you are using**

frameworks v4.0.78
theme-gcweb-v14.9.0

**Desktop/Smartphone (please complete the following information)**
- Tested on desktop, in Edge and Chrome, failed in both cases.

Thanks!

EricDunsworth:
When field flow's redirect (``redir``) action is used, submitting causes the plugin to "transform" the currently-selected dropdown ``option``'s URL parameters into hidden ``input`` elements. The ``input``s are created by passing "raw" HTML strings to the jQuery object.

That setup used to play nicely with URL parameters whose keys corresponded to the names of HTML attributes (e.g. ``lang=anything``). But it stopped working when wet-boew/wet-boew#9210 introduced DOMPurify into WET's jQuery 2.x implementation. Why? Because DOMPurify's ``sanitize()`` method filters-out ``name="[any HTML attribute name]"`` to prevent potential DOM clobbering attacks (see cure53/DOMPurify#980). End result is that jQuery ultimately returns ``name``-less ``input``s to the plugin, which in turn causes affected parameters to go missing.

This fixes it by using "pure" JavaScript (instead of jQuery) to create the ``input``s. Also adds a query string example to the redirection demo (with a mix of key naming schemes).

Fixes #2406.

### DIFF
--- a/components/wb-fieldflow/fieldflow.js
+++ b/components/wb-fieldflow/fieldflow.js
@@ -1112,7 +1112,7 @@ $document.on( "submit", selectorForm + " form", function( event ) {
 		$elm = $( elm ),
 		wbFieldFlowRegistered = $elm.data( registerJQData ),
 		wbRegisteredHidden = $elm.data( registerHdnFld ) || [],
-		$hdnField,
+		hdnField,
 		i, i_len = wbFieldFlowRegistered ? wbFieldFlowRegistered.length : 0,
 		$wbFieldFlow, fieldOrigin,
 		lstFieldFlowPostEvent = [],
@@ -1207,9 +1207,14 @@ $document.on( "submit", selectorForm + " form", function( event ) {
 						cacheName = items[ 0 ];
 						cacheParam = items[ 1 ];
 					}
-					$hdnField = $( "<input type='hidden' name='" + cacheName + "' value='" + wb.escapeAttribute( cacheParam ) + "' />" );
-					$elm.append( $hdnField );
-					wbRegisteredHidden.push( $hdnField.get( 0 ) );
+
+					hdnField = document.createElement( "input" );
+					hdnField.type = "hidden";
+					hdnField.name = cacheName;
+					hdnField.value = wb.escapeAttribute( cacheParam );
+
+					$elm.append( hdnField );
+					wbRegisteredHidden.push( hdnField );
 				}
 				$elm.data( registerHdnFld, wbRegisteredHidden );
 			}
--- a/components/wb-fieldflow/fieldflow-en.html
+++ b/components/wb-fieldflow/fieldflow-en.html
@@ -6,7 +6,7 @@ description: Transform a basic list into a selectable list.
 tag: fieldflow
 parentdir: fieldflow
 altLangPage: fieldflow-fr.html
-dateModified: 2024-04-08
+dateModified: 2024-08-27
 ---
 <ul class="list-inline">
 	<li><a class="btn btn-primary" href="fieldflow-doc-en.html">Documentation</a></li>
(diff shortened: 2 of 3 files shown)


components/wb-fieldflow/fieldflow.js
```diff
@@ -1207,9 +1207,14 @@ $document.on( "submit", selectorForm + " form", function( event ) {
...
 						cacheParam = items[ 1 ];
 					}
-					$hdnField = $( "<input type='hidden' name='" + cacheName + "' value='" + wb.escapeAttribute( cacheParam ) + "' />" );
-					$elm.append( $hdnField );
-					wbRegisteredHidden.push( $hdnField.get( 0 ) );
+
+					hdnField = document.createElement( "input" );
+					hdnField.setAttribute( "type", "hidden" );
+					hdnField.setAttribute( "name", cacheName );
+					hdnField.setAttribute( "value", wb.escapeAttribute( cacheParam ) );
```
EricDunsworth, line 1214: Should I get rid of the call to [``wb.escapeAttribute()``](https://github.com/wet-boew/wet-boew/blob/e01bc308c1cc123dbe7948cd8dccc14829ed8e70/src/core/helpers.js#L1313-L1315) at this spot?

I don't think ``setAttribute()`` technically has a need for it, but figured it can't hurt to ask before going through with it.

Garneauma:

components/wb-fieldflow/fieldflow.js
duboisp, line 1214:
```suggestion
hdnField.type = "hidden";
hdnField.name = cacheName;
hdnField.value = wb.escapeAttribute( cacheParam );
```

duboisp (changes requested):
To review and fully test.

This is a **patch**.

EricDunsworth:
FWIW I retested locally with @duboisp's suggested change and it still worked fine 💯.

EricDunsworth:
Rebased, updated modified dates and force-pushed.

This wasn't running into any merge conflicts, but since it hadn't been tested by any project maintainers yet... I figured it'd make sense to rebase to make it node 20-friendly.

Garneauma (approved):
Tested locally. Works as intended.

MatthieuBonenfant-GOC:
Thank you all for your contribution!

components/wb-fieldflow/fieldflow.js
EricDunsworth: Sorry - must've overlooked this :(.

In any case, I ended up implementing @duboisp's suggestion to set the attribute properties directly instead of using ``setAttribute()`` later on. Given that, I think it might be a safer bet to keep ``wb.escapeAttribute()`` in play.

I won't bother sending another PR to remove it.

PR #2410, opened 2024-08-01
Header: Removing main menu

Garneauma:
Updating Header component to remove the GCWeb menu.

Looking through the history of the Header component, I noticed that an extra version was added in the version history that never took place. The version 3 and 4 were done with the same commit. Not sure why they have been separated into two versions. That said, I have fixed the component history, so now the component remains at version 4.0 (current version should be 3.0).

This update to the Header meant it was necessary to edit the Breadcrumbs component. The Breadcrumb component has been updated to version 2.0.

Although the Header has been updated to a new major version, the previous version is still supported.


Garneauma:

duboisp:
Need DTO approval

PR #2402, opened 2024-07-18
Node update for compatibility with lts versions 18 and 20

Ricokola:
Update node for compatibility with lts versions 18 and 20.

PR #2411, opened 2024-08-06
ESLint: Indent switch case clauses

EricDunsworth:
Port of wet-boew/wet-boew#9793... and depends on it.

Doesn't cover "frozen" méli-mélo JS files.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,7 +15,7 @@
     "JSON": true
   },
   "rules": {
-    "indent": ["error", "tab", { "outerIIFEBody": 0 }],
+    "indent": ["error", "tab", { "outerIIFEBody": 0, "SwitchCase": 1 }],
     "eqeqeq": [2, "allow-null"],
     "no-eq-null": 2,
     "no-unused-expressions": [2, { "allowTernary": true }],
--- a/components/wb-actionmng/actionmng.js
+++ b/components/wb-actionmng/actionmng.js
@@ -383,32 +383,32 @@ var $document = wb.doc,
 
 			switch ( cache_action ) {
 
-			case "patch":
-				var ops = i_cache.patches,
-					basePntr = i_cache.base || "/";
-				if ( !ops ) {
-					ops = [ patchDefault ];
-					i_cache.cumulative = true;
-				}
-				if ( !$.isArray( ops ) ) {
-					ops = [ ops ];
-				}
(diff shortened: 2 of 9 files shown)


EricDunsworth:
Should the scope of this PR be expanded to cover méli-mélo?

The only reason I didn't cover it was because I didn't want to interfere with the ["frozen" compilation files](https://github.com/wet-boew/GCWeb/tree/master/m%C3%A9li-m%C3%A9lo/compilation-gel%C3%A9). I could've just modified the ["main" méli-mélo JS files](https://github.com/wet-boew/GCWeb/tree/master/m%C3%A9li-m%C3%A9lo), but felt like it'd be really weird to make them deviate from their "frozen" counterparts.

**PS:**
The reason the CI build is currently failing is because [GCWeb's ``eslint`` task is setup to load its WET dependency's .eslintrc file](https://github.com/wet-boew/GCWeb/blob/91947fc6f04dd22a3752ea75930f850b710aebec/Gruntfile.coffee#L1086). So even though I've updated [GCWeb's own .eslintrc file](https://github.com/wet-boew/GCWeb/blob/master/.eslintrc.json), the build won't start passing until GCWeb's WET dependency gets bumped to a newer commit that includes https://github.com/wet-boew/wet-boew/pull/9793. I'll also need to rebase this PR afterwards.

EricDunsworth:
Based on my discussion with @duboisp in the PR review meeting, I think I'll go ahead and indent the case statements in the "main" méli-mélo files too.

Apparently some of the "main" and "frozen" files already have major differences due to the former receiving non-trivial updates overtime. So I guess case indents would reasonably fall under that category of updates.

EricDunsworth:
Btw the "main" méli-mélo files are now covered in the latest force-push.

So all that's left now is for GCWeb to update its WET dependency and to rebase to get the CI build to pass.

EricDunsworth:
@duboisp @Ricokola I'll look into bumping GCWeb's WET dependency via this PR whenever wet-boew/wet-boew#9799 gets merged.

I don't think it'll be possible to bump GCWeb's WET dependency to 4.0.81.2 via any other PR (not even #2402) without causing the CI build to fail. GCWeb's build system depends on WET's ``.eslintrc.json`` file - which now _requires_ ``switch case`` indents.

Garneauma (approved):
Tested while cherry-picking this PR after resetting to PR #2402 and it works well.

PR #2413, opened 2024-08-07
update CSS for high contrast

Christopher-O:
Added new exception to remove line separator  for side by side designs in high contrast mode for firefox as cannot resolve gradient pattern issue (going through text)

### DIFF
--- a/méli-mélo/2021-05-steps/meta.md
+++ b/méli-mélo/2021-05-steps/meta.md
@@ -52,7 +52,7 @@ changes:
     description: Reduced excessive spacing for small amount of content. Removed background color for default list steps number (except in striped mode) as was not showing up properly in high contrast mode. Updated governance towards stabilization.
     departmentImpact: No change from a departmental perspective.
     publicImpact: Created a cleaner look for items using list steps with a smaller amount of content. Made the numbers more visible for edge case scenarios of print and high contrast.
- - date: 2024-08-07
+  - date: 2024-08-07
     description: Fixed issue with printing pages as some text would be cut off 
     departmentImpact: No change from a departmental perspective.
     publicImpact: Improved look of print document
--- a/méli-mélo/2021-05-steps/meta.md
+++ b/méli-mélo/2021-05-steps/meta.md
@@ -52,6 +52,9 @@ changes:
     description: Reduced excessive spacing for small amount of content. Removed background color for default list steps number (except in striped mode) as was not showing up properly in high contrast mode. Updated governance towards stabilization.
     departmentImpact: No change from a departmental perspective.
     publicImpact: Created a cleaner look for items using list steps with a smaller amount of content. Made the numbers more visible for edge case scenarios of print and high contrast.
-
+ - date: 2024-08-07
+    description: Fixed issue with printing pages as some text would be cut off 
+    departmentImpact: No change from a departmental perspective.
+    publicImpact: Improved look of print document
 output: false
 ---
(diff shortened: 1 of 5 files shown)


duboisp (dismissed):
See my previous comment.

* Need review according to our experimental checklist
* Need a description + impact description.

Christopher-O:
@duboisp   per your request

Clear and simple explanation of the change/update:

Change was to remove the side-by-side line separator in Firefox (but keep the shape and text) when in High contrast mode.  The previous pattern had the line separator "cut" through the words (and/or) which could make it slightly illegible. 

This does not affect the stacked (default) pattern, as it will still have a visible line to separate the pattern from surrounding content.

The impact on the sponsored department (CRA) for that change/update:

There is minimal impact, as we do not normally use Firefox as a browser, nor does the majority of users use high contrast mode on their devices. 


The impact on the public for that change/update:

Only those who use high contrast mode with the Firefox browser, and will be a slight improvement on readability of the text

Christopher-O:
> See my previous comment.
> 
> * Need review according to our experimental checklist
> * Need a description + impact description.

@duboisp  not sure what to put here.  This is a very minor change.  Nothing substantial that has changed from previous descriptions or review.

méli-mélo/2021-05-steps/meta.md
```diff
@@ -52,6 +52,9 @@ changes:
     description: Reduced excessive spacing for small amount of content. Removed background color for default list steps number (except in striped mode) as was not showing up properly in high contrast mode. Updated governance towards stabilization.
     departmentImpact: No change from a departmental perspective.
     publicImpact: Created a cleaner look for items using list steps with a smaller amount of content. Made the numbers more visible for edge case scenarios of print and high contrast.
-
+ - date: 2024-08-07
```
Garneauma, line 55: Add another space to pass build.
```suggestion
  - date: 2024-08-07
```

Garneauma (changes requested):
Tested and the fix for Conjunction does work. The fix for Steps also works. It seems to add a lot of spacing above the steps which looks a bit weird, but it seems intentional (`padding-top: 1em;`).

@Christopher-O Could you also add the changes description in the 
Conjunction meta.md like you did for the Steps and add the history item in the HTML for the Steps like you did for the Conjunction? Thank you!

Christopher-O:
> Only those who use high contrast mode with the Firefox browser, and will be a slight improvement on readability of the text

But a good one.  We will leave it in for now then.

Christopher-O:
> > @Garneauma The media print was added as I noticed that sometimes lines within list items would be cut off close to the border of the page and be unreadable. The padding was added as the CSS numbers created didn't adhere to the break-inside CSS class. But I have tried it on a few other patterns and without the padding, it doesn't seem to do much of a difference (a bit of splitting of the CSS numbers across two pages on the rare occasion. I would approve removing it for now (just the padding-top not the break-inside)
> 
> It's your choice. It was only a comment. :)

But a good one.  okay we will leave it in, and comment was added to meta.md and space issue was resolved.

Garneauma:
@Christopher-O I just realized that the change history is mixed with the implementation plan in `index.html`. Could you please separate them? You can do it dynamically by pulling from the metadata like the implementation plan in Steps or you can just do it static for now if you prefer.

Christopher-O:
> @Christopher-O I just realized that the change history is mixed with the implementation plan in conjunction `index.html`. Could you please separate them? You can do it dynamically by pulling from the metadata like the implementation plan in Steps or you can just do it static for now if you prefer.
> 
> Also, please add the extra space in Conjunction meta.md like in the Steps meta.md.

@Garneauma  can you confirm that I completed this correctly?

Garneauma:
@Christopher-O I have discussed with @duboisp. We will handle the `meta.md` files on our side when doing the release. So, no more work needed on your part. Thank you for doing the changes!

Garneauma (approved):

PR #2417, opened 2024-08-15
Content - gc-servinfo & gc-most-requested: content update

Garneauma:
Updating content of Services and information and Most requested components to align with DTO's own documentation.


duboisp (approved):
Reviewed and this is only a content change.

PR #2415, opened 2024-08-12
H1 thickline: Fix border rendering bug in older browsers

EricDunsworth:
Some older browsers (most notably Safari <= 15.3) incorrectly set ``border-style: solid;`` whenever the ``border-image`` property was used. That in turn caused affected browsers to render the thickline style as a three-sided border.

This works around the issue by setting the thickline style's left/right/top ``border-*-width`` properties to 0.

Thanks to @donmcdill for pitching this approach (other ways I prototyped didn't work).

Fixes #2396.

PR #2418, opened 2024-08-16
Main page title: Show breaking change in VU 2024 iteration

EricDunsworth:
The visual update 2024 iteration's breaking change previously wasn't appearing.

It seems to have been related to structure issues in the JSON-LD file's "breaking" object:
* Array using curly brackets ( ``{...}`` )
* Key/value pair placed in the array (key was named "style")

This fixes it by:
* Using square brackets ( ``[...]`` ) for the array
* Replacing the "style" key with a "Style: " prefix in the value


EricDunsworth:
Notes:
* This hasn't been tested on my end
* Came very close to sneaking this into #2415... but ultimately decided it'd be better off as its own PR to avoid adding "bloat" to the other one

duboisp (commented):
This is a content change. Pre-approved upon local testing

duboisp:
I confirm this is only a content change

PR #2422, opened 2024-09-05
Clipboard plugin (méli-mélo): Restore space indents

EricDunsworth:
#2411 accidentally-introduced tab indents into the plugin's JS file.

It was caused by a clash of EditorConfig settings (which apply GCWeb's basic linting settings to all méli-mélo files) vs ESLint (which excludes méli-mélo). The added indents adhered to the EditorConfig settings, but the file itself was using a different kind of indent style (groups of 4 spaces).

This changes the tab characters back into spaces.


Garneauma:
Pre-approved upon successful review. Also will need to check the warning about "The head ref may contain hidden characters: "clipboard-m\u00E9li-m\u00E9lo-restore-space-indents"

Garneauma (approved):

PR #2423, opened 2024-09-10
Release v15.6.1

ouafaaetta:
Release v15.6.1


Garneauma (dismissed):

Garneauma (changes requested):
The build doesn't seem to pass because it installs version 4.0.83.3 of WET.

Can you try to update line 9 and line 13 of package-lock.json to the current versions of the projects?

ouafaaetta:
> The build doesn't seem to pass because it installs version 4.0.83.3 of WET.
> 
> Can you try to update line 9 and line 13 of package-lock.json to the current versions of the projects?

The requested change is done

Garneauma (approved):

PR #2434, opened 2024-10-21
GC Home Page: Reduced Doormats and added all services button

BrahimMahadi:
### This pull request includes modifications for the GC home page to reduce number of Doormats and to add All Services button for DTO A/B testing
General checklist
- [x] Updated the GC home page to reduce number of Doormats from 15 to 9.
- [x] Added all services button.
- [x] Changes were made to both the English and French pages.
- [x] Reference JIRA ticket WET-489.


BrahimMahadi:
@duboisp This PR is ready for review

templates/home/home-en.html
```diff
@@ -65,48 +65,28 @@ <h3><a href="https://www.canada.ca/en/services/business.html">Business and indus
...
 				<h3><a href="https://www.canada.ca/en/services/finance.html">Money and finances</a></h3>
 				<p>Personal finance, credit reports, fraud protection, paying for education</p>
 			</div>
-			<div class="col-lg-4 col-md-6">
-				<h3><a href="https://www.canada.ca/en/services/science.html">Science and innovation</a></h3>
-				<p>Scientific research on health, environment and space, grants and funding</p>
-			</div>
+		</div>
+		<div class="mrgn-top-sm" style="max-width: 140px;">
+			<p style="font-weight: bold; margin-top:20px; margin-bottom:20px; border: 2px solid #26374a; padding: .75em; border-radius: 5px; text-align: center;"><a style="text-decoration: none !important; color:#26374a !important;" href="https://www.canada.ca/en/services.html">All services</a></p>
```
duboisp, line 86: No inline style, please add the CSS in the SASS file for the home page.

See there is one SCSS per view port here: https://github.com/wet-boew/GCWeb/tree/master/templates/home

The _base.scss is the default CSS applied for the home page.

You can learn about SASS here: https://sass-lang.com/documentation/syntax/

duboisp (changes requested):
Please see the requested change, I didn't compiled nor tested.

Add the CSS in the proper SCSS file.

templates/home/_base.scss
```diff
@@ -50,6 +50,26 @@
 		margin: 1em 0px;
 	}
 
+	// All services button
+	.all-services-btn {
```
duboisp, line 54: FYI - 
See CSS used for the super task button here: https://github.com/wet-boew/GCWeb/blob/master/sites/baseline/_buttons.scss#L16
And here a working example/code sample: https://github.com/wet-boew/GCWeb/blob/master/templates/content-page/content-en.md?plain=1#L24

duboisp (changes requested):

templates/home/home-en.html
BrahimMahadi: Removed inline css

templates/home/_base.scss
BrahimMahadi: I removed the custom css and used the call-to-action button class for styling.

templates/home/home-en.html
```diff
@@ -65,48 +65,28 @@ <h3><a href="https://www.canada.ca/en/services/business.html">Business and indus
...
 				<h3><a href="https://www.canada.ca/en/services/finance.html">Money and finances</a></h3>
 				<p>Personal finance, credit reports, fraud protection, paying for education</p>
 			</div>
-			<div class="col-lg-4 col-md-6">
-				<h3><a href="https://www.canada.ca/en/services/science.html">Science and innovation</a></h3>
-				<p>Scientific research on health, environment and space, grants and funding</p>
-			</div>
+		</div>
+		<div>
+			<a class="btn btn-call-to-action" href="https://www.canada.ca/en/services.html">All services</a>
```
duboisp, line 86: I didn't say to use the call-to-action, my point was more do an implementation similar like the call to action which will achieve the desired behaviour and visual effect as illustrated by their tested prototype.

If it is easier, you can extract each desired behaviour/style from their prototype and log that as individual Technical Requirement in the Jira request like Sous toute reserve et seulement pour présenter le principe:  "TR-1: button style: border 2px, padding 3px...." "TR-2: button behaviour: ensure the button always remain on its own row" (I didn't check the desired effect of the use of max-width)

duboisp (changes requested):
@Garneauma FYI

templates/home/test.css
duboisp, line 1: This CSS should be included in the base.scss and need to be unique for the home page.
The home page do use a global CSS class named "home", so the above CSS selector will look like this once rendered by SASS : `.home .btn-all-services`

templates/home/home-fr.html
```diff
@@ -63,46 +63,25 @@ <h3><a href="https://www.canada.ca/fr/services/entreprises.html">Entreprises et
...
 				<h3><a href="https://www.canada.ca/fr/services/finance.html">Argent et finances</a></h3>
 				<p>Finances personnelles, dossiers de crédit, se protéger de la fraude, financer votre éducation</p>
 			</div>
-			<div class="col-lg-4 col-md-6">
-				<h3><a href="https://www.canada.ca/fr/services/science.html">Science et innovation</a></h3>
-				<p>Recherche scientifique sur la santé, l'environnement et l'espace, subventions et financement</p>
-			</div>
+		</div>
+		<div>
+			<a class="btn btn-call-to-action" href="https://www.canada.ca/fr/services.html">Tous les services</a>
```
duboisp, line 84: HTML show the use of the newly added CSS style

```suggestion
			<a class="btn btn-all-services" href="https://www.canada.ca/fr/services.html">Tous les services</a>
```

duboisp (changes requested):

templates/home/_base.scss
```diff
@@ -77,4 +76,22 @@
...
 	}
+
+	// All services button under services and information
+	.home-all-services {
+		margin-bottom: 20px;
+		margin-top: 20px;
+
+		.btn {
+			border: 2px solid #26374a;
+			color:#26374a;
```
duboisp, line 87:
```suggestion
			color: #26374a;
```

duboisp (changes requested):

duboisp (approved):
This code state is approved from a functional point of view. I reviewed and tested locally. We can use that state as 

Documentation still required prior to merge

duboisp (changes requested):
Need to produce the technical documentation

duboisp (approved):
Review and tested, do work as expected

This is a Patch

PR #2439, opened 2024-10-25
Build: updating build dependencies

Garneauma:
Updating build dependencies to synchronize with WET-BOEW.

Dependent on wet-boew/wet-boew #9820 being merged.


duboisp (changes requested):
Reviewed and do work as expected when building with the WET-BOEW PR 9820. Almost ready to be approved.

Request changes only because the "package-lock.json" would need to be fully rebuild once the wet-boew along with the updated dependencies are released and this PR do reference that new wet-boew version.

Garneauma:
@duboisp The PR has been updated to use the latest version of WET-BOEW and I have installed all the new dependencies.

duboisp (approved):
Reviewed and tested locally. 

Not with the last package-lock but I validated that the updated wet-boew dependencies was also listed there. I confirm it is there.

PR #2444, opened 2024-11-08
Update _base.scss

Christopher-O:
Clear and simple explanation of the change/update:

Updated input elements to be invisible simply by using opacity: 0 and z-index:2 in order for inputs to be selected by Dragon Naturally Speaking adaptive technology voice reader

Added fallback to show focus of checkboxes and radio buttons and selection of radio buttons in Windows high contrast mode. 

The impact on the sponsored department (CRA) for that change/update:

Improved accessibility for users of the adaptive technology or visual mode (High contrast)

The impact on the public for that change/update:

No change for current use cases, but allows easier selection of radio button and checkboxes using Dragon Naturally Speaking, and provides proper focus in high contrast mode when using keyboard and selecting radio buttons

### DIFF
--- a/components/gc-chckbxrdio/_base.scss
+++ b/components/gc-chckbxrdio/_base.scss
@@ -38,6 +38,7 @@ fieldset.gc-chckbxrdio {
 
 	input[type="radio"],
 	input[type="checkbox"] {
+		margin-left: 10px;
 		opacity: 0;
 		z-index: 2;
 
--- a/components/gc-chckbxrdio/_base.scss
+++ b/components/gc-chckbxrdio/_base.scss
@@ -145,7 +145,7 @@ fieldset.gc-chckbxrdio {
 	}
 }
 @media (prefers-contrast: more) {
-	.gc-chckbxrdio-provisional {
+	.gc-chckbxrdio {
 		input[type=checkbox] {
 			&:focus+label {
 				&::before {
--- a/components/gc-chckbxrdio/_base.scss
+++ b/components/gc-chckbxrdio/_base.scss
@@ -38,14 +38,8 @@ fieldset.gc-chckbxrdio {
 
 	input[type="radio"],
 	input[type="checkbox"] {
-		border: 0;
-		clip: rect(0 0 0 0);
-		height: 1px;
-		margin: -1px;
-		overflow: hidden;
-		padding: 0;
-		position: absolute;
-		width: 1px;
+		opacity: 0;
+		z-index: 2;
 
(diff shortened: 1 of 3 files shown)


duboisp:
Pre-approved upon review and testing.

components/gc-chckbxrdio/_base.scss
```diff
@@ -38,14 +38,8 @@ fieldset.gc-chckbxrdio {
...
-		border: 0;
-		clip: rect(0 0 0 0);
-		height: 1px;
-		margin: -1px;
-		overflow: hidden;
-		padding: 0;
-		position: absolute;
-		width: 1px;
+		opacity: 0;
+		z-index: 2;
```
Garneauma, line 42: @Christopher-O Was there a reason for this change? It makes it possible for users to select the checkbox when clicking on the white space beside the checkbox. This could confuse users.

Christopher-O: It's also to allow Dragon Naturally Speak users to select these items.  The current design does not allow users to select the checkbox or radio buttons.   All it does is increase the clickable area but it still close to each item so it should not be an issue.  If you would prefer I can update the positioning of the inputs when vertically aligned so they are more centred within the clickable area.

Garneauma: I would add `margin-left: 0;`. Right now, when inside a `.checkbox` element, it adds a `margin-left: -20px` to the input.

Christopher-O: I added margin-left: 10px, as margin-left: 0px did not have the input on top of the small checkbox design, so this way it the input is now inside the enlarged design for both sizes.

Garneauma:
To do by the WET-BOEW team: contribute to this pull request by updating the API documentation so that it reflects this change.

components/gc-chckbxrdio/_base.scss
Garneauma: @Christopher-O Hmmm, it messes up the hover styling since it is over the large checkbox. I don't think we should impede on the user experience of the vast majority simply because one software is not able to do its job. I would suggest to revert those changes unless there is a way to place it behind the large checkbox and still being accessible by Dragon Naturally Speak.

@duboisp Would you like to chime in?

Garneauma: I have pushed a fix to support this new checkbox position without impacting the hover behaviour. I have also added the documentation. Thank you for your contribution @Christopher-O !

PR #2446, opened 2024-11-19
Most requested: display items in columns

Garneauma:
This PR changes the distribution of Most requested items from top to bottom and from left to right instead of from left to right and top to bottom.

Before:
1 2
3 4
5 6

After:
1 4
2 5
3 6

Changes related to WET-488

### DIFF
--- a/components/gc-most-requested/index.json-ld
+++ b/components/gc-most-requested/index.json-ld
@@ -14,10 +14,10 @@
 		"en": "Features top tasks related to the page it is on.",
 		"fr": "Comprend des tâches principales liées à la page sur laquelle elles se trouvent."
 	},
-	"modified": "2024-02-14",
+	"modified": "2025-05-08",
 	"componentName": "gc-most-requested",
 	"status": "stable",
-	"version": "1.0",
+	"version": "1.1.0",
 	"pages": {
 		"docs": [
 			{
@@ -70,7 +70,7 @@
 				"en": "https://design.canada.ca/common-design-patterns/most-requested.html",
 				"fr": "https://conception.canada.ca/configurations-conception-communes/en-demande.html"
 			},
-			"iteration": "_:iteration_mostrequested_1",
+			"iteration": "_:iteration_mostrequested_2",
 			"example": [
 				{
 					"en": { "href": "gc-most-requested-en.html", "text": "Most requested" },
@@ -86,6 +86,10 @@
 			],
 			"history": [
 				{
+					"en": "May 2025 - Items are now distributed vertically instead of horizontally.",
+					"fr": "Mai 2025 - Les items sont maintenant distribués verticalement plutôt que horizontalement."
+				},
+				{
 					"en": "February 2024 - Initial implementation of the variation.",
 					"fr": "Février 2024 - Implémentation initiale de la variante."
 				},
@@ -99,7 +103,7 @@
 	"implementation": [
 		{
 			"@id": "_:implement_mostrequested",
-			"iteration": "_:iteration_mostrequested_1",
+			"iteration": "_:iteration_mostrequested_2",
 			"name": {
 				"en": "Standard",
 				"fr": "Standard"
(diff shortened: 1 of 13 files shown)


components/gc-most-requested/_screen-lg-min.scss
```diff
@@ -0,0 +1,11 @@
+/*
+	Most requested component
+*/
+
+// Make sure the component is not inside a container
+main:not(.container) .gc-most-requested,
+div:not(.container) .gc-most-requested {
```
duboisp, line 7: This don't seem to work as expected. I moved the "gc-most-requested" inside a component and the style was still applied.

Also, I am unable to observe the effect of that style in FF neither in Chrome. What it does? Should we add an additional working example for that edge case?

duboisp, line 7:
```suggestion
:not(.container) .gc-most-requested {
```

duboisp (changes requested):

components/gc-most-requested/_screen-lg-min.scss
duboisp: Alternatively, you can add a style to break bad implementation like undo:
* the h2 float
* col-count set to 1
* No background color.

CSS selector sample: `.container .gc-most-request`

duboisp (changes requested):

Garneauma:
> **This change is very perceptible** and I do evaluate if as a major change because of the visual change. Additional communication might be required prior to apply this change.

In my opinion this is more a minor change than a major one. "Minor change: The change are noticeable and complement, add to the user experience and remain at least understandable as it was prior." I don't believe a user would be lost because some items have changed in order. It might be even more understandable for a user now, and easier to coomprehend for an author. We can discuss this further.

duboisp:
We received a confirmation to move ahead. Yes this change is perceptible and an user will notice that change but it don't change the component essential behaviour. So this is a Minor change.

Note: I am going to do a last PR review.

components/gc-most-requested/index.json-ld
```diff
@@ -145,12 +149,31 @@
...
 	"iteration": [
+		{
+			"@id": "_:iteration_mostrequested_2",
+			"name": "Most requested - Iteration 2",
+			"date": "2024-11",
+			"detectableBy": ".gc-most-requested:not(.provisional)",
+			"predecessor": "_:iteration_mostrequested_1",
+			"fixes": [
+				"Changed display of items from flex to column-count."
+			],
```
duboisp, line 160: Can you change this for an "addition" and append the justification that I wrote in the Github PR about why this is a mino change. Thanks

components/gc-most-requested/index.json-ld
```diff
@@ -85,6 +85,10 @@
 				"_:implement_mostrequested"
 			],
 			"history": [
+				{
+					"en": "November 2024 - Items are now distributed vertically instead of horizontally.",
+					"fr": "Novembre 2024 - Les items sont maintenant distribués verticalement plutôt que horizontalement."
```
duboisp, line 90: Let's update that date for like May 2025 which is closer to the reality, or you can add an additional item and prepend "Draft" to this item

duboisp (changes requested):
Reviewed and tested, do work as expected. There is only a little adjustment to do at the documentation.

Review checklist:

* [x] feature integration - look good
* [x] documentation - All good except the following:
  * [ ] Change/update the history date to be closer to today
  * [ ] Describe/add the minor change justification
* [x] code - Completed and look good
* [x] technical test - Completed and pass
* [x] functional check - Completed
* [x] versioning impact - This is a minor change for the most-requested component
* [x] maintenance plan - Nothing to note
* [x] Previous comment/concerns - Was all addressed

duboisp (approved):
The documentation change was applied as requested.

The CSS/JS don't seem to have modified since my last review, so I maintain my previous review as is. I did a quick functional check and it still work as expected.

PR #2450, opened 2024-12-02
Main: fix word-break issue for long words in small screens

Garneauma:


duboisp (approved):
Reviewed and manually tested a very long word that was overflowing outside the main container. It does work as expected.

Garneauma:
@duboisp Working example added.

PR #2457, opened 2024-12-19
Build: Replace Node Sass with Dart Sass

EricDunsworth:
Minimum changes needed to get it to successfully build.

Contains countless deprecation warnings, but those can be dealt with in the future.

GCWeb port of wet-boew/wet-boew#9840.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1,6 +1,6 @@
 #global module:false
 path = require("path")
-sass = require("node-sass")
+sass = require("sass")
 yaml = require("js-yaml")
 
 module.exports = (grunt) ->
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -503,11 +503,11 @@ module.exports = (grunt) ->
 			expectedFull = this.data[1]
 			expected = expectedFull.substring(expectedFull.indexOf("#") + 2)
 
-			if (installed != expected)
-				grunt.log.writeln(">> "['red'] + "wet-boew: installed: " + installed['red'] + ", expected: " + expected['green'])
-				grunt.fail.fatal(">> "['red'] + "Invoke " + "npm install"['green'] + " to update WET-BOEW version")
-			else
-				grunt.log.writeln(">> "['green'] + "WET-BOEW version up to date with package's version.")
+			#if (installed != expected)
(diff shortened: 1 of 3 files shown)


EricDunsworth:
Marking this as a draft since I haven't tried going over its outputted CSS at all and since the second commit technically makes it unmerge-able (only point of it was to prove a GCWeb+WET Dart Sass setup is able to successfully build).

Gruntfile.coffee
```diff
@@ -503,11 +503,11 @@ module.exports = (grunt) ->
...
-			if (installed != expected)
-				grunt.log.writeln(">> "['red'] + "wet-boew: installed: " + installed['red'] + ", expected: " + expected['green'])
-				grunt.fail.fatal(">> "['red'] + "Invoke " + "npm install"['green'] + " to update WET-BOEW version")
-			else
-				grunt.log.writeln(">> "['green'] + "WET-BOEW version up to date with package's version.")
+			#if (installed != expected)
+			#	grunt.log.writeln(">> "['red'] + "wet-boew: installed: " + installed['red'] + ", expected: " + expected['green'])
+			#	grunt.fail.fatal(">> "['red'] + "Invoke " + "npm install"['green'] + " to update WET-BOEW version")
+			#else
+			grunt.log.writeln(">> "['green'] + "WET-BOEW version up to date with package's version.")
```
duboisp, line 510: Note: You can execute > `grunt build` which don't check for the wet-boew version.

duboisp:
I tried to build on a fresh system (`npm install` and `npm ci`) and I can't because I am getting an error related to "PhantomJS not found on PATH"

But it does work on my other system. I am able to successfully compile with `npm install`.
I did a spot functional test and and the site seem to be rendered as expected.

cc @Garneauma

Garneauma:

Gruntfile.coffee
EricDunsworth: Good catch :).

Although one of my goals with that change was to show-off that GCWeb CI's build would pass too :P.

Although in retrospect, I think I overdid it with the commented-out lines. Only really needed to comment-out ``grunt.fail.fatal(...)`` in practice...

Garneauma (approved):

PR #2461, opened 2025-01-09
All services: Adding expand/collapse all buttons

Garneauma:


templates/all-services/index.json-ld
```diff
@@ -102,6 +102,13 @@
 		}
 	],
 	"iteration": [
+		{
+			"@id": "_:iteration_allservices_2",
+			"name": "All services page - Iteration 2",
+			"date": "2025-01",
+			"additions": [ "Adding expand/collapse all buttons." ],
+			"detectableBy": "Three navigation bands followed by a series of <details> elements for all goc themes and topics."
```
duboisp, line 110: Add a mention about the expand/collapse button which is what makes it different from iteration 1.

duboisp (changes requested):
Review and tested locally, do work as expected.

This is a patch change because this is a "limited use" template which is fully manged by PP.

Once the precision is completed in the documentation, we will be able to merge this.

duboisp (approved):

PR #2464, opened 2025-01-17
If a column was invisible in the Datatable, the wrong Column target was used

bozzit:
Workaround was to put all invisible columns as the last columns of the table, but this fix should take care of the issue


duboisp:
Pre-approved upon review

duboisp:
@bozzit Can you comment here the answer at the following question that describe your change (what + impact):

* Clear and simple explanation of the change/update;
* The impact on the sponsored department for that change/update;
* The impact on the public for that change/update.

bozzit:
- Change was to fix a bug with tables that have hidden columns.  The hidden columns caused the formatting to apply to the wrong column
- The bug is now fixed and the workaround of having to move the hidden columns to the end is no longer necessary
- The bug is now fixed and the workaround of having to move the hidden columns to the end is no longer necessary

duboisp (changes requested):
Missing a working example that show the functional aspect of that code addition.

May be one column for one of the tables in the existing working example can be set to invisible.

duboisp:
I added the working example. I reviewed and look good.

With the updated working example, I was able to confirm the added code get executed as expected.

@bozzit can you confirm that you tested the code you added. Once I have that we will be able to merge it.

bozzit:
@duboisp Yes I developed and tested the fix on a local copy of the datatable plugin here at PSPC

duboisp (approved):
Reviewed and approved

PR #2477, opened 2025-02-21
News: update news template

Garneauma:
Changes related to WET-522.


duboisp (approved):
Review and this is a content change

PR #2482, opened 2025-02-26
Feedback: hide PFT when server is down

Garneauma:


Garneauma:

PR #2496, opened 2025-03-04
Github actions: remove unnecessary refresh of pages

Garneauma:
Pages are already refreshed when deploying.


duboisp (approved):
looks good.

I confirm the gh-pages are already configured in the repository setting.

This is a build change only

PR #2518, opened 2025-03-12
Build: adjust stylelint to ignore local build files

Garneauma:


duboisp (approved):
Looks good,

this is only an update to our site generated content which was already applied and a fix to our build.

This has no impact on the distributed product according to our versioning API. It is just a build change.

PR #2536, opened 2025-03-31
GCWeb Jekyll: adding possibility to include GCDS in global proprties

Garneauma:
In `_config.yml`, Set: `gcds: true` in the `global` property to activate GCDS globally. 

Optionally, set `gcds: [version number]` to define a specific version of GCDS. For example: `gcds: "0.33.0"`.

Finally, you can set `gcds: false` to deactivate GCDS, or simply remove the property.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -47,6 +47,12 @@ global:
   applicationURL:
     en: "#"
     fr: "#"
+# Include GCDS component libraries, excluding its CSS utilities
+# - (default) unspecified: GCDS librarires are not going to be load
+# - true: Will use the latest GCDS libraries
+# - String: The value must represent a specific GCDS version number, when specified it will use that version number
+#  gcds: "0.33.0"
+
 #
 # Override include to use
 # includes:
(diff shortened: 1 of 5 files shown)


duboisp (changes requested):
I tested with all the range of possible value and with a few GCDS version to see if we can emulate backward compatibility check.

It does work as expected. We will be able to merge once the configuration adjustment is completed.

This addition only impact the GCWeb Jekyll template product which is not version controlled yet.

duboisp (changes requested):
Review and tested, the GCDS integration work as expected. That will allow us to spin up test website and run some automated testing prior to upgrade/change the GCDS version we use.

But there is a second commit that need to be removed.

Garneauma:
@duboisp Removed the other commit. Sorry about that.

duboisp (approved):
Reviewed and tested.

Work as expected.

This has no impact to our GCWeb product, only an addition to our GCWeb Jekyll template which we use for testing and demos purpose.

PR #2538, opened 2025-04-03
Brand: updating logo red to #EB2D37

Garneauma:
Changing red of Canada branding from #F00 to #EB2D37.

related to wet-531


duboisp (approved):
Reviewed and look good.

This is a patch change, The change are barely visually noticeable and don't impact how an user are interacting with those images.

PR #2541, opened 2025-04-15
Search: adding site search documentation

Garneauma:
Changes related to WET-536

### DIFF
--- a/sites/search/search-doc-en.html
+++ b/sites/search/search-doc-en.html
@@ -1,0 +1,64 @@
+---
+{
+	"altLangPage": "search-doc-fr.html",
+	"dateModified": "2025-04-14",
+	"description": "Documentation on how to implement the site search box",
+	"language": "en",
+	"title": "Site search box"
+}
+---
+<span class="wb-prettify all-pre hide"></span>
+
+<dl class="horizontal">
+	<dt>Status</dt>
+	<dd>Stable</dd>
+	<dt>Type</dt>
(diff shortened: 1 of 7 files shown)


sites/search/search-doc-en.html
```diff
@@ -0,0 +1,64 @@
...
+	<dt>Component version</dt>
+	<dd>Version 1.0</dd>
+</dl>
+
+<h2>Purpose</h2>
+<p>The site search box is an element of the global header. It allows people to search Government of Canada content using a simple search field.</p>
+<p>Results for site search are at the level of all Government of Canada web content, or a subset of content at the department or agency level.</p>
+
+<h2>How to implement</h2>
+<p>Refer to the below code samples.</p>
```
duboisp, line 28: We will need to document another variant for the top right search and we will need to demote this current pattern. 

The hidden text for the search button must match the hidden label. So instead of "Search" it need to be "Search Canada.ca". See our internal ticket WET-230 for more info.

duboisp (changes requested):
See inline comment.

* [ ] Create a new variant and demote the current one. Add also the migration instruction to switch from the demote one toward the new variant.

duboisp (commented):
I reviewed and it looks good

Note: The accessibility concern was review and we can consider it addressed.

duboisp (changes requested):

duboisp (approved):
Review completed and tested locally.

This is only a content change that formally provide implementation documentation for the existing top right search.

PR #2543, opened 2025-04-23
GC-Subway: adding mobile-specific links to index page

Garneauma:
Duplicating links in index page just like in sections page so that in mobile view they anchor to the content.


Garneauma:

PR #2547, opened 2025-04-29
Build: adding workflow to build GCWeb using gh-pages

Garneauma:


duboisp (changes requested):
Documentation are missing

At least a quick debrief about the required configuration

docs/developing-en.md
```diff
@@ -124,6 +124,44 @@ Where:
...
+5. Under "Environment secrets", select "Add environment secret":
+   - **Name**: my_secret
+   - **Value**: [Paste the token you copied in step 1.7]
+
+### 3. Setup Github Pages
+To be able to view GCWeb through Github, you'll have to setup your repo to use Github pages. Here's how to do so:
+1. In your GCWeb fork, go to the Settings tab.
+2. In the Settings page, go to the Pages tab.
+3. **Source**: deploy from a branch
+4. **Branch**: gh-pages | /(root) (if you don't have the `gh-pages` branch already, you'll have to create it)
```
duboisp, line 154:
```suggestion
4. **Branch**: gh-pages | /(root) (if you don't have the `gh-pages` branch already, you'll have to create it or it will be created after the first run)
```

duboisp (approved):
Tested and work as expected.

Did you wanted to fix the link to the wet-boew working example too?

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,83 @@
...
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install Bower and Grunt
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
```
nschonni, line 36: This would replace the previous "Cache node modules" step.
```suggestion
          node-version: ${{ matrix.node-version }}
          cache: npm
```

nschonni (commented):
The NPM install should also use `npm ci`

.github/workflows/gh-pages.yml
Garneauma: @nschonni Yeah, I was trying some stuff as this solution was still taking a lot of time to install dependencies. After many trial and errors, I realized that the issue was with WET-BOEW and not the dependencies installation. I have reverted back to using `cache: npm`. I will be working on changing the "prepare" action from WET-BOEW as it runs many unnecessary tasks unrelated to building the dist files which take a lot of time. It should reduce the installation time by about 2-2.5min.

Garneauma:
> FYI, there is an Action from GitHub that might be better to use https://github.com/actions/deploy-pages

@nschonni Not sure how this action would replace this workflow. Would you mind explaining further? I would love to optimize this, if possible. :)

nschonni:
> > FYI, there is an Action from GitHub that might be better to use https://github.com/actions/deploy-pages
> 
> @nschonni Not sure how this action would replace this workflow. I guess it would replace the last 2 actions? Would you mind explaining further? I would love to optimize this, if possible. :)

You got it right, it doesn't replace this, but it might "simplify" the token/environment parts since it uses built-in tokens. I think you need to use the upload action https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site instead of the commit step too

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,77 @@
...
+
+on:
+  push:
+    branches: dev
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    environment: dev
```
nschonni, line 11: I think the GitHub action will create a `github-pages` one from a quick skim of the docs

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,77 @@
...
+      #   uses: ad-m/github-push-action@v0.8.0
+      #   with:
+      #     branch: gh-pages
+      #     force: true
+      #     github_token: ${{ secrets.my_token }}
+
+      - name: Upload static files as artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: /
```
nschonni, line 64: Note sure if this just needs to be the `jekyll-dist` or something smaller

Garneauma: @nschonni I'm trying to figure out which path to set. I think I might need to build Jekyll first instead of doing it in the gh-pages.

Garneauma: @nschonni Since this is a time sensitive issue, I'll stick with the current process and revise at a later time.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Build GCWeb
+        run: grunt build-gh-pages
+
+      # Overwrite all layout and includes and remove dist from .gitignore
+      - name: Update GCWeb Jekyll theme files
+        run: |
+          cp -R ~jekyll-dist/_includes/* _includes/
+          cp -R ~jekyll-dist/_layouts/* _layouts/
+          sed -i '/^dist\/$/s/^/#/' .gitignore
```
duboisp, line 45: We don't have the wet-boew-demos working example copied over. Not a blocker as you can do that at the same time of when you will fix the relative link to it.

duboisp (approved):
Reviewed and looks good.

This is a "Build" change - no impact 

* [x] Feature integration
* [x] Documentation
* [x] Review all the code
* [x] Technical test 
* [x] Functional check
* [x] Versioning impact.
* [x] Work items completion

PR #2550, opened 2025-05-01
GCWeb Jekyll: fixing demos links

Garneauma:


wet-boew/wb-randomize/index.json-ld
duboisp, line 1: Note - to be addressed later: L'exemple aléatoire avec les images ne fonctionne pas.

(Les liens sont ok)

wet-boew/zebra/index.json-ld
duboisp, line 1: Note - to be addressed later: L'exemple "Plusieurs groupe" en français manque une espace dans la 5e ligne, afin qu'il soit render comme en anglais.

(Les liens sont ok)

duboisp (changes requested):
Tested locally and in my fork, it is working as expected. I also reviewed/tested each modified content.

This is only a content and build change.

Although I found a issue with the links in the index page that require an immediate fix.

I identified a few working example that need some love and will need some adjustment that could happen via another PR. @Ricokola for your attention and can you ensure they will get addressed eventually. Thanks.

wet-boew/add-cal/index.json-ld
```diff
@@ -22,12 +22,12 @@
 			{
 				"title": "Add to Calendar",
 				"language": "en",
-				"url": "https://wet-boew.github.io/gcweb-compiled-demos/wetboew-demos/add-cal/add-cal-en.html"
+				"wetboew": "add-cal/add-cal-en.html"
```
nschonni, line 25: Changing `url` to `wetboew` in a bunch of these seems like a bad find/replace

duboisp: That property is currently only used by our Jekyll site and it point to a special location inside GCWeb build where we are copying all the wet-boew working example over to GCWeb. That allow us to test easily the wet-boew example inside the GCWeb theme locally.

PR #2558, opened 2025-05-15
Footer: align with GCDS

Garneauma:
Fixed footer alignment with GCDS.


duboisp (approved):
Reviewed and tested locally

It works as expected and it render as expected.

PR #2566, opened 2025-05-26
Move Features section below search results on news template for GCWeb

mmoya064:
Move Features section below search results on news template for GCWeb
For WET-552 ticket


Garneauma (dismissed):
The section is inside the "col-md-9". It should be outside of it. After the "row" element.

BrahimMahadi (changes requested):

PR #2577, opened 2025-06-06
Thematic: Choose Canada - remove nested CSS

Garneauma:
Nested CSS are ignored when passed through our minifier. I have moved the nested CSS styles out of their parent and used regular CSS instead.

### DIFF
--- a/méli-mélo/th-choose-canada/style.css
+++ b/méli-mélo/th-choose-canada/style.css
@@ -1,31 +1,31 @@
 .choose-canada-banner {
 	background-image: url("https://canada.ca/content/dam/pch/images/campaigns/choose-canada/Banner1920x315.jpg");
+}
 
-	.well {
-		background-image: linear-gradient(to right, #8C0D13, #BB1119);
-		border: 0;
-		border-radius: 0;
-		color: #fff;
-		margin-bottom: 0;
-		padding: 15px;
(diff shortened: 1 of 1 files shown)


duboisp:
Pre-approved once change are completed + review + tested.

Garneauma:
> Btw CSS selectors nesting is widely-supported is modern browsers ([Can I use... considers currently lists it as a baseline feature as of 2023 with 90.61% browser support](https://caniuse.com/css-nesting)).
> 
> So... there might be value in updating or replacing that minifier at some point.

@EricDunsworth I totally agree. We'll have to look at updating our CSS minifier.

duboisp (approved):
Reviewed and tested the minified version.

It's work as expected. Although I find out the images included in the working example are not copied over inside the minified version here: /méli-mélo-demos/gc-thématique/th-choose-canada/choose-canada.html But however, we assume the image path set in the bg-srcset are going to be updated to point toward a valid asset, so this is not an isssue.

Thanks @Garneauma 

---
Update Choose Canada thématique
**Description**: Update CSS selectors to ensure they remain functional after minification.
**Impact on the sponsor department**: Enables removal of the temporary workaround on their promotion page.
**Impact on the public:** None. The experience will remain exactly the same.

PR #2581, opened 2025-06-13
Search: adjusting font-size of h3 and paragraphs

mmoya064:
Search: adjusting font-size of h3 and paragraphs for search results working examples and documentation
Related to WET-557


Garneauma:
Putting this PR on draft as the business requirements still need to be defined.

templates/_search.scss
```diff
@@ -81,4 +81,9 @@
 		padding-bottom: 8px;
 		padding-top: 8px;
 	}
+
+	.paragraph-font-size {
+		font-size: 18px;
+	}
```
duboisp, line 87: The search page should already have a global unique css class associated with it. I think it is "page-type-search".

Can we avoid to use an utility class here to avoid to re-open and re-edit all search result page? 

Also, I know our reference implementation don't match the current implementation of the search result managed by the search-ui application. Prior to merge this, we must ensure that both are matching, adjustment can be made in either side as long it pass WCAG and match the expected design + structure.

PR #2582, opened 2025-06-13
Global: deprecating demoted versions

Garneauma:
- Deprecating demoted versions of GCWeb components
- Removing elements that have been deprecated for more than 1 year


Garneauma:

duboisp (approved):
Component impacted:

* Major - Byline - Feature deprecated exception when context of use is a news article
* Major - Contributor - Context of use changed to be limited inside the page details section, along with the date modified.
* Major - GC Feedback - Deprecated of the initial ajax fragment version which didn't included the PII check.
* Major - Site footer - Removed CSS that was used by deprecated pattern prior GCWeb v13.0.0
* Major - Site footer - Deprecated version 8.1 and 8.0
* Major - Page details - Deprecated version 3.0, 3.1 and 3.2
* Major - Template, Institutional landing page - Removal of deprecated template v1
* Major - Template, Topic - Deprecate version 2.0.0 (should use v2.2 instead)

Todo:
* [ ] Add documentation for byline regarding its exception - A new variation should be added. But let's keep "Deprecated" to avoid miss use.

Check and adjust if needed:
* [ ] Ensure the asset in the sub-folder "deprecated" are not part of our distribution. For example, this file should not be there: assets/deprecated/page-feedback-v1-en.html

---
Reviewed and tested locally

PR #2587, opened 2025-06-17
Static header: adding a visible border to the search bar

mmoya064:
Static header: adding a visible border to the search bar
Related to WET-264


sites/search/index.json-ld
```diff
@@ -1,6 +1,6 @@
 {
 	"@context": {
-		"@version": 1.1,
+		"@version": 1.2,
```
Garneauma: Do not change this, it's the version of the terms context. It's not related to the component itself. :)

sites/search/_base.scss
```diff
@@ -40,6 +40,11 @@
 		box-shadow: none;
 		color: #555;
 		position: relative;
+
+		:active {
```
Garneauma: Pretty sure this would not do anything. Have you tested it?

sites/search/_base.scss
```diff
@@ -40,6 +40,14 @@
 		box-shadow: none;
 		color: #555;
 		position: relative;
+
+		&:focus,
+		&:active {
+			border-color: #66afe9;
```
duboisp: We should use the "outline" style instead of the border. The outline wont interfere with the placement of the input field.

sites/search/_base.scss
```diff
@@ -40,6 +40,15 @@
 		box-shadow: none;
 		color: #555;
 		position: relative;
+
+		&:focus,
+		&:active {
+			box-shadow: inset 0 1px 1px rgb(0, 0, 0), 0 0 8px #66afe999;
```
Garneauma: Why the double box shadow? I don't think the first one is needed. It looks a bit weird with only a black top border.

sites/search/_base.scss
```diff
@@ -40,6 +40,15 @@
...
+			outline-color: #66afe9;
+			outline-style: solid;
+			outline-width: 1px;
```
Garneauma:
```suggestion
			outline: #66afe9 solid 1px;
```

PR #2591, opened 2025-07-02
Build: Use JSONLint to format JSON files

nschonni:
## What does this pull request(PR) do?

Separate from the larger PR that enables Lintspaces on all the files, I was looking for a way to repeat the formatting consistently.
This adds a separate task to run the formatting, but I noticed the regular `jsonlint:all` pass doesn't actually check the format (outside of broken JSON), so relied on Lintspaces for now.
An alternate way to check this in CI, would be to run the format and then fail if there is any `git diff`

## Contribution category (please only choose one)

- [ ] Stateless or basic component
- [ ] Interactive component
- [ ] Helper component
- [ ] Site wide component
- [ ] Baseline layout
- [ ] Page layout

## General checklist

- [ ] [Technical documentation](FEATURE.md) created/updated
- [ ] Conforms to the [style guides](https://www.canada.ca/fr/gouvernement/a-propos/systeme-conception.html)
- [ ] Changelog entry added, if necessary
- [ ] Tests added for this feature/bug


_data/common.json
```diff
@@ -1,589 +1,662 @@
...
-				"title": "Assessment d'accessibilité #4 - Boutons",
-				"language": "fr",
-				"path": "reports/a11y-4-fr.html"
-			}
-		]
-	}
-}
-,{
-	"@context": {
-		"@version": 2.0,
```
nschonni, line 188: Seems like a bug, and should probably be quoted

PR #2598, opened 2025-07-08
Conjunction stabilization

BrahimMahadi:
Stabilized conjunction feature. Reference JIRA ticket WET-562.


duboisp (changes requested):
Partial review

Change required:
* [ ] We don't need the prefix "gc-" for the component name, you can remove it.
* [ ] Convert the CSS file into SCSS and add them in our theme.scss file. Also prefix them with "_" to ensure they don't generate new CSS file.
* [ ] Move the Ux research + its images in its own folder under report. And we will quickly draft a French version too.
* [ ] Move the accessibility test in their own report folder + create the placeholder HTML page used to display the report in HTML
* [ ] Need to create the corresponding index.json-ld file + Draft with the info available the variation, implementation, changeset and iteration. Include the accessibility note that was documented.
* [ ] Change the current file identified as "doc" to be "guidance" and remove section that are duplicated from our documentation template. Also add the alert (Jekyll include) about "alter-community-stable.html". 
* [ ] Conjunction - When I look at "Edge-case" that is about deprecated code which happened during the experimentation phase, please remove the example also the CSS that was specifically related to that.
* [ ] We don't have "meta.md" file in our stabilized component.

Garneauma:

BrahimMahadi:
@Garneauma Conjunctions and list steps have been placed in separate PR's, this PR now only contains the changes for conjunctions.

Garneauma (changes requested):
@BrahimMahadi Here are some things to do to begin with:
- [ ] Try to optimize the CSS by converting it to SASS (nested CSS, variables, etc.)
- [ ] Move the different media queries to different files, i.e.: _screen_sm_min.scss)
- [ ] Move the accessibility report under the "report" folder (please rename the folder to "reports") and add the report link in the documentation (under the "assessment", which is under "pages"
- [ ] Create the HTML pages for the a11y report and link to them in the index.json-ld under the "reports" property (refer to the GC Feedback for an example).
- [ ] Move the content of the guidance to the doc page
- [ ] Add the guidance from the méli-mélo to the guidance page (purpose and guidance)
- [ ] Add a link to the UX research somewhere in the doc
- [ ] Make sure to add the configuration options in the instructions

components/conjunctions/conjunction-docs-en.html
```diff
@@ -0,0 +1,19 @@
...
+}
+---
+
+{% include alert-community-stable.html %}
+
+<h2>Other documentation</h2>
+<ul>
+	<li><a href="conjunction-guidance-en.html">Conjunction (and/or) - guidance</a></li>
+	<li><a href="./reports/ux-research.html">Conjunctions (and/or) - ux research</a></li>
+</ul>
```
Garneauma: Move the links inside the implementation tab. The guidance should be in the implementation instructions and the UX research should be in the "notes" property. Then this block could be removed.

components/conjunctions/index.json-ld
```diff
@@ -0,0 +1,187 @@
...
+				"title": "Conjonction (et/ou) - guidance",
+				"language": "fr",
+				"path": "conjunction-guidance-fr.html"
+			},
+			{
+				"title": "Conjonction (et/ou) - guidance",
+				"language": "fr",
+				"path": "conjunction-guidance-fr.html"
+			}
+		]
```
Garneauma, line 67: The links to the a11y assessments are missing. They should be under the "reports" property.

components/conjunctions/conjunction-docs-en.html
BrahimMahadi: Addressed in latest commit.

components/conjunctions/index.json-ld
BrahimMahadi: Addressed in latest commit.

PR #2603, opened 2025-07-16
In-page table of contents: Fix code sample inconsistency

EricDunsworth:
This component's codes sample previously used a ``role="navigation"`` attribute on the pattern's ``nav`` element... but the demo page itself didn't. Not to mention that the attribute is irrelevant in that context (the ``nav`` element already has an implicit ARIA ``navigation`` role and hardcoding it triggers HTML validation warnings).

This resolves it by removing the offending attribute from the code samples.

### DIFF
--- a/components/gc-toc/toc-en.html
+++ b/components/gc-toc/toc-en.html
@@ -3,7 +3,7 @@
 	"title": "In-page table of contents",
 	"language": "en",
 	"altLangPage": "toc-fr.html",
-	"dateModified": "2023-06-16"
+	"dateModified": "2025-07-16"
 }
 ---
 <div class="wb-prettify all-pre"></div>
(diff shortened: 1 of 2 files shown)


components/gc-toc/toc-en.html
```diff
@@ -32,7 +32,7 @@ <h2 id="section3">[Third section heading]</h2>
 </section>
 
 <h2 id="code">Code</h2>
-<pre><code>&lt;nav role="navigation" class="gc-toc"&gt;
+<pre><code>&lt;nav class="gc-toc"&gt;
```
EricDunsworth, line 35: Btw is the ``gc-toc`` class obsolete? It doesn't appear to be tied to any CSS styles. GCWeb's own "On this page" implementations don't even use it.

I wasn't even aware that class existed prior to prepping this PR. Might be a carry-over from long ago when GCWeb used to use "stand-in" classes on all of its components (to future-proof them?). IIRC those kinds of classes were abandoned years ago...

Garneauma: Yes, I believe it was abandoned at some point.

EricDunsworth: @Garneauma In that case, should the ``gc-toc`` class be removed from that page?

AFAIK that page is supposed to be the reference implementation for the "On this page" section. So it's kind of awkward for it not to match what's being done in practice anywhere else :P.

PR #2609, opened 2025-07-23
Theme: selectors adjustments

Garneauma:
Some selector adjustments for easier integration into Canada.ca.


polmih (approved):
Local build with no issues.

PR #2633, opened 2025-08-20
Download link: new pattern

Garneauma:
Changes related to WET-574


duboisp (approved):
Reviewed and tested. Look good.

We can merge once Security check has passed

duboisp:
This will be a minor change because we are adding a new supported pattern for download link.

PR #2639, opened 2025-08-28
Display list items vertically when using .list-col

tristandasilva:
Reference Jira ticket [WET-563](https://jtickets.atlassian.net/browse/WET-563)


tristandasilva:
The issue that these code changes were addressing can be solved without creating a PR. Instead, the class .colcount-no-break can be added to fix the issue identified in [WET-563](https://jtickets.atlassian.net/browse/WET-563).

PR #2658, opened 2025-09-26
Fixing most requested display issue when only 2 items

tristandasilva:
Jira Ticket [WET-590](https://jtickets.atlassian.net/browse/WET-590)

### DIFF
--- a/components/gc-most-requested/_screen-md-min.scss
+++ b/components/gc-most-requested/_screen-md-min.scss
@@ -36,6 +36,12 @@
 				display: block;
 				width: 335px;
 			}
+
+			&:nth-child(2):last-child {
+				break-after: column;
+				break-inside: avoid-column;
+				display: block;
+			}
 		}
 	}
 }
--- a/components/gc-most-requested/gc-most-requested-two-items-en.html
+++ b/components/gc-most-requested/gc-most-requested-two-items-en.html
@@ -1,0 +1,48 @@
+---
+{
+	"altLangPage": "gc-most-requested-two-items-fr.html",
+	"dateModified": "2025-09-29",
+	"description": "Working examples for the Most requested component with only two items.",
+	"language": "en",
+	"title": "Most requested - Two items",
+	"breadcrumbs": [
+		{ "title": "Most requested - Documentation", "link": "components/gc-most-requested/gc-most-requested-doc-en.html" }
+	],
+	"layout": "no-container",
+	"pageclass": "wb-prettify all-pre"
+}
+---
+
+<div class="container">
+	<h1 id="wb-cont">Most requested - Two items</h1>
+	<p class="mrgn-bttm-lg">Below is a working example for the Most requested component with only two items.</p>
+</div>
+
+<section class="gc-most-requested">
+	<div class="container">
+		<h2>Most requested</h2>
+		<ul>
+			<li><a href="#">[Top task hyperlink 1]</a></li>
+			<li><a href="#">[Top task hyperlink 2]</a></li>
+		</ul>
+	</div>
+</section>
+
+<div class="container">
+	<details>
+		<summary>Code</summary>
+		<pre><code>
+		...
+		&lt;section class="gc-most-requested">
+			&lt;div class="container">
+				&lt;h2>Most requested&lt;/h2>
+				&lt;ul>
+					&lt;li>&lt;a href="#">[Top task hyperlink 1]&lt;/a>&lt;/li>
+					&lt;li>&lt;a href="#">[Top task hyperlink 2]&lt;/a>&lt;/li>
+				&lt;/ul>
+			&lt;/div>
+		&lt;/section>
+		...
+		</code></pre>
+	</details>
+</div>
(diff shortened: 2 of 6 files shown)


Ricokola (approved):
Request is complete and working example behave as expected: https://servicecanada.github.io/wet-boew-demos/gcweb-pr2658/components/gc-most-requested/gc-most-requested-en.html

Garneauma (changes requested):
Please document the change in the index.json-ld. It will need a new iteration, this is a minor change since it is a new behaviour. Add history item too.

components/gc-most-requested/_base.scss
```diff
@@ -23,6 +23,11 @@
 	}
 }
 
+.gc-most-requested ul li:nth-child(2):last-child {
+	break-after: column;
+	break-inside: avoid-column;
+}
+
```
Garneauma: Not sure why this is needed? Items on xs and sm breakpoints are stacked vertically, so they would not need this addition.

tristandasilva: @Garneauma The reason for this is because with the display: block, it removed the bullet point when stacking the list items vertically. I can find an alternative solution if you prefer!

tristandasilva: @Garneauma Yup you're absolutely right, the change is only needed in _screen-md-min.scss, I'll make those changes :)

components/gc-most-requested/_base.scss
```diff
@@ -23,6 +23,7 @@
 	}
 }
 
+
```
Garneauma: Please remove this line, so no change is visible to this file.

components/gc-most-requested/_screen-md-min.scss
```diff
@@ -40,6 +40,12 @@
 	}
 }
 
+.gc-most-requested ul li:nth-child(2):last-child {
+	break-after: column;
+	break-inside: avoid-column;
+	display: block;
+}
+
```
Garneauma: Move this rule to the declaration above. It can be included in the `li` like so: `&:nth-child(2):last-child {`.

components/gc-most-requested/gc-most-requested-two-items-en.html
```diff
@@ -0,0 +1,48 @@
...
+
+<section class="gc-most-requested">
+	<div class="container">
+		<h2>{% if page.language == "en" %}Most requested{% else %}En demande{% endif %}</h2>
+		<ul>
+			<li><a href="#">[{% if page.language == "en" %}Top task hyperlink{% else %}Lien vers une tâche principale{% endif %} 1]</a></li>
+			<li><a href="#">[{% if page.language == "en" %}Top task hyperlink{% else %}Lien vers une tâche principale{% endif %} 2]</a></li>
+		</ul>
+	</div>
+</section>
```
Garneauma, line 29: Please remove the "if" for language. We know we are in the EN page.

components/gc-most-requested/gc-most-requested-two-items-fr.html
```diff
@@ -0,0 +1,48 @@
...
+
+<section class="gc-most-requested">
+	<div class="container">
+		<h2>{% if page.language == "en" %}Most requested{% else %}En demande{% endif %}</h2>
+		<ul>
+			<li><a href="#">[{% if page.language == "en" %}Top task hyperlink{% else %}Lien vers une tâche principale{% endif %} 1]</a></li>
+			<li><a href="#">[{% if page.language == "en" %}Top task hyperlink{% else %}Lien vers une tâche principale{% endif %} 2]</a></li>
+		</ul>
+	</div>
+</section>
```
Garneauma, line 29: Same here.

components/gc-most-requested/index.json-ld
```diff
@@ -149,6 +163,24 @@
...
+			],
+			"assets": [
+				{
+					"@type": "source-code",
+					"@language": "en",
+					"description": "Code sample",
+					"code": "<section class=\"gc-most-requested\">\n\t<div class=\"container\">\n\t\t<h2>Most requested</h2>\n\t\t<ul>\n\t\t\t<li><a href=\"#\">[Top task hyperlink 1]</a></li>\n\t\t\t<li><a href=\"#\">[Top task hyperlink 2]</a></li>\n\t\t</ul>\n\t</div>\n</section>"
+				}
+			]
+		},
```
Garneauma, line 187: Great job on the documentation! Please just add the successor for the previous iteration being "_:iteration_mostrequested_3". Thanks!

components/gc-most-requested/index.json-ld
```diff
@@ -70,7 +80,7 @@
 				"en": "https://design.canada.ca/common-design-patterns/most-requested.html",
 				"fr": "https://conception.canada.ca/configurations-conception-communes/en-demande.html"
 			},
-			"iteration": "_:iteration_mostrequested_2",
+			"iteration": "_:iteration_mostrequested_3",
```
Garneauma, line 83: Please add the newly added example in the example list below.

Garneauma (changes requested):

PR #2664, opened 2025-10-03
List steps: revert to meli-melo CSS

Garneauma:


Ricokola (commented):
Did a visual comparison and it behaves has expected.

@duboisp @Garneauma

PR #2677, opened 2025-10-28
Fix most requested list item alignment for the two item layout

tristandasilva:
Jira ticket reference: [WET-609](https://jtickets.atlassian.net/browse/WET-609)

Demo page: [Most requested - Two items](https://servicecanada.github.io/wet-boew-demos/gcweb-pr2677/components/gc-most-requested/gc-most-requested-two-items-en.html)

### DIFF
--- a/components/gc-most-requested/_screen-md-min.scss
+++ b/components/gc-most-requested/_screen-md-min.scss
@@ -36,11 +36,14 @@
 				display: block;
 				width: 335px;
 			}
+		}
 
-			&:nth-child(2):last-child {
-				break-after: column;
-				break-inside: avoid-column;
-				display: block;
+		// If there are only two items in the list, display them side by side, horizontally aligned
+		&:has(>li:nth-child(2):last-child) {
+			display: flex;
+
+			> li {
+				width: 50%;
 			}
 		}
 	}
--- a/_data/components.json
+++ b/_data/components.json
@@ -657,6 +657,40 @@
 		"modified": "dct:modified"
 	},
 	"title": {
+		"en": "Deprecated markup",
+		"fr": "Balisage déconseillé"
+	},
+	"description": {
+		"en": "Reference documents and used to test old practice to ensure compatibility with the latest changes applied at this theme.",
+		"fr": "Ce document est à titre de référence et afin de mettre à l'essaie d'ancienne pratique pour s'assurer qu'ils sont compatible avec les derniers changements apporté au thème."
+	},
+	"modified": "2025-09-23",
+	"componentName": "deprecated",
+	"status": "stable",
+	"pages": {
+		"examples": [
+			{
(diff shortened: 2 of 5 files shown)


Garneauma (approved):

components/gc-most-requested/_screen-md-min.scss
```diff
@@ -17,7 +17,6 @@
 
 		li {
 			display: inline-block;
-			line-height: 1.25em;
```
Ricokola, line 20: This also affect the 2 columns pattern when there are 3 items and more. I suggest narrowing the change to the 2 items pattern only. 

`&:nth-child(2):last-child {
				break-after: column;
				break-inside: avoid-column;
				display: block;
			}`

Garneauma, line 20: Undo this change. It should target when there are only 2 items, not all the time. Also, the line height should not be changed. See how it affects the style when the links wrap over 2 lines.

Garneauma (changes requested):

components/gc-most-requested/_screen-md-min.scss
tristandasilva: Noted! I did test it with wrapped links over 2 lines when I made the change...but I didn't realize that aspect should be untouched, I'll find a better solution :)

Ricokola (dismissed):
Tested locally and it behaves as expected without affecting the pattern with 3 items and more.

cc/ @Garneauma

components/gc-most-requested/_screen-md-min.scss
Garneauma, line 17: The items become misaligned when one of the item wraps over 2 lines. Here's a possible solution (remove everything related to this from before and apply this):
```
&:has(>li:nth-child(2):last-child) {
    display: flex;

    > li {
        width: 50%;
    }
}
```

Works in Chrome, Firefox, and Safari.

Garneauma (changes requested):

components/gc-most-requested/_screen-md-min.scss
```diff
@@ -43,6 +43,14 @@
 				display: block;
 			}
```
Garneauma: Remove this. It is no longer needed.

components/gc-most-requested/_screen-md-min.scss
```diff
@@ -43,6 +43,14 @@
...
 		}
+
```
Garneauma, line 40: Please add a comment above so we know why this is here.

Garneauma (changes requested):

components/gc-most-requested/_screen-md-min.scss
tristandasilva: @Garneauma should be good now :)

Garneauma (dismissed):

BeraJosh (approved):
Approved, thank you!

Garneauma (approved):

PR #2685, opened 2025-11-19
Theme: adding multiple components to theme template

Garneauma:
Changes related to WET-607.


components/gc-most-requested/samples/gc-most-requested-highlight.html
```diff
@@ -0,0 +1,26 @@
+<section class="gc-most-requested highlight">
+	<div class="container">
```
duboisp: We need to implement something that will enforce to have a normal "most-requested" in order to allow the use of an highlighted most request. Something in CSS will be preferable or something in JS. 

I am open to any option. Here an idea for brainstorm. Let's say we leverage a JS check. We can go by assuming the implementer did it right and then we can degrade the styling if it is not by removing the "highlight" and resetting the title if a subsequent most-requested has not been found.

If both need to be immediate sibling, I would like to suggest to wrap both section with a `div.gc-most-requested` instead of having the author to declare 2 independent gc-most-requested pattern.

templates/theme/theme-ircc-en.html
```diff
@@ -0,0 +1,216 @@
...
+						</div>
+						<div class="col-md-6">
+							<h3><a href="https://ircc.canada.ca/iec/index.asp">International Experience Canada (IEC)</a></h3>
+							<p>Special program for youth aged 18 to 35 to work and travel abroad, including working holiday visas</p>
+						</div>
+					</div>
+					<div class="well">
+						<h3>Not sure where to start?</h3>
+						<p><a href="#" class="stretched-link">Answer a few questions and explore all immigration programs</a></p>
+					</div>
```
duboisp, line 104: I am not sure how we will be able to get this implemented in mws. I do like the simplicity of the markup.

May be this should be an extension of our Service and Information component which do leverage a fix pattern inside. Let's discuss.

duboisp (changes requested):
Partial review

templates/theme/theme-ircc-en.html
Garneauma: @duboisp We could have a "Display more information" checkbox, which would show a title field and a freetext field for the content?

components/gc-most-requested/index.json-ld
```diff
@@ -168,9 +209,63 @@
...
+		{
+			"@id": "_:implement_mostrequested_highlight",
+			"iteration": "_:iteration_mostrequested_highlight_1",
+			"name": {
+				"en": "Standard",
+				"fr": "Standard"
+			},
+			"introduction": {
+				"en": "This implementation is meant for developers/publishers adding the component manually.",
+				"fr": "Cette implémentation est destinée aux développeurs/éditeurs qui ajoutent le composant manuellement."
```
duboisp: Add a mention that they must look at the guidance to know when or not to use it.


```suggestion
				"en": "This implementation is meant for developers/publishers adding the component manually. Please check with the guidance on when you can use such pattern.",
				"fr": "Cette implémentation est destinée aux développeurs/éditeurs qui ajoutent le composant manuellement. Veuillez verifié le context d'utilisation de ce pattern selon les lignes directrices"
```

templates/theme/theme-ircc-en.html
duboisp: Yeah, that will work in MWS.

But we are going to allow that only in the gc-serv-info component. Can you extend the gc-serv-info with this new optional config? Thanks

duboisp (changes requested):
Look amazing, I reviewed all example + all code + quickly explored the implementation feasibility into our MWS.

There is : 
* one small addition to do in our documentation.
* Extend the gc service information to include the internal highlighted doormat by using what you did directly inside the theme working example.

Cheers

templates/theme/theme-en.html
duboisp: Let make this alert dismissable.

We can probably use this: https://wet-boew.github.io/gcweb-compiled-demos/wetboew-demos/dismissable-content/dismissable-content-en.html to wrap this alert

duboisp (changes requested):

sites/authentication/includes/authentication.html
```diff
@@ -7,7 +7,7 @@ <h2 class="wb-inv">
 		{%- endunless -%}
 	</h2>
 {%- if page.auth.type == "contextual" -%}
-	{% assign label = page.auth.label | truncate: 12, "" %}
+	{% assign label = page.auth.label | truncate: 25, "" %}
```
duboisp, line 10: Note: This change has not been replicated in our MWS yet

eslint.config.mjs
```diff
@@ -23,6 +23,7 @@ export default defineConfig( [ globalIgnores( [
 	"**/_wetboew-demos/",
 	"**/~jekyll-dist/",
 	"**/~sites/*",
+	"**/_site/*",
```
duboisp, line 26: Is this change related to this PR or a maintenance that this PR need?

duboisp (commented):
Reviewed and did a functional test, all looks good.

This is a minor impact on a few components.
* Minor: template - theme - Revamps of the whole page, but this is a limited use where we can coordinate update on all instances.
* Minor: component - gc-most-requested - Add highlight feat for theme template
* Minor: component - gc-srvinfo - Add 'Not sure where to start'
* Patch: site - authentification - Increase the number of character for sign in short label
* Build: Add "_site" to ignore list for files linting

Todo: 
* Need to revisit the AEM working example variant. I am just waiting for getting the latest AEM implementation of the theme page
* Need to review description for the technical impact

duboisp:
Pre-approved after the AEM code sample is updated as per the latest AEM theme template markup. That don't impact previous revision, it is just content for our info.

duboisp (approved):
Content change - Copy of MWS implementation was updated.

--

Last completed technical review remain accurate

For impact change, see: https://github.com/wet-boew/GCWeb/pull/2685#pullrequestreview-4533146920

Yves-Denis (approved):

sites/authentication/includes/authentication.html
duboisp: Will be addressed in mws code base later on.

PR #2695, opened 2025-11-28
Deprecate demoted elements and remove deprecated elements

Garneauma:
### Deprecated
- Download link version 1. User version 2 instead.
- Site footer version 8, 8.1, 9 and 10. User version 11 instead.
- Page details version 3.2. Use version 3.3 instead.
- Topic template version 2.1. Use version 2.2 instead
- Ministerial profiles. Use Topic template instead.


duboisp:
We should froze the méli-mélo 2025-04-nahanni which is overdue a few months

templates/all-services/index.json-ld
duboisp, line 1: Comment - Can be completed for the subsequent release. Add in the documentation the URL of the live pages that correspond to this template, this is to clearly describe this template are for "Limited used" and the DD14 do require that we are listing those live corresponding URLs.

templates/topic/deprecated/topic-stackedheader_v2-en.html
duboisp, line 1: Comment - Content change to be made in a subsequent version - In the documentation, move the deprecated link under "working example" into the corresponding iteration.

PR #2709, opened 2025-12-23
Build: fix Jekyll build warnings, add build using Bundler instead of Docker

Garneauma:
- Removing all Jekyll build warnings
- Adding command in package.json to build the site using Bundler instead of docker
  - Required: Ruby and Bundler

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -879,6 +879,14 @@ module.exports = (grunt) ->
 				options: {
 					process: (content, srcpath) ->
 						return content.replace(/{{>alertariahidden}}/g, "")
+										.replace(/acr_wrote_en-en.hbs/g, "acr_wrote_en-en")
+										.replace(/acr_wrote_en-fr.hbs/g, "acr_wrote_en-fr")
+										.replace(/assessment_wrote_en-en.hbs/g, "assessment_wrote_en-en")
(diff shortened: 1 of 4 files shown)


duboisp:
@BeraJosh For your review

Gruntfile.coffee
```diff
@@ -879,6 +879,14 @@ module.exports = (grunt) ->
...
 					process: (content, srcpath) ->
 						return content.replace(/{{>alertariahidden}}/g, "")
+										.replace(/acr_wrote_en-en.hbs/g, "acr_wrote_en-en")
+										.replace(/acr_wrote_en-fr.hbs/g, "acr_wrote_en-fr")
+										.replace(/assessment_wrote_en-en.hbs/g, "assessment_wrote_en-en")
+										.replace(/assessment_wrote_en-fr.hbs/g, "assessment_wrote_en-fr")
+										.replace(/{{#escape}}/g, "{% raw %}")
+										.replace(/{{\/escape}}/g, "{% endraw %}")
+										.replace(/{{#stripbanner}}/g, "")
+										.replace(/{{\/stripbanner}}/g, "")
```
duboisp, line 889: This change does remove the warning, but it does impact how the page is rendering.

I am looking now at the impact. May be the best will be to remove those upstream instead, its only about 10 pages.

Expected:

Image

After this change

Image

But that wont change what we see currently

Image

PR #2719, opened 2026-01-27
GC Combo-box component (WET-631)

BrahimMahadi:
An accessible combo-box component that provides a modern, user-friendly interface for selecting multiple items. Reference JIRA ticket WET-631.

- [x] Complete feature implementation and implement/remove any additional features after review
- [x] Complete documentation
- [x] Refactor code
- [ ] Perform accessibility assessment


components/combo-box/combo-box.js
```diff
@@ -0,0 +1,614 @@
...
+		this.allOptions = options;
+		this.selectedItems = this.selectedItems.filter( item => this.allOptions.includes( item ) );
+		this.updateFilteredOptions();
+		this.renderTags();
+		this.renderOptions();
+	}
+}
+
+// Register the custom element
+customElements.define( "combo-box-component", ComboBoxComponent );
```
Garneauma: Let's go with "gc-combobox" for now.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,614 @@
...
+
+			// This breaks the grunt dist build, if we do not need it in this context, we can leave it out
+			// '"': "&quot;",
+			"'": "&#039;"
+		};
+		return text.replace( /[&<>"']/g, m => map[ m ] );
+	}
+
+	// Returns the encapsulated styles for the Shadow DOM
+	getStyles() {
```
Garneauma: Can you find a way to get styles from a CSS file instead? This would be easier to test using our linting tools.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,614 @@
...
+			return Array.from( slot.querySelectorAll( "option" ) ).map( opt => opt.textContent );
+		}
+
+		return [];
+	}
+
+	// Renders the component template and styles into Shadow DOM
+	render() {
+		const label = this.getAttribute( "label" ) || "Select items";
+		const placeholder = this.getAttribute( "placeholder" ) || "Type to search or add items...";
```
Garneauma: Those default values would need to use i18n strings.

Garneauma (changes requested):
This is just a quick partial review.

duboisp (changes requested):
A few comment after your presentation.

* Use/allow the slot for label
* Use named slot for options instead of JSON array in an attribute
* Use series of hidden input for an integration into form API
* Use "is" ex: <input is="combo-box-component">. (But this will require functional + compatibility testing in safari + safari mobile)
* Allow an option where we manually add my own custom tag. vs using datalist (for more info, check the HTML5 specification around it)
* Allow the customization of colour used for tags (background+colour) by using the `part` attribute

Authoring code sample example
```
<label for="mycbx">Label</label>
<input id="mycbx" is="combo-box-component" multiple name="cbx" list="function-types">
<datalist id="function-types">
  <option value="function">function</option>
  <option value="async function">async function</option>
  <option value="function*">generator function</option>
  <option value="=>">arrow function</option>
  <option value="async =>">async arrow function</option>
  <option value="async function*">async generator function</option>
</datalist>
```
Or alternatively
```
<combo-box-component
	id="demo"
	placeholder="Select technologies...">

<div slot="label">Select<sup>sup</sup> Technologies label</div>

<datalist slot="options">
  <option>JavaScript</option>
  <option>TypeScript</option>
  <option>React</option>
  <option>Vue</option>
  <option>Angular</option>
  <option>Node.js</option>
</datalist>


</combo-box-component>

```


Other ref.
* https://html.spec.whatwg.org/multipage/input.html#attr-input-multiple
* https://html.spec.whatwg.org/multipage/input.html#attr-input-list

And regardless of my above comment, we know that eventually this component, after our immediate need satisfied, we are going to transform / re-code it as a GCDS component, so feel free to take decision that will be aligned with how GCDS component will be develop.

components/combo-box/combo-box.js
BrahimMahadi: Component name has been changed to gc-combobox.

components/combo-box/combo-box.js
BrahimMahadi: I moved all of the css to combo-box.css.

components/combo-box/combo-box.js
BrahimMahadi: Updated in latest commit.

components/combo-box/combo-box.css
```diff
@@ -0,0 +1,212 @@
+* {
```
Garneauma, line 1: To avoid code duplication, you can nest CSS rules. Check out Mozilla's documentation to know what's available: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Nesting/Using

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+	// Loads styles from combo-box.css file
+	static async loadStyles() {
+		if ( ComboBoxComponent.stylesCache !== null ) {
+			return ComboBoxComponent.stylesCache;
+		}
+
+		try {
+			const response = await fetch( "combo-box.css" );
+			if ( !response.ok ) {
+				throw new Error( `Failed to load _combo-box.scss: ${ response.statusText }` );
```
Garneauma:
```suggestion
				throw new Error( `Failed to load combo-box.css: ${ response.statusText }` );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			}
+			ComboBoxComponent.stylesCache = await response.text();
+			return ComboBoxComponent.stylesCache;
+		} catch ( e ) {
+			console.error( "Failed to load styles from _combo-box.scss:", e );
```
Garneauma:
```suggestion
			console.error( "Failed to load styles from combo-box.css:", e );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			} catch ( e ) {
+				console.error( "Invalid options format. Expected JSON array.", e );
+			}
+		}
+
+		// Map label types to JSON files
+		const fileMap = {
+			"regions": "regions.json",
+			"institutions": "institutions.json"
+		};
```
Garneauma: I don't believe hardcoded files should be in the ready for review version of the component...

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			return ComboBoxComponent.stylesCache;
+		} catch ( e ) {
+			console.error( "Failed to load styles from _combo-box.scss:", e );
+			return ""; // Return empty string as fallback
+		}
+	}
+
+	// Parses options from JSON files based on the name attribute
+	async parseOptions() {
+		const labelType = this.getAttribute( "name" ) || this.getAttribute( "placeholder" );
```
Garneauma: Not sure I understand the relationship between "labelType" and the values given to it.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			}
+		}
+
+		// Map label types to JSON files
+		const fileMap = {
+			"regions": "regions.json",
+			"institutions": "institutions.json"
+		};
+
+		const jsonFile = fileMap[ labelType?.toLowerCase() ];
```
Garneauma: The file fetched should be defined on the component so that the user can define a specific file path for the options. For example: `options="institutions.json"`.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+		if ( !jsonFile ) {
+
+			// Fallback to parsing from slot content if no matching label type
+			const slot = this.querySelector( "[slot='options']" );
```
Garneauma:
```suggestion
			const slotOptions = this.querySelector( "[slot='options']" );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+
+			// Extract the display values based on the page language
+			const options = Object.values( data ).map( item => {
+
+				// Support both nested language objects and simple values
+				if ( typeof item === "object" && item !== null ) {
+					return item[ pageLanguage ] || item.en || item.fr || Object.values( item )[ 0 ];
+				}
+				return String( item );
+			} );
```
Garneauma: Instead of having a logic that fetches the correct language, it would be better to have a file for the list of options in FR and one in EN. The author could then simply supply the correct file for the current page's language.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+				"fr": {
+					regions: "Sélectionner région(s)", // text for the french region label
+					institutions: "Sélectionner institution(s)", // text for the french institution label
+					selectAll: "Sélectionner tous" // text for the french select all checkbox
+				}
+			}
+		};
+
+		// Determine page language (default to English)
+		const pageLanguage = ( document.documentElement.lang || "en" );
```
Garneauma: You can get the page language from `wb.lang`.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+
+	// Renders the component template and styles into Shadow DOM
+	async render() {
+
+		var defaults = {
+			i18n:
+			{
+				"en": {
+					regions: "Select region(s)", // text for the region label
+					institutions: "Select institution(s)", // text for the institution label
```
Garneauma: Why is this here?

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+
+		// Determine page language (default to English)
+		const pageLanguage = ( document.documentElement.lang || "en" );
+
+		// Get label type from attribute (e.g., "regions" or "institutions")
+		const labelType = this.getAttribute( "name" );
+
+		// Get label and placeholder from i18n based on page language and label type
+		const label = defaults.i18n[ pageLanguage ]?.[ labelType ] || this.getAttribute( "label" );
+		const placeholder = defaults.i18n[ pageLanguage ]?.[ labelType ] || this.getAttribute( "placeholder" );
```
Garneauma: The label text should either come from the "label" attribute or from the slot. Not from options in this file.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+				${ styles }
+			</style>
+
+			<div class="combo-box-wrapper">
+				<label for="combo-box-input" class="combo-box-label">
+				<slot name="label">${ this.escapeHtml( label ) }</slot>
+				</label>
+
+				<div class="combo-box-input-row">
+					<div class="combo-box-container" role="combobox">
```
Garneauma: The combobox role should be on the input. For help on accessibility pattern for the multiselect combobox, check out this example: https://cauldron.dequelabs.com/components/Combobox#multiselect

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+						<div class="tags-container" id="tagsContainer" role="group" aria-label="Selected items">
+							<!-- Tags will be dynamically inserted here -->
+							<input
+								type="text"
+								id="combo-box-input"
+								class="combo-box-input"
+								placeholder="${ this.escapeHtml( placeholder ) }"
+								autocomplete="off"
+								aria-label="Search and select items"
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			</style>
+
+			<div class="combo-box-wrapper">
+				<label for="combo-box-input" class="combo-box-label">
+				<slot name="label">${ this.escapeHtml( label ) }</slot>
+				</label>
+
+				<div class="combo-box-input-row">
+					<div class="combo-box-container" role="combobox">
+						<div class="tags-container" id="tagsContainer" role="group" aria-label="Selected items">
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+							>
+						</div>
+					</div>
+
+					<div class="select-all-wrapper">
+						<input
+							type="checkbox"
+							id="combo-box-select-all"
+							class="combo-box-select-all-checkbox"
+							aria-label="Select all options"
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+						<label for="combo-box-select-all" class="select-all-label">${ this.escapeHtml( selectAllLabel ) }</label>
+					</div>
+				</div>
+
+				<ul
+					id="combo-box-list"
+					class="combo-box-list"
+					role="listbox"
+					aria-label="Available options"
+					hidden
```
Garneauma: Add `aria-multiselectable="true"`

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			case "Enter":
+				e.preventDefault();
+				if ( this.highlightedIndex >= 0 && this.isOpen ) {
+					this.selectOption( this.filteredOptions[ this.highlightedIndex ] );
+				}
+				break;
+			case "Escape":
+				e.preventDefault();
+				this.closeList();
+				this.input.value = "";
```
Garneauma: According to W3C guidelines: "Escape: If the listbox is displayed, closes it. If the listbox is not displayed, clears the textbox."

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+			this.renderTags();
+			this.input.value = "";
+			this.updateFilteredOptions();
+			this.renderOptions();
+			this.highlightedIndex = -1;
+			this.openList();
+			this.input.focus();
+
+			// Announce selection to screen readers
+			this.announce( `${ option } selected` );
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+
+		// Uncheck select all if not all items are selected anymore
+		if ( this.selectedItems.length !== this.allOptions.length ) {
+			this.selectAllCheckboxChecked = false;
+			this.selectAllCheckbox.checked = false;
+			this.enableInput();
+		}
+
+		// Announce removal to screen readers
+		this.announce( `${ option } removed` );
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+
+		this.selectedItems.forEach( item => {
+			const tag = document.createElement( "div" );
+			tag.className = "tag";
+			tag.innerHTML = `
+				<span class="tag-text">${ this.escapeHtml( item ) }</span>
+				<button
+					type="button"
+					class="tag-remove-btn"
+					aria-label="Remove ${ item }"
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+	renderTags() {
+		const hadFocus = this.shadowRoot.activeElement === this.input;
+		this.tagsContainer.innerHTML = "";
+
+		this.selectedItems.forEach( item => {
+			const tag = document.createElement( "div" );
+			tag.className = "tag";
+			tag.innerHTML = `
+				<span class="tag-text">${ this.escapeHtml( item ) }</span>
+				<button
```
Garneauma: The remove button might be too small for accessibility (minimum 44x44). You could make the whole tag a button with an aria label of "Remove [item]".

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+		this.tagsContainer.appendChild( this.input );
+		this.input.classList.toggle( "has-selections", this.selectedItems.length > 0 );
+
+		if ( hadFocus ) {
+			this.input.focus();
+		}
+
+		// Update aria-label with selection count
+		const count = this.selectedItems.length;
+		const label = count === 0 ? "Selected items" : `Selected items (${ count } selected)`;
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+	renderOptions() {
+		this.list.innerHTML = "";
+		this.input.removeAttribute( "aria-activedescendant" );
+
+		if ( this.filteredOptions.length === 0 ) {
+			const emptyOption = document.createElement( "li" );
+			emptyOption.className = "combo-box-option empty-state";
+			emptyOption.setAttribute( "role", "option" );
+			emptyOption.setAttribute( "aria-disabled", "true" );
+			emptyOption.textContent = "No matching options";
```
Garneauma: i18n

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,684 @@
...
+	setSelectedItems( items ) {
+		this.selectedItems = items.filter( item => this.allOptions.includes( item ) );
+		this.updateFilteredOptions();
+		this.renderTags();
+		this.renderOptions();
+
+		// Update select all checkbox state
+		this.updateSelectAllCheckbox();
+
+		this.dispatchEvent( new CustomEvent( "change", {
```
Garneauma: Do you need to create a new custom event every time you want to trigger the change or could you store it somewhere and trigger it when appropriate?

components/combo-box/index.json-ld
```diff
@@ -0,0 +1,163 @@
...
+				"en": "Standard",
+				"fr": "Standard"
+			},
+			"introduction": {
+				"en": "This implementation is meant for developers/publishers adding the component manually.",
+				"fr": "Cette implémentation est destinée aux développeurs/éditeurs qui ajoutent le composant manuellement."
+			},
+			"instructions": {
+				"en": [
+					"Refer to the following code samples."
```
Garneauma: Make sure you document every available option and slot and their expected value format. This is not a simple component without options. The documentation will need to be very detailed.

Garneauma (changes requested):
This is a partial review. Apart from my inline comments, here are some other thoughts:
- I don't like the fact that all the tags blink every time you add a new tag. The fade in animation should only happen on the first one.
- You missed quite a few i18n strings. Make sure that EVERY text string is translated.
- The text box does not take the full height of the box, so when you click on the top part or bottom part of the box, nothing happens. Make sure that every part of the box click will trigger the focus to the input.
- If you type something that returns no match, then click out, then click back in, there should still be no match. Currently it shows all the options like you have an empty text box.

components/combo-box/combo-box.js
BrahimMahadi: Whole tag was turned into a button with an appropriate aria label to remove the item.

components/combo-box/combo-box.js
BrahimMahadi: Event was put into a method and gets called when necessary.

components/combo-box/combo-box.js
BrahimMahadi: Hardcoded files have been removed and replaced with a fetch call to retrieve the data

components/combo-box/combo-box.js
BrahimMahadi: Code has been refactored to define the file fetched in the options of the component.

components/combo-box/combo-box.js
BrahimMahadi: This was used to add the text for the labels. Labels have been properly implemented, therefore this code has been removed.

components/combo-box/combo-box.js
BrahimMahadi: FIxed in "Updated changes" commit

components/combo-box/combo-box.js
BrahimMahadi: This was used to determine how the options were being parsed from the json. The functionality has been reworked and so "labeltype" is no longer necessary.

components/combo-box/combo-box.js
BrahimMahadi: I agree, the code has been refactored so that the json given should be in the language of the html page so that the JS file can just read in the page language.

components/combo-box/combo-box.css
BrahimMahadi: The css has been refactored so that the appropriate rules are nested.

components/combo-box/combo-box-en.html
```diff
@@ -0,0 +1,16 @@
...
+altLangPage: combo-box-fr.html
+---
+<div class="mrgn-bttm-lg"></div>
+<gc-combobox id="demo-regions" placeholder="Select region(s)..." name="regions" options="https://api.io.canada.ca/io-server/gc/news/institutions-and-regions-tags/v1/en/regions">
+	<span slot="label">Select region(s)</span>
+</gc-combobox>
+<br>
+<gc-combobox id="demo-institutions" name="institutions" placeholder="Select institution(s)..." options="https://api.io.canada.ca/io-server/gc/news/institutions-and-regions-tags/v1/en/institutions">
+	<span slot="label">Select institution(s)</span>
+</gc-combobox>
```
Garneauma, line 62: - Add an example where the options are added through the slot.
- Add an option where the label is not a slot.
- It seems you handle when a placeholder is not provided, please add an example for that too.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,752 @@
...
+							role="listbox"
+							aria-multiselectable="true"
+							aria-label="${ this.escapeHtml( this.constructor.defaults.i18n[ this.pageLanguage ].availableOptions ) }"
+							hidden
+						>
+					<!-- Options will be dynamically inserted here -->
+				</ul>
+					</div>
+
+					<div class="select-all-wrapper">
```
Garneauma: The "Select all" feature should be optional. Also, I think the checkbox should be either above or under the input field.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,752 @@
...
+						<ul
+							id="combo-box-list"
+							class="combo-box-list"
+							role="listbox"
+							aria-multiselectable="true"
+							aria-label="${ this.escapeHtml( this.constructor.defaults.i18n[ this.pageLanguage ].availableOptions ) }"
+							hidden
+						>
+					<!-- Options will be dynamically inserted here -->
+				</ul>
```
Garneauma: Why is there two identical `<ul>` (other one on line 209)?

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,752 @@
...
+
+	// Renders the component template and styles into Shadow DOM
+	async render() {
+
+		// Get label from attribute or use slot content
+		const label = this.getAttribute( "label" ) || this.querySelector( "[slot='label']" )?.textContent.trim() || "";
+
+		// Get placeholder from attribute
+		const placeholder = this.getAttribute( "placeholder" ) || "";
+		const selectAllLabel = this.constructor.defaults.i18n[ this.pageLanguage ]?.selectAll;
```
Garneauma: `this.constructor.defaults.i18n` This part could be stored in a variable.

Garneauma (changes requested):
This is a partial review.

- [ ] This might be more for CEO to define, but after discussing with @GormFrank, I think the best option for the "Select all" would be to, instead of having a checkbox, have an option at the top of the list like "Select all options" with a different styling. When selecting that option, all the options would disappear, a tag with "All options" would appear in the box, and the suggestions box would show something like "No options available".
- [ ] Fix working examples page. No `mrgn-bttm-lg` div. No `<br>`. Add a description paragraph under the H1 and add margin bottom to your component.
- [ ] For the styles, try to make the label styles match what is on GCWeb. I guess you'll have to copy the GCWeb styles in this CSS.

components/combo-box/combo-box-en.html
BrahimMahadi: Additional examples have been added

components/combo-box/combo-box.js
BrahimMahadi: @Garneauma The implementation for select all has been changes from check-box to tag option in the list. Knowing this, do we still want to add functionality to make it optional?

components/combo-box/combo-box.js
BrahimMahadi: The duplicate ul's have been removed

components/combo-box/combo-box.js
BrahimMahadi: Variable was added and used wherever necessary.

components/combo-box/combo-box.css
```diff
@@ -0,0 +1,212 @@
...
+		&:hover {
+			background-color: transparent;
+		}
+	}
+
+	&.combo-box-option-select-all {
+		background-color: #f5f5f5;
+		border-bottom: 1px solid #ddd;
+		font-weight: 600;
+		padding: 8px 12px;
```
Garneauma: I think a simple `font-style: italic` would be sufficient.

components/combo-box/combo-box-en.html
```diff
@@ -0,0 +1,33 @@
+---
+title: GC Combo Box Web Component
+dateModified: 2026-05-13
+description: "An accessible combo-box component that implements WCAG 2.1 accessibility standards and provides a modern,
+user-friendly interface for selecting multiple items."
+lang: en
+altLangPage: combo-box-fr.html
```
Garneauma: Add breadcrumb item to the documentation page.

Garneauma (changes requested):
- There seems to be an issue when navigating with the arrow now. The index is messed up because of the "Select all" option.
- When I mentioned to match the label styles, I didn't mean to match the inline form style. Just the font. Please put the label back on its own row. The inline form can certainly be an option if you want to, but it won't be mandatory for this component.
- Please add the code sample below each example.
- Once this is all done. Please get started on the documentation. All options will need to be documented:
  - The default value
  - The type
  - Whether a slot is available or not
  - Whether the option is mandatory or not
- For the "options" option, make sure to document how the data must be stored in the JSON. I see that it uses specific property names. Have you tried with a simple JSON file instead of a feed?

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi I'll have to discuss with Francis. Let's leave it as is for now.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+		}
+		this.updateHighlight();
+	}
+
+	// Updates the visual highlight for the current option
+	updateHighlight() {
+		const options = this.list.querySelectorAll( "[role='option']" );
+		options.forEach( ( option, index ) => {
+			if ( index === this.highlightedIndex ) {
+				option.setAttribute( "aria-selected", "true" );
```
Garneauma: Not sure if this option is necessary. The item is not necessarily selected, simply highlighted. It might confuse screen-reader users as it will announce "Selected" when moving from an item to another. If it is for CSS, another attribute or a class should be used.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+			const emptyOption = document.createElement( "li" );
+			emptyOption.className = "combo-box-option empty-state";
+			emptyOption.setAttribute( "role", "option" );
+			emptyOption.setAttribute( "aria-disabled", "true" );
+			emptyOption.textContent = this.constructor.defaults.i18n[ this.pageLanguage ].noMatchingOptions;
+			this.list.appendChild( emptyOption );
+		} else if ( this.filteredOptions.length === 0 ) {
+			const emptyOption = document.createElement( "li" );
+			emptyOption.className = "combo-box-option empty-state";
+			emptyOption.setAttribute( "role", "option" );
```
Garneauma: When showing the "No matching option" text, the `<li>` should not have the attribute `role="option"`. Rather it should have the `role="presentation"`.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+			} ) );
+		}
+		return [];
+	}
+
+	// Renders the component template and styles into Shadow DOM
+	async render() {
+
+		// Get label from attribute or use slot content
+		const label = this.getAttribute( "label" ) || this.querySelector( "[slot='label']" )?.textContent.trim() || "";
```
Garneauma, line 196: The label should be mandatory. If no label is defined, the component should not render. Same for the list of options.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+	openList() {
+		if ( !this.isOpen ) {
+			this.isOpen = true;
+			this.list.removeAttribute( "hidden" );
+			this.input.setAttribute( "aria-expanded", "true" );
+		}
+	}
+
+	// Closes the dropdown list
+	closeList() {
```
Garneauma, line 736: You should clear the `aria-activedescendant` attribute when closing the list. Otherwise, it still refers to a hidden option.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+								type="text"
+								role="combobox"
+								id="combo-box-input"
+								class="combo-box-input"
+								placeholder="${ this.escapeHtml( placeholder ) }"
+								autocomplete="off"
+								aria-label="${ this.escapeHtml( this.getLocalizedText().comboBoxInput ) }"
+								aria-autocomplete="list"
+								aria-controls="combo-box-list"
+								aria-expanded="false"
```
Garneauma, line 225: Missing `aria-haspopup="listbox"`.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+							>
+						</div>
+						<ul
+							id="combo-box-list"
+							class="combo-box-list"
+							role="listbox"
+							aria-multiselectable="true"
```
Garneauma: After some more thinking, I'm not sure if this attribute is applicable to our case. Since the list of options removes the selected item each time, it is not possible to select multiple items at the same time. Should we instead of removing the items, simply add something like a checkmark next to the item? What do you think?

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+
+		this.selectedItems = this.selectedItems.filter( item => item.value !== value );
+		this.allOptionsSelected = false;
+		this.renderTags();
+		this.updateFilteredOptions();
+		this.renderOptions();
+		this.input.focus();
+
+		// Announce removal to screen readers
+		this.announce( `${ option.label }` + `${ this.getLocalizedText().removed }` );
```
Garneauma: Missing a space. It would read Labelremoved.
```suggestion
		this.announce( `${ option.label }` + ` ${ this.getLocalizedText().removed }` );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+				break;
+			case "Escape":
+				e.preventDefault();
+				if ( this.isOpen ) {
+					this.closeList();
+				} else {
+					this.input.value = "";
+				}
+				break;
+			case "Backspace":
```
Garneauma, line 390: This should be announced to screen-readers like you do in the removeTag() method.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+				} else {
+					const optionText = option.getAttribute( "data-option-text" );
+					if ( optionText ) {
+						this.selectOption( optionText );
+					}
+				}
+			}
+		} );
+
+		this.list.addEventListener( "mouseenter", ( e ) => {
```
Garneauma: Shouldn't this be `mouseover` ?
```suggestion
		this.list.addEventListener( "mouseover", ( e ) => {
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,755 @@
...
+		// Always append the input
+		this.tagsContainer.appendChild( this.input );
+
+		if ( hadFocus ) {
+			this.input.focus();
+		}
+	}
+
+	// Renders the filtered options in the dropdown
+	renderOptions() {
```
Garneauma, line 635: This method would benefit from having a data layer on the component. It is not efficient to re-render the items, re-generate IDs, etc. every time the popup opens. Instead, it could be stored in a data layer then fetched when opening the popup. Not critical, but would be nice.

Garneauma (changes requested):
Outside of the requested changes. Please explore ElementInternals concept. I should be a more future-proof solution to send data instead of the hidden inputs. Sorry, I just learned about that concept today. It is also useful to manage form reset and form validation.

Once we are ready, there will be the outstanding question about the styles. Since GCWeb implementers only have theme.css and theme.js loaded. They wouldn't have access to the combo-box.css file. I'll have to talk with Pierre about that.

Thank you @BrahimMahadi. We are getting close!

components/combo-box/combo-box.js
BrahimMahadi: Yes I agree, if we aren't removing tags then we shouldn't be using the multiselectable attribute. In the case of the checkmark, as a user I would prefer an item be removed from the list when I select it but it's worth asking the media advisory team / client their thoughts on this.

Garneauma (commented):
Here's a snippet of code to add a task to replace a string in the JS file with the CSS from the component in the CSS file:
```
	@registerTask(
		"embed-component-styles"
		->
			# Add your component URL here. Make sure to include
			# @[component-name]_styles@ in your JS file, where [component-name]
			# equals to the name of the JS file, CSS file, and the name of your component folder
			components = [
				"components/combo-box"
			]
			themeJsPath = "dist/GCWeb/js/theme.js"

			unless grunt.file.exists themeJsPath
				grunt.fail.warn "Could not find #{themeJsPath}"

			themeJs = grunt.file.read themeJsPath

			for componentPath in components
				componentName = componentPath.split("/").pop()
				cssFiles = grunt.file.expand {cwd: componentPath}, ["*.css"]

				unless cssFiles.length
					grunt.log.warn "No CSS file found in #{componentPath}"
					continue

				cssPath = "#{componentPath}/#{cssFiles[0]}"
				cssContent = grunt.file.read cssPath
				placeholder = "@#{componentName}_styles@"

				if themeJs.indexOf(placeholder) is -1
					grunt.log.warn "Placeholder not found: #{placeholder}"
					continue

				escapedPlaceholder = placeholder.replace /[.*+?^${}()|[\]\\]/g, "\\$&"
				themeJs = themeJs.replace new RegExp(escapedPlaceholder, "g"), cssContent
				grunt.log.ok "Embedded styles for #{componentName}"
				grunt.file.write themeJsPath, themeJs
	)
```

You can add this snippet on line 263, right below the last custom task. Then, add the task to the end of the list for "core-dist" on line 225.

You could the remove the `loadStyles()` and `getStyles()` methods and replace the styles variable in you render method to the following:
```
const styles = `
			@combo-box_styles@
		`;
```

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: Yes it should, Fixed in commit "Added support to load css through the build as well as fixes"

components/combo-box/combo-box.js
BrahimMahadi: I added a buildOptionsCache() so that all the options are create one time with stable ID's, then referenced whenever necessary

BrahimMahadi:
> Here's a snippet of code to add a task to replace a string in the JS file with the CSS from the component in the CSS file:
> 
> ```
> 	@registerTask(
> 		"embed-component-styles"
> 		->
> 			# Add your component URL here. Make sure to include
> 			# @[component-name]_styles@ in your JS file, where [component-name]
> 			# equals to the name of the JS file, CSS file, and the name of your component folder
> 			components = [
> 				"components/combo-box"
> 			]
> 			themeJsPath = "dist/GCWeb/js/theme.js"
> 
> 			unless grunt.file.exists themeJsPath
> 				grunt.fail.warn "Could not find #{themeJsPath}"
> 
> 			themeJs = grunt.file.read themeJsPath
> 
> 			for componentPath in components
> 				componentName = componentPath.split("/").pop()
> 				cssFiles = grunt.file.expand {cwd: componentPath}, ["*.css"]
> 
> 				unless cssFiles.length
> 					grunt.log.warn "No CSS file found in #{componentPath}"
> 					continue
> 
> 				cssPath = "#{componentPath}/#{cssFiles[0]}"
> 				cssContent = grunt.file.read cssPath
> 				placeholder = "@#{componentName}_styles@"
> 
> 				if themeJs.indexOf(placeholder) is -1
> 					grunt.log.warn "Placeholder not found: #{placeholder}"
> 					continue
> 
> 				escapedPlaceholder = placeholder.replace /[.*+?^${}()|[\]\\]/g, "\\$&"
> 				themeJs = themeJs.replace new RegExp(escapedPlaceholder, "g"), cssContent
> 				grunt.log.ok "Embedded styles for #{componentName}"
> 				grunt.file.write themeJsPath, themeJs
> 	)
> ```
> 
> You can add this snippet on line 263, right below the last custom task. Then, add the task to the end of the list for "core-dist" on line 225.
> 
> You could the remove the `loadStyles()` and `getStyles()` methods and replace the styles variable in you render method to the following:
> 
> ```
> const styles = `
> 			@combo-box_styles@
> 		`;
> ```

This seems to have worked. The css gets loaded into the theme.js file on build

Garneauma (changes requested):
From David's review, here's what has not been implemented:
- Instructions on how to navigate the combobox (up/down arrows to navigate the selected options, left/right arrows to navigate) should be added via a "aria-describedby" attribute that points a hidden div.
- When using keyboard navigation, the only item that should be tabbable is the text input, not the buttons. You should then be able to navigate the selected options via the arrow keys.
- When focused on a selected option, enable the "DELETE" button to remove an option.
- The aria-label attribute on the text input overrides the associated label element. The text "Search and select items" should be added to the above description. Here's what it could be: "Search and select options. To navigate the available options, use the up and down arrow keys. To navigate the selected options, use the left and right arrow keys."

Other things I noticed:
- The text input should not be inside the "tags-container". Because it is, the SR always reads "Selected items" when focusing on the text input. This will probably mean some CSS adjustments.
- The backspace action should not delete the last selected item right away. It should focus on the last selected item and then if you press again, it should delete the item. For an example, refer back to the example mentioned before: https://cauldron.dequelabs.com/components/Combobox/#multiselect

tristandasilva:
Accessibility improvements:
- Tags can be navigated by using the arrow keys
- Tags can be deleted by using the delete key, backspace, or focusing + enter/spacebar
- Only the input field can be tabbed into
- Changed role="group" to role="toolbar" which also allows the user to be able to use the arrow keys for navigation between the tags
- Instructions removed entirely as discussed, there is enough context and screen reader users will be used to this behaviour
- Improved live region announcement text

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+	// Builds the data layer: creates all option <li> elements once with stable IDs
+	buildOptionsCache() {
+		this.optionsCache = new Map();
+
+		this.allOptions.forEach( ( option, index ) => {
+			const optionElement = document.createElement( "li" );
+			optionElement.id = `combo-option-${ this.instanceId }-${ index }`;
+			optionElement.className = "combo-box-option";
+			optionElement.setAttribute( "role", "option" );
+			optionElement.setAttribute( "aria-selected", "false" );
```
Garneauma: Remove `aria-selected` from the list of attributes for the options. I think this attribute is misleading as the items is not selected, but rather "active", which is already handled with the `aria-activedescendant` attribute. Remove also at lines: 138, 445, 453, 636, and 644.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+	updateFilteredOptions() {
+		this.filteredOptions = [ ...this.allOptions ].filter(
+			item => !this.selectedItems.map( i => i.value ).includes( item.value )
+		);
+	}
+
+	// Renders the selected items as tags
+	renderTags() {
+		const hadFocus = this.shadowRoot.activeElement === this.input;
+		this.tagsContainer.innerHTML = "";
```
Garneauma: Instead of completely rebuilding the tags container every time, simply diff the tags. Either add or remove a tag. This is not a blocker, but it is a bit overkill.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+					this.input.setAttribute( "aria-activedescendant", option.id );
+				}
+			} else {
+				option.setAttribute( "aria-selected", "false" );
+			}
+		} );
+	}
+
+	// Announces messages to screen readers via live region
+	announce( message ) {
```
Garneauma, line 456: Clear the content of the announce div before adding new text. If the same text is added one time after the other, it will not be announced.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+
+		this.highlightedIndex = -1;
+		this.renderOptions();
+		this.openList();
+
+		if ( this.filteredOptions.length === 0 && !this.allOptionsSelected ) {
+			this.liveRegion.textContent = "";
+			setTimeout( () => {
+				this.announce( this.getLocalizedText().noMatchingOptions );
+			}, 1000 );
```
Garneauma: Shorten the delay to about 500. I think 1 full second is a bit long.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+			allTag.className = "tag tag-all-options";
+			allTag.type = "button";
+			allTag.dataset.tagValue = "all-options";
+			allTag.tabIndex = -1;
+			allTag.setAttribute( "part", "tag" );
+			allTag.setAttribute( "aria-label", `${ this.getLocalizedText().remove } ${ this.getLocalizedText().allOptionsSelected }` );
+
+			allTag.innerHTML = `
+				<span class="tag-text">${ this.getLocalizedText().allOptionsSelected }</span>
+				<span type="button" aria-hidden="true">×</span>
```
Garneauma: This should not have the type button.
```suggestion
				<span aria-hidden="true">×</span>
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+
+		// Always append the input
+		this.tagsContainer.appendChild( this.input );
+
+		if ( hadFocus ) {
+			this.input.focus();
+		}
+
+		if ( this.selectedItems.length > 0 ) {
+			this.tagsContainer.setAttribute( "role", "toolbar" );
```
Garneauma: Please add a comment explaining why this role was added (NVDA issue). Talk with @tristandasilva for a detailed explanation.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+		const map = {
+			"&": "&amp;",
+			"<": "&lt;",
+			">": "&gt;",
+
+			// This breaks the grunt dist build, if we do not need it in this context, we can leave it out
+			// '"': "&quot;",
+			"'": "&#039;"
+		};
+		return text.replace( /[&<>"']/g, m => map[ m ] );
```
Garneauma: You have removed the quotation replace string, but not the one in the regexp.
```suggestion
		return text.replace( /[&<>']/g, m => map[ m ] );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+		if ( !text ) {
+			return "";
+		}
+		const map = {
+			"&": "&amp;",
+			"<": "&lt;",
+			">": "&gt;",
+
+			// This breaks the grunt dist build, if we do not need it in this context, we can leave it out
+			// '"': "&quot;",
```
Garneauma: You can remove this. I think removing the opening and closing brackets is enough.
```suggestion
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+		this.selectedItems = [];
+		this.filteredOptions = [];
+		this.highlightedIndex = -1;
+		this.isOpen = false;
+		this.allOptions = [];
+		this.allOptionsSelected = false;
+		this.allowSelectAll = false; // Feature flag for select-all option
+		this.pageLanguage = "en"; // Default language
+		this.originalPlaceholder = ""; // Store original placeholder
+		this.instanceId = Math.random().toString( 36 ).substr( 2, 9 ); // Unique ID per instance
```
Garneauma: Is there a better solution to this that would ensure a unique ID?

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,853 @@
...
+						<div class="tags-container" id="tagsContainer">
+							<!-- Tags will be dynamically inserted here -->
+							<input
+								type="text"
+								role="combobox"
+								id="combo-box-input"
+								class="combo-box-input"
+								placeholder="${ this.escapeHtml( placeholder ) }"
+								autocomplete="off"
+								aria-label="${ this.escapeHtml( this.getLocalizedText().comboBoxInput ) }"
```
Garneauma: I don't think adding an aria-label to the input is valuable. Users would already know it's a combo box from the role. Remove the associated i18n string.

Garneauma (changes requested):
@BrahimMahadi Only some minor changes left!

@tristandasilva Thank you very much for your contribution. The component's navigation works very well!

FYI, I will ask for a review from another member of the team since I fear I might have missed some obvious issues as I have been so much involved in this.

components/combo-box/combo-box.js
BrahimMahadi: Removed in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Changed the code and created createTag(item) and createAllOptionsTag() to extract the element-creation logic out of renderTags() so it's reusable and renderTags() stays focused on diffing logic.

components/combo-box/combo-box.js
BrahimMahadi: Fixed in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Fixed in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Fixed in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Added comment in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Fixed in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Removed in recent commit

components/combo-box/combo-box.js
BrahimMahadi: Added a static counter for generating unique instance IDs

components/combo-box/combo-box.js
BrahimMahadi: Fixed in recent commit

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+ * Single file containing:
+ * - ComboBoxComponent Web Component class
+ * - Demo initialization code
+ */
+
+// ============================================
+// Accessible Combo Box Web Component
+// ============================================
+
+class ComboBoxComponent extends HTMLElement {
```
duboisp: I recommend to also include a reference to the namespace here to avoid name collision, not just in the HTML.

```suggestion
class GcComboBoxComponent extends HTMLElement {
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+				if ( !response.ok ) {
+					throw new Error( `Failed to load ${ optionsFile }: ${ response.statusText }` );
+				}
+				const data = await response.json();
+
+				// Convert array of objects into the expected format
+				// Each object has "name" (display label) and "tag" (internal value)
+				const options = data.map( ( item ) => ( {
+					value: item.tag,
+					label: item.name
```
duboisp, line 168: Not a blocker, but would it be possible to allow the author to specify those mapping value? Like if their JSON data array is not "tag" or "name".

Also, there was a security concern when JSON root was an Array. I think that was fix in browser today but I do think we should also support object like:

```
{
   "myoptions": [
      { "key": "yep", "value": "nope" },
      { "key": "yep1", "value": "nope1" },
      { "key": "yep2", "value": "nope2" },
  ]
}
``` 

or

```
{
   "myoptions": [
      { "yep": "nope" },
      { "yep1": "nope1" },
      { "yep2": "nope2" },
  ]
}
``` 

or

```
{
   "myoptions": [
      ["yep", "nope"],
      ["yep1", "nope1"],
      ["yep2", "nope2"]
  ]
}
``` 

Don't need to support all those format, but at least build some flexibility to allow the author to connect to their datasource and map it for the component itself.

I recommend to add a code sample in the working example of what that JSON document should look like.

---

If internally you are calling this "value / label" why to not use the same property naming in the JSON source? Something to think about, not an issue for now.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+				console.error( `Error loading options from ${ optionsFile }:`, e );
+				return [];
+			}
+		}
+
+		// Fallback to parsing from slot content if no options attribute
+		const slotOptions = this.querySelector( "[slot='options']" );
+		if ( slotOptions ) {
+			return Array.from( slotOptions.querySelectorAll( "option" ) ).map( opt => ( {
+				value: opt.value,
```
duboisp, line 185: By looking at this code, I think the pattern is enforced, I recommend to enforce it too all the way.

```suggestion
		const slotOptions = this.querySelector( "select[slot='options'], datalist[slot='options']" );
		if ( slotOptions ) {
			return Array.from( slotOptions.querySelectorAll( "option" ) ).map( opt => ( {
				value: opt.value,
```

After thinking, I am not sure if the select element is the best. Take a look at: https://html.spec.whatwg.org/multipage/form-elements.html#the-datalist-element

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+								aria-expanded="false"
+								aria-haspopup="listbox"
+							>
+						</div>
+						<ul
+							id="combo-box-list"
+							class="combo-box-list"
+							role="listbox"
+							aria-label="${ this.escapeHtml( this.getLocalizedText().availableOptions ) }"
+							tabindex="-1"
```
duboisp, line 234:
```suggestion
```

Is that required? `<ul>` is not a focusable element by default. Should we use `inert` attribute?

Note: The role `listbox` is map to the HTML element `datalist`. See : https://w3c.github.io/html-aria/#el-datalist I am not sure if that is or not an issue for now. I didn't cross check with the Aria Practice

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+			}
+		} );
+
+		// Event delegation for tag keyboard interactions
+		this.tagsContainer.addEventListener( "keydown", ( e ) => this.handleTagKeyDown( e ) );
+
+		// Event delegation for list options (click and hover)
+		this.list.addEventListener( "click", ( e ) => {
+			const option = e.target.closest( "[role='option']" );
+			if ( option && !option.hasAttribute( "aria-disabled" ) ) {
```
duboisp: Is an author can set themselves the aria-disabled?

We should not rely on Aria state for our logic. It will be better to have a sibling property that will be used instead and ARIA are simply complementing its value/signification. 

Like, if that element is an `<option>` you can set the attribute `disabled`. See: https://html.spec.whatwg.org/multipage/form-elements.html#attr-option-disabled

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+						this.selectOption( optionText );
+					}
+				}
+			}
+		} );
+
+		this.list.addEventListener( "mouseover", ( e ) => {
+			const option = e.target.closest( "[role='option']" );
+			if ( option ) {
+				const options = Array.from( this.list.querySelectorAll( "[role='option']" ) );
```
duboisp:
```suggestion
				const options = Array.from( this.list.querySelectorAll( "option" ) );
```

We should not select element based on their ARIA role/state. ARIA is only there to complement. 

Alternatively, you can add a CSS class to that element, at the same time your are assigning its role, and then use that CSS class to select the element here:

Or:

```suggestion
				const options = Array.from( this.list.querySelectorAll( ".option" ) );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+		}
+	}
+
+	// Handles keyboard navigation and interactions
+	handleKeyDown( e ) {
+		const key = e.key;
+
+		switch ( key ) {
+
+			case "ArrowLeft":
```
duboisp, line 351: We don't have "ArrowRight"?

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+				this.highlightNext();
+				break;
+			case "ArrowUp":
+				e.preventDefault();
+				this.highlightPrevious();
+				break;
+			case "Enter":
+				e.preventDefault();
+				if ( this.highlightedIndex >= 0 && this.isOpen ) {
+					const options = Array.from( this.list.querySelectorAll( "[role='option']" ) );
```
duboisp:
```suggestion
					const options = Array.from( this.list.querySelectorAll( "option" ) );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+	}
+
+	// Moves highlight to the next option
+	highlightNext() {
+		if ( !this.isOpen ) {
+			this.openList();
+			return;
+		}
+
+		const options = this.list.querySelectorAll( "[role='option']:not([aria-disabled='true'])" );
```
duboisp:
```suggestion
		const options = this.list.querySelectorAll( "option:not([disabled])" );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+	}
+
+	// Moves highlight to the previous option
+	highlightPrevious() {
+		if ( !this.isOpen ) {
+			this.openList();
+			return;
+		}
+
+		const options = this.list.querySelectorAll( "[role='option']:not([aria-disabled='true'])" );
```
duboisp:
```suggestion
		const options = this.list.querySelectorAll( "option:not([disabled])" );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+			this.highlightedIndex--;
+		} else {
+			this.highlightedIndex = options.length - 1;
+		}
+		this.updateHighlight();
+	}
+
+	// Updates the visual highlight for the current option
+	updateHighlight() {
+		const options = this.list.querySelectorAll( "[role='option']:not([aria-disabled='true'])" );
```
duboisp:
```suggestion
		const options = this.list.querySelectorAll( "option:not([disabled])" );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+	}
+
+	// Dispatches the change event for external listeners
+	dispatchChangeEvent() {
+		this.dispatchEvent( new CustomEvent( "change", {
+			detail: { selectedItems: [ ...this.selectedItems ] },
+			bubbles: true,
+			composed: true
+		} ) );
+	}
```
duboisp, line 470: Should the custom element be subscribed as a form element?

see: https://html.spec.whatwg.org/multipage/custom-elements.html#form-associated-custom-elements

Or, can you explain how that change event is used? I didn't see it from the working example.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+
+	// Creates a tag element for a selected item
+	createTag( item ) {
+		const tag = document.createElement( "button" );
+		tag.className = "tag";
+		tag.type = "button";
+		tag.dataset.tagValue = item.value;
+		tag.tabIndex = -1;
+		tag.setAttribute( "part", "tag" );
```
duboisp: Such styling possibility should be documented or it can be simply demoed in a separate working example page.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+		tag.setAttribute( "aria-label", `${ this.getLocalizedText().remove } ${ item.label }` );
+		tag.innerHTML = `<span class="tag-text">${ this.escapeHtml( item.label ) }</span><span class="tag-dismiss" aria-hidden="true">×</span>`;
```
duboisp: Do we need a label here or can we simply organize its inner content to be SR friendly.

```suggestion
		tag.innerHTML = `<span class="wb-inv">${ this.getLocalizedText().remove } </span><span class="tag-text">${ this.escapeHtml( item.label ) }</span><span class="tag-dismiss" aria-hidden="true">×</span>`;
```
Using innerHTML is risky,

Please parse it by DomPurify or use other safe method.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+	// Creates the "All options selected" tag element
+	createAllOptionsTag() {
+		const tag = document.createElement( "button" );
+		tag.className = "tag tag-all-options";
+		tag.type = "button";
+		tag.dataset.tagValue = "all-options";
+		tag.tabIndex = -1;
+		tag.setAttribute( "part", "tag" );
+		tag.setAttribute( "aria-label", `${ this.getLocalizedText().remove } ${ this.getLocalizedText().allOptionsSelected }` );
+		tag.innerHTML = `<span class="tag-text">${ this.getLocalizedText().allOptionsSelected }</span><span class="tag-dismiss" aria-hidden="true">×</span>`;
```
duboisp: See my previous comment about `aria-label` and `ìnnerHTML` optimization.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+				}
+			} );
+		}
+
+		// Update input state
+		this.input.value = "";
+		if ( allSelected ) {
+			this.input.placeholder = "";
+			this.tagsContainer.setAttribute( "aria-label", this.getLocalizedText().allOptionsSelected );
+			this.tagsContainer.setAttribute( "role", "toolbar" );
```
duboisp:
```suggestion
			this.tagsContainer.setAttribute( "role", "toolbar" ); // AT Compatibility - See following comment about the role=toolbar used the subsequent instruction
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+		} else {
+			this.input.placeholder = this.originalPlaceholder;
+			this.input.classList.toggle( "has-selections", this.selectedItems.length > 0 );
+			this.tagsContainer.removeAttribute( "aria-label" );
+			if ( this.selectedItems.length > 0 ) {
+				this.tagsContainer.setAttribute( "role", "toolbar" ); // Without role="toolbar", NVDA stays in browse mode when it's inside the tags container and intercepts the arrow key event before they reach our handlers. The "toolbar" role tells NVDA that this is a widget-like element and to allow arrow key events to pass through to our JavaScript. The role is removed when empty to avoid announcing "toolbar" when there are no selected tags
```
duboisp:
```suggestion
				this.tagsContainer.setAttribute( "role", "toolbar" ); // AT Compatibility - Without role="toolbar", NVDA stays in browse mode when it's inside the tags container and intercepts the arrow key event before they reach our handlers. The "toolbar" role tells NVDA that this is a widget-like element and to allow arrow key events to pass through to our JavaScript. The role is removed when empty to avoid announcing "toolbar" when there are no selected tags
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+			return;
+		}
+
+		// Create hidden input for each selected item to be submitted with forms
+		this.selectedItems.forEach( item => {
+			const input = document.createElement( "input" );
+			input.type = "hidden";
+			input.name = name;
+			input.value = item.value;
+			input.dataset.comboValue = "true";
```
duboisp, line 775: Why this flag? Do we have other hidden input in this combobox context?


```suggestion
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+			"<": "&lt;",
+			">": "&gt;"
+		};
+		return text.replace( /[&<>']/g, m => map[ m ] );
+	}
+
+	syncHiddenInputs() {
+
+		// Remove any existing hidden inputs created by this component
+		this.querySelectorAll( "input[type=\"hidden\"][data-combo-value]" ).forEach( el => el.remove() );
```
duboisp, line 762:
```suggestion
		this.querySelectorAll( "input[type=\"hidden\"]" ).forEach( el => el.remove() );
```

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+			const input = document.createElement( "input" );
+			input.type = "hidden";
+			input.name = name;
+			input.value = item.value;
+			input.dataset.comboValue = "true";
+			this.appendChild( input );
+		} );
+	}
+
+	// Public API: Get selected values as array (useful for form handling)
```
duboisp, line 780: We will need a little demo of all those public function to ensure they work as expected.

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+	getSelectedValues() {
+		return this.selectedItems.map( item => item.value );
+	}
+
+	// Public API: Get selected items
+	getSelectedItems() {
+		return [ ...this.selectedItems ];
+	}
```
duboisp, line 788: Should we be closer to the DOM API of the  `<select>` element?

For example:
```
getSelectedOptions() {
    return [ ...this.selectedItems ];
}
```

ref. https://html.spec.whatwg.org/multipage/form-elements.html#dom-select-selectedoptions

components/combo-box/combo-box.js
```diff
@@ -0,0 +1,858 @@
...
+		);
+		this.buildOptionsCache();
+		this.updateFilteredOptions();
+		this.renderTags();
+		this.renderOptions();
+	}
+}
+
+// Register the custom element
+customElements.define( "gc-combobox", ComboBoxComponent );
```
duboisp:
```suggestion
customElements.define( "gc-combobox", GcComboBoxComponent );
```

duboisp (dismissed):
@Garneauma For your consideration, see my inline comments. Some items could be completed at a later date.

Note: If you can't leverage directly a core HTML element (datalist/option) then I am ok if we are using the ul/li from a AT software compatibility point of view.

I recommend to not depend on ARIA for any logic, instead add some CSS class and use those class as elements selectors.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Easy change, please implement.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi This is a great improvement opportunity. Let's keep it in our list of improvements for the subsequent release.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi This is a great improvement opportunity. Let's keep it in our list of improvements for the subsequent release.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Let's look at the datalist element in our subsequent release.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Add a css class "disabled" and use this class for the checks. This brings another question though, is a disabled option the right HTML for when there are no items available? To be discussed.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Use the "combo-box-option" that is already available.

components/combo-box/combo-box.js
Garneauma: @duboisp There's no ArrowRight navigation when outside the tags container.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi use ".combo-box-option".

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Use ".combo-box-option:not(.disabled)".

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Same as comment above.

components/combo-box/combo-box.js
Garneauma: Same as comment above.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Please explore.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Please comment out for now. Let's add it as a todo in a subsequent release.

components/combo-box/combo-box.js
Garneauma: @duboisp escapeHtml() is safe. @BrahimMahadi Please use a hidden text instead of an aria-label.

components/combo-box/combo-box.js
Garneauma: Same comment as above.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi No problem adding this comment.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi No problem editing this comment.

components/combo-box/combo-box.js
Garneauma: @duboisp I think this was added to add context to the input. Not sure. @BrahimMahadi Is this valuable?

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Same comment as above.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Add a working example with a bunch of buttons under that call the various publicly accessible functions and log the values in the console.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Please explore.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi Please make the change.

components/combo-box/combo-box.js
tristandasilva: @duboisp @Garneauma This was added to prevent the removal of other hidden inputs not related to the items selected in the combobox when `syncHiddenInputs` is called. It's only valuable if the person using the component adds their own hidden input(s), which may not happen that often...but if it does, syncHiddenInputs would remove them

Garneauma: @tristandasilva @BrahimMahadi Can't you simply scope to the shadow root of this specific component? For example: `this.shadowRoot.querySelectorAll('input[type=hidden]);`

tristandasilva: @Garneauma Yeah that should also work and would be cleaner too. We can make that change!

tristandasilva: @Garneauma On second thought, I don't think this would work since the hidden inputs aren't being appended to the shadow DOM but rather the regular DOM. If they were appended to the shadow DOM the data wouldn't be sent in a form submission

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: Addressed in commit "addressed final changes"

components/combo-box/combo-box.js
BrahimMahadi: I agree, I updated the code so that we use selectedOptions instead of selectedItems for consistency.

components/combo-box/combo-box.js
Garneauma: @BrahimMahadi To explore: use ElementInternals instead of using hidden inputs.

BrahimMahadi:
## Things to be addressed in subsequent release(s)
- [ ] Explore ElementInternals concept to replace hidden inputs for form management and validation
- [ ] Add a working example that shows different mapping values for the JSON options
- [ ] Add possibility to define the root path for the data containing the array of items
- [ ] Look into leveraging Datalists for our options instead of Ul/Li
- [ ] Add "part" attributes to various items to enable customization outside the component

Garneauma (approved):

BeraJosh (approved):
As long as escapeHTML operates correctly for the InnerHTML I don't have any additional concerns in this js.

If this is tested to be the case then this is approved.

PR #2796, opened 2026-06-08
Update site-menu for accessibility

BrahimMahadi:
Update site-menu to be more accessible.

### DIFF
--- a/sites/gcweb-menu/gcweb-menu-docs-fr.html
+++ b/sites/gcweb-menu/gcweb-menu-docs-fr.html
@@ -4,7 +4,7 @@
 	"language": "fr",
 	"altLangPage": "gcweb-menu-docs-en.html",
 	"secondlevel": false,
-	"dateModified": "2026-05-20",
+	"dateModified": "2026-06-08",
 	"share": "true",
 	sitemenuPath: "ajax/sitemenu-v5-fr.html"
 }
@@ -17,7 +17,7 @@
 	<dt>Type</dt>
 	<dd>Fonctionnalité du site Canada.ca</dd>
 	<dt>Dernière révision</dt>
-	<dd>2026-05-20</dd>
+	<dd>2026-06-08</dd>
 </dl>
 
 <h2>Objectif</h2>
@@ -32,7 +32,7 @@
 <h2>Évaluation et rapport</h2>
 <p>Il n'y a pas d'évaluation ni de rapport disponible pour ce composant.</p>
 
-<h2>API (version 3.1.1)</h2>
+<h2>API (version 3.1.2)</h2>
 <table class="table bordée de table">
 	<tr>
 		<th>Classe CSS</th>
@@ -42,7 +42,7 @@
 	</tr>
 	<tr>
 		<td>Version 1.0</td>
-		<td>Version 3.1.1</td>
+		<td>Version 3.1.2</td>
 		<td>Version 1.0.1</td>
 		<td>n.d.</td>
 	</tr>
@@ -52,6 +52,9 @@
 <p><code>gcweb-menu</code></p>
 
 <h2>Historique</h2>
+
+<h3>Version 3.1.2</h3>
+<p>Ajout des attributs id et aria-labelledby au bouton de menu pour améliorer son accessibilité. Ajout également de l'attribut aria-hidden à l'élément span du bouton.</p>
 
 <h3>Version 3.1.1</h3>
 <p>L'attribut aria-label a été remplacé par aria-describedby afin d'améliorer l'accessibilité.</p>
--- a/sites/gcweb-menu/menu.js
+++ b/sites/gcweb-menu/menu.js
@@ -63,7 +63,13 @@ var componentName = "gcweb-menu",
 			setMnu3LevelOrientationExpandState( false, isMediumView );
 		}
 
-		const desc = document.createElement( "p" );
+		const nav = document.querySelector( " nav" + selector ),
+			buttonID = "mainMenuBtn",
+			desc = document.createElement( "p" );
+		subElm.previousElementSibling.setAttribute( "id", buttonID );
+		nav.setAttribute( "aria-labelledby", buttonID );
+		subElm.previousElementSibling.lastElementChild.setAttribute( "aria-hidden", "true" );
+
 		desc.id = "gcweb-menu-desc";
 		desc.hidden = true;
 		desc.textContent = i18nInstruction;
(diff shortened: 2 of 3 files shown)


Garneauma (changes requested):
Some aria changes missing. Also, please update the documentation.

BrahimMahadi:
I had added the aria-hidden = true to the wrong span tag, it should be good now. Documentation has also been updated.

sites/gcweb-menu/menu.js
```diff
@@ -63,15 +63,13 @@ var componentName = "gcweb-menu",
...
 		}
 
-		const desc = document.createElement( "p" );
-		desc.id = "gcweb-menu-desc";
-		desc.hidden = true;
-		desc.textContent = i18nInstruction;
-
-		subElm.previousElementSibling.insertAdjacentElement( "beforebegin", desc );
+		subElm.previousElementSibling.setAttribute( "id", "mainMenuBtn" );
+		const nav = document.getElementById( "wb-auto-2" );
```
Garneauma: Please do not use an auto-generated ID as a string. This would be different at any time if another plugin is initialized before this one. You could use `nav + selector` which would result in `nav.gcweb-menu`.

sites/gcweb-menu/menu.js
```diff
@@ -63,15 +63,13 @@ var componentName = "gcweb-menu",
 			setMnu3LevelOrientationExpandState( false, isMediumView );
 		}
 
-		const desc = document.createElement( "p" );
-		desc.id = "gcweb-menu-desc";
-		desc.hidden = true;
-		desc.textContent = i18nInstruction;
-
-		subElm.previousElementSibling.insertAdjacentElement( "beforebegin", desc );
+		subElm.previousElementSibling.setAttribute( "id", "mainMenuBtn" );
```
Garneauma: Store the id value in a variable so you can then reference it in the aria-label.

Garneauma (changes requested):

sites/gcweb-menu/menu.js
BrahimMahadi: Fixed to use nav + selector

sites/gcweb-menu/menu.js
BrahimMahadi: Id placed in variable and referenced

sites/gcweb-menu/menu.js
```diff
@@ -63,15 +63,14 @@ var componentName = "gcweb-menu",
...
-		const desc = document.createElement( "p" );
-		desc.id = "gcweb-menu-desc";
-		desc.hidden = true;
-		desc.textContent = i18nInstruction;
-
-		subElm.previousElementSibling.insertAdjacentElement( "beforebegin", desc );
+		const nav = document.querySelector( " nav" + selector ),
+			buttonID = "mainMenuBtn";
+		subElm.previousElementSibling.setAttribute( "id", buttonID );
+		nav.setAttribute( "aria-labelledby", subElm.previousElementSibling.id );
```
Garneauma:
```suggestion
		nav.setAttribute( "aria-labelledby", buttonID );
```

Garneauma (changes requested):

sites/gcweb-menu/menu.js
BrahimMahadi: Fixed

Garneauma (approved):

sites/gcweb-menu/menu.js
```diff
@@ -63,15 +63,14 @@ var componentName = "gcweb-menu",
...
-		subElm.previousElementSibling.insertAdjacentElement( "beforebegin", desc );
+		const nav = document.querySelector( " nav" + selector ),
+			buttonID = "mainMenuBtn";
+		subElm.previousElementSibling.setAttribute( "id", buttonID );
+		nav.setAttribute( "aria-labelledby", buttonID );
+		subElm.previousElementSibling.lastElementChild.setAttribute( "aria-hidden", "true" );
 
 		// Add menu navigation instruction
-		subElm.previousElementSibling.setAttribute( "aria-describedby", "gcweb-menu-desc" );
+		subElm.previousElementSibling.setAttribute( "aria-description", i18nInstruction );
```
nschonni: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-description mentions that this attribute is still draft for aria 1.3, and not official

Garneauma: @nschonni The current solution does not work as it point to a hidden element, which is not supported. Also, adding an aria-label overrides the button's text. We can't add a visually hidden element because we don't want screen-reader users to read it unnecessarily. This is the best solution we could come up with given the very urgent nature of the request. Testing it using NVDA and MacOS VoiceOver, it seems to work.

If you have a better solution, feel free to propose it. Thanks!

Keep in mind that we are planning to remove the GCWeb menu completely before EOF. Although a complete rework of the component for better accessibility would be ideal, we are focusing on removing current screen-reader annoyances.

nschonni: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-details maybe, but maybe someone from https://www.canada.ca/en/shared-services/services/employees-accessibility/aaact-program.html could give a better answer in terms of what devices might support `aria-description`, even if it's only in an draft for the next ARIA spec

Garneauma: @nschonni @BrahimMahadi After some investigation, we'll put back the aria-describedby since it seems it works well even if the target is hidden. Tested in NVDA, JAWS, and VoiceOver. We were mislead by our a11y trainer who believed the target could not be hidden.

Garneauma (changes requested):

sites/gcweb-menu/gcweb-menu-docs-fr.html
```diff
@@ -53,6 +53,9 @@ <h3>Classe CSS (v1.0)</h3>
 
 <h2>Historique</h2>
 
+<h3>Version 3.1.2</h3>
+<p>Ajout des attributs id, aria-labelledby et aria-description au bouton de menu pour améliorer son accessibilité. Ajout également de l'attribut aria-hidden à l'élément span du bouton.</p>
```
Garneauma: SVP mettre à jour la documentation pour refléter le changement.

Garneauma (changes requested):

sites/gcweb-menu/gcweb-menu-docs-fr.html
BrahimMahadi: Fixed

Garneauma (approved):

BeraJosh (approved):
Approved

PR #2806, opened 2026-06-19
combo-box hot fix

BrahimMahadi:
Hot fix for the combo-box to fix issue where after all options are selected, the list shows no matching options. Also made it so that when all options are selected and the combo box is submitted inside of a form, instead of sending the values of all selected tags, combo-box sends a value of "all".


duboisp:
@BrahimMahadi do we have request number for this? And what is the priority?

cc @tristandasilva

duboisp:
Closing this PR because it get superseded by #2807

PR #2807, opened 2026-06-22
Combobox additions and accessibility refinements

tristandasilva:
_Related to [WET-677](https://jtickets.atlassian.net/browse/WET-677)_

Preivew page: [GC Combo Box Web Component
](https://servicecanada.github.io/wet-boew-demos/gcweb-pr2807/components/combo-box/combo-box-en.html)
Changes include:
- Added the option to style the "Select all options" option from an external CSS file using the `::part(select-all)` selector
- Added optional class `select-all-unstyled` that users can add on the component to remove the default styling on the "Select all options" option
- Renamed the `select-all-options` attribute to `enable-all-options` and added the option to assign a value to replace the default "Select all options" text
- Added the optional attribute `all-options-tag` to allow for the customization of the "All options" tag text
- Changed the all options tag default text to "All options" instead of "All options selected" (screen readers still announce "selected")
- Hide empty state when all options are selected
- In form submissions: send "all" instead of every individual tag when all options are selected
- Updated documentation to reflect changes

Closes PR #2806

### DIFF
--- a/components/combo-box/combo-box-fr.html
+++ b/components/combo-box/combo-box-fr.html
@@ -1,7 +1,7 @@
 ---
 {
 	"title": "Composant Boîte Combo GC",
-	"dateModified": "2026-05-13",
+	"dateModified": "2026-07-06",
 	"breadcrumbs": [
 		{ "title": "Boîte Combo GC - Documentation", "link": "components/combo-box/combo-box-docs-fr.html" }
 	],
@@ -99,7 +99,7 @@
 </details>
 
 <h2>Exemple 6 : Avec l'option Sélectionner tout activée</h2>
-<gc-combobox placeholder="Fruit(s)..." name="fruits-select-all" select-all-options>
+<gc-combobox placeholder="Fruit(s)..." name="fruits-select-all" enable-select-all>
 	<span slot="label">Fruit(s)</span>
 	<select slot="options">
 		<option value="apple">Pomme</option>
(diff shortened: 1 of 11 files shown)


components/combo-box/combo-box-fr.html
```diff
@@ -122,7 +122,55 @@ <h2>Exemple 6 : Avec l'option Sélectionner tout activée</h2>
 &lt;/gc-combobox&gt;</code></pre>
 </details>
 
-<h2>Exemple 7 : Tests d'API publique</h2>
+<h2>Exemple 7 : Libellés « Sélectionner tout » personnalisés</h2>
```
Garneauma:
```suggestion
<h2>Exemple 7 : Libellé « Sélectionner tout » personnalisé</h2>
```

Garneauma (changes requested):
@tristandasilva 

Very good work on the "unstyled-select-all".

One note: I don't like the fact that the combobox does not announce if there are items selected when you focus on it. I think it should either say "[number] options selected" or "All options selected". What do you think?

tristandasilva:
@Garneauma Good catch! It would be good for AT users to know the state of the input when they focus into it. It was a little tricky to implement reliably, I got it to announce on NVDA, but on VoiceOver it's not very reliable. Check it out and let me know what you think :)

components/combo-box/combo-box.js
```diff
@@ -241,6 +253,7 @@ class GcComboBoxComponent extends HTMLElement {
 
 				<!-- Live region for screen reader announcements -->
 				<div id="liveRegion" class="sr-only" aria-live="polite" aria-atomic="true"></div>
+				<span id="combo-box-selection-desc" class="sr-only"></span>
```
Garneauma: For consitency, let's use WET's class.
```suggestion
				<span id="combo-box-selection-desc" class="wb-inv"></span>
```

Garneauma (commented):
@tristandasilva It works well. There is only one small thing. When selecting an item, it announces the number of items selected twice. I think we could remove the announcement either on the select action, or on the input focus.

tristandasilva:
> @tristandasilva It works well. There is only one small thing. When selecting an item, it announces the number of items selected twice. I think we could remove the announcement either on the select action, or on the input focus.

@Garneauma Updated! It should now only announce the number of items once. I removed the announcement on the select action and let the input focus handle it (note that I kept the announcement on the delete action since the focus goes to the next tag and not the input).

Garneauma (approved):

Garneauma:

Yves-Denis (approved):

PR #2825, opened 2026-07-27
fix fieldflow behaviour when restoring page from cache

tristandasilva:
Before a form submission, the plugin removes the `name` attribute from the controls that it generates (e.g. radio buttons). So if a user submits a form, gets redirected to a confirmation page, then presses the browser back button, the restored version is broken.

This fix stores the `name` attribute for each generated control so that on a page restore it can put them back. If any panels were selected/open prior to navigating away, their state gets restored accordingly.

**To reproduce the bug:**
- Go to https://canada-aem-s3-stage.adobecqms.net/en/news/subscribe-media-advisories.html **_using Edge of Firefox_** (doesn't happen always in Chrome)
- Make a test submission selecting "customized subscription" and whichever instiutions/regions
- Click subscribe
- Once you see the confirmation page, navigate back using the browser back button
- Notice customized subscription is selected but the panel doesn't show, and if you select "all advisories" both radio buttons are selected at once

You can see a working example with the fix implemented here:
https://servicecanada.github.io/wet-boew-demos/gcweb-pr2825/components/wb-fieldflow/alternative-en.html

### DIFF
--- a/components/wb-fieldflow/ajax/confirmation-en.html
+++ b/components/wb-fieldflow/ajax/confirmation-en.html
@@ -1,0 +1,9 @@
+---
+title: "Submission received"
+language: "en"
+altLangPage: confirmation-fr.html
+dateModified: "2026-07-29"
+---
+<p>This page only exists as a confirmation page for the Fieldflow submitting form example.</p>
+<h2 class="h4">Testing restore from cache</h2>
+<p>Press your browser's <strong>Back</strong> button to return to the form. Your selection should be restored as you left it.</p>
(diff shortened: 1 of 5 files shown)


Garneauma (changes requested):
@tristandasilva Can you please add a working example at the end of "alternative-en|fr.html" so we can see this in action? I don't think we have an example that has the issue right now.

Garneauma (approved):

Garneauma:

Yves-Denis (approved):

PR #2837, opened 2026-08-12
add fieldflow prop to preserve content visibility on submit

tristandasilva:
When a form is submitted, a clean event is triggered before the submission goes through, which undoes any live toggle or ajax action. That results in the content being shown by a selection collapsing/emptying before the browser navigates away.

This PR creates an option for users to opt out of the clean event that is triggered on submit to keep content visible during submission. To opt out the clean event, the user can add `"keepOnSubmit": true` to the config options.

Live example can be seen [here](https://servicecanada.github.io/wet-boew-demos/gcweb-pr2837/components/wb-fieldflow/alternative-en.html#wb-auto-88)

### DIFF
--- a/components/wb-fieldflow/alternative-en.html
+++ b/components/wb-fieldflow/alternative-en.html
@@ -6,7 +6,7 @@ category: "Plugins"
 description: "Find styling alternatives for field flow."
 tag: "fieldflow"
 parentdir: "fieldflow"
-dateModified: "2023-12-01"
+dateModified: "2026-08-12"
 ---
 <ul class="list-inline">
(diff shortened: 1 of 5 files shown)


Garneauma (approved):
This is a patch change.

Works as intended. I went through a lot of examples and everything works as before, so it does not seem to introduce any regression.

Garneauma:

Yves-Denis (approved):
Security Assessment Completed - Passed, no security issues introduced. 
Only interaction with Form-submitted content, provided input is sanitized and validated, I don't see any concerns.

